PR #3, opened 2021-06-18
Generalize HTTP client code

djoreilly:
Allow verbs other than GET.
Provide parameter for request body.
Check range of status codes for success.

### DIFF
--- a/connect/api.go
+++ b/connect/api.go
@@ -15,7 +15,7 @@ const (
 func UpToDate() bool {
 	// REVIST 404 case - see original
 	// Should fail in any case. 422 error means that the endpoint is there and working right
-	_, err := DoGET(Credentials{}, "/connect/repositories/installer")
+	_, err := callHTTP("GET", "/connect/repositories/installer", nil, nil, Credentials{})
 	if err == nil {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


skazi0 (changes requested):
We also need query params (see https://github.com/SUSE/connect-ng/pull/4/commits/5ec7e83892a833ea587b12895b03ea8a0dc87c63)

djoreilly:
> We also need query params (see [5ec7e83](https://github.com/SUSE/connect-ng/commit/5ec7e83892a833ea587b12895b03ea8a0dc87c63))

Good point. I'll wait until your changes are merged and rebase.

skazi0:
@djoreilly I think it would be better if you add params handling to your change (as it's more generic)... and then I'll rebase :o)

djoreilly:
> @djoreilly I think it would be better if you add params handling to your change (as it's more generic)... and then I'll rebase :o)

ok will do

skazi0 (approved):
tested in https://github.com/SUSE/connect-ng/pull/4

skazi0 (approved):

skazi0 (approved):

skazi0 (approved):

PR #4, opened 2021-06-18
Add list-extensions CLI 

skazi0:
New CLI parameter was added for listing extension products.

### DIFF
--- a/connect/api.go
+++ b/connect/api.go
@@ -44,3 +44,16 @@ func GetActivations() (map[string]Activation, error) {
 	}
 	return activeMap, nil
 }
+
+func GetProduct(productQuery Product) (Product, error) {
+	resp, err := callHTTP("GET", "/connect/systems/products", nil, productQuery.toQuery(), authSystem)
+	remoteProduct := Product{}
+	if err != nil {
+		return remoteProduct, err
+	}
+	err = json.Unmarshal(resp, &remoteProduct)
+	if err != nil {
+		return remoteProduct, err
+	}
+	return remoteProduct, nil
+}
--- a/connect/api_test.go
+++ b/connect/api_test.go
@@ -84,3 +84,44 @@ func TestUpToDateOkay(t *testing.T) {
 		t.Error("Expecting UpToDate()==true, got false")
 	}
 }
+
+func TestGetProduct(t *testing.T) {
+	createTestCredentials("", "", t)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Write(readTestFile("product.json", t))
+	}))
+	defer ts.Close()
+
+	CFG.BaseURL = ts.URL
+	productQuery := Product{Name: "SLES", Version: "15.2", Arch: "x86_64"}
+	product, err := GetProduct(productQuery)
+	if err != nil {
+		t.Fatalf("%s", err)
+	}
+	if l := len(product.Extensions); l != 1 {
+		t.Fatalf("len(product.Extensions) == %d, expected 1", l)
+	}
+	if l := len(product.Extensions[0].Extensions); l != 8 {
+		t.Fatalf("len(product.Extensions[0].Extensions) == %d, expected 8", l)
+	}
+
+}
+
+func TestGetProductError(t *testing.T) {
+	createTestCredentials("", "", t)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		errMsg := "{\"status\":422,\"error\":\"No product specified\",\"type\":\"error\",\"localized_error\":\"No product specified\"}"
+		http.Error(w, errMsg, http.StatusUnprocessableEntity)
+	}))
+	defer ts.Close()
+
+	CFG.BaseURL = ts.URL
+	productQuery := Product{Name: "Dummy"}
+	_, err := GetProduct(productQuery)
+	if ae, ok := err.(APIError); ok {
+		if ae.Code != http.StatusUnprocessableEntity {
+			t.Fatalf("Expecting APIError(422). Got %s", err)
+		}
+	}
+}
--- a/connect/config_test.go
+++ b/connect/config_test.go
@@ -1,6 +1,7 @@
 package connect
 
 import (
+	"reflect"
 	"strings"
 	"testing"
 )
@@ -27,7 +28,7 @@ func TestParseConfig(t *testing.T) {
 	expect := Config{BaseURL: "https://smt-azure.susecloud.net", Language: "en_US.UTF-8"}
 	c := Config{}
 	parseConfig(r, &c)
-	if c != expect {
+	if !reflect.DeepEqual(c, expect) {
 		t.Errorf("got %+v, expected %+v", c, expect)
 	}
 }
@@ -37,7 +38,7 @@ func TestParseConfig2(t *testing.T) {
 	expect := Config{BaseURL: "http://example.com", Language: "en_US.UTF-8", Insecure: true}
 	c := Config{}
 	parseConfig(r, &c)
-	if c != expect {
+	if !reflect.DeepEqual(c, expect) {
 		t.Errorf("got %+v, expected %+v", c, expect)
 	}
 }
--- a/connect/extensions.go
+++ b/connect/extensions.go
@@ -1,0 +1,94 @@
+package connect
+
+import (
+	"bytes"
+	_ "embed" // golint
+	"sort"
+	"strings"
+	"text/template"
+)
+
+var (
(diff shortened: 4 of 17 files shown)


connect/zypper_test.go
```diff
@@ -1,29 +1,19 @@
...
-    <product isbase="false" name="SUSE-MicroOS" version="5.0" release="1" epoch="0" arch="x86_64" vendor="SUSE" summary="SUSE Linux Enterprise Micro 5.0" repo="@System" productline="SUSE-MicroOS" registerrelease="" shortname="SUSE Linux Enterprise Micro" flavor="" installed="true"><endoflife time_t="1606694400" text="2020-11-30T00:00:00Z"/><registerflavor/><description>SUSE Linux Enterprise Micro 5.0</description></product>
-    <product isbase="false" name="suse-openstack-cloud" version="8" release="0" epoch="0" arch="x86_64" vendor="SUSE LINUX GmbH, Nuernberg, Germany" summary="SUSE OpenStack Cloud 8" repo="@System" productline="suse-openstack-cloud" registerrelease="" shortname="SOC8" flavor="POOL" installed="true"><endoflife time_t="1622419200" text="2021-05-31T00:00:00+00"/><registerflavor>extension</registerflavor><description>SUSE OpenStack Cloud 8</description></product>
-  </product-list>
-</stream>
-`
+var (
+	//go:embed testdata/products.xml
+	testProducts string
+	//go:embed testdata/products-no-base.xml
+	testProductsNoBase string
```
djoreilly, line 15: I'm not sure about moving testdata into the packager directory. I think it's more conventional to have it outside. Maybe we don't need  to use go:embed for test binaries if we do something like this https://github.com/SUSE/connect-ng/pull/10

connect/zypper.go
```diff
@@ -67,3 +67,17 @@ func parseProductsXML(xmlDoc []byte) ([]Product, error) {
 	}
 	return products.Products, nil
 }
+
+// TODO: memoize?
```
djoreilly, line 71: yeah, we should come back to that - zypper is surprising slow.

connect/api.go
```diff
@@ -44,3 +44,17 @@ func GetActivations(creds Credentials) (map[string]Activation, error) {
 	}
 	return activeMap, nil
 }
+
+func GetProduct(productQuery Product, creds Credentials) (Product, error) {
+	urlSuffix := "/connect/systems/products"
+	resp, err := DoGETWithParams(creds, urlSuffix, productQuery.toQuery())
```
djoreilly: If I checkout this commit,  DoGETWithParams does not exist.

connect/api_test.go
```diff
@@ -86,3 +89,44 @@ func TestUpToDateOkay(t *testing.T) {
 		t.Error("Expecting UpToDate()==true, got false")
 	}
 }
+
+func TestGetProduct(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Write([]byte(testProduct))
```
djoreilly, line 96: Another thing about using go embed is that now you have to scroll to the top to see which file testProduct comes from.

connect/api.go
skazi0: Bad rebase. Fixed. Thx!

connect/zypper_test.go
skazi0: #10 looks great! It solves the problem of test data in the code (which doesn't have syntax coloring, is hard to compare samples and which is generally ugly) and saves us trouble of inventing new names for test variables every time we get a new test file. I'll rebase once we have that merged.

connect/api_test.go
skazi0: Will be solved by #10

connect/zypper_test.go
skazi0: done

connect/api_test.go
```diff
@@ -82,3 +82,44 @@ func TestUpToDateOkay(t *testing.T) {
 		t.Error("Expecting UpToDate()==true, got false")
 	}
 }
+
+func TestGetProduct(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Write(readTestFile("product.json", t))
```
djoreilly, line 89: if readTestFile() calls Fatal(), it won't actually stop the test, because it's being called from the handler goroutine and not the test goroutine. See https://golang.org/pkg/testing/#T.FailNow

connect/config_test.go
```diff
@@ -27,7 +28,7 @@ func TestParseConfig(t *testing.T) {
 	expect := Config{BaseURL: "https://smt-azure.susecloud.net", Language: "en_US.UTF-8"}
 	c := Config{}
 	parseConfig(r, &c)
-	if c != expect {
+	if !reflect.DeepEqual(c, expect) {
```
djoreilly, line 31: I missed that when adding Product to the config

connect/extensions.go
```diff
@@ -0,0 +1,98 @@
...
+	for _, e := range extensions {
+		_, activated := activations[e.toTriplet()]
+		ret = append(ret, displayExtension{
+			Product:       e,
+			Code:          e.toTriplet(),
+			Activated:     activated,
+			Subextensions: preformatExtensions(e.Extensions, activations, level+1),
+			Indent:        strings.Repeat("    ", level),
+			// TODO: root_fs_writable? ? 'SUSEConnect' : 'transactional-update register'
+			ConnectBinary: "SUSEConnect",
```
djoreilly, line 68: there is  AppName const in main.go

connect/zypper_test.go
```diff
@@ -1,20 +1,66 @@
...
+func TestZypperHelper(t *testing.T) {
+	// if ran directly, do nothing
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+	testcase := os.Getenv("GO_TESTCASE")
+	switch testcase {
+	case testCaseInstalledProducts:
+		fmt.Fprintf(os.Stdout, string(readTestFile("products.xml", t)))
+		os.Exit(0)
```
djoreilly, line 37: why is os.Exit here?

connect/zypper_test.go
```diff
@@ -1,20 +1,66 @@
...
+	case testCaseInstalledProductsNoBase:
+		fmt.Fprintf(os.Stdout, string(readTestFile("products-no-base.xml", t)))
```
djoreilly, line 39: you don't need to use the string() if you
`fmt.Fprintf(os.Stdout, "%s", readTestFile("products-no-base.xml", t))`
but it's all the same

djoreilly (commented):
Looking really good. Added a few comments.

connect/api_test.go
skazi0: The test will fail with:
```
=== RUN   TestGetProduct
    helpers_test.go:14: 
        api_test.go:92: open ../testdata/product.json: no such file or directory
    api_test.go:100: Get "http://127.0.0.1:38759/connect/systems/products?arch=x86_64&identifier=SLES&version=15.2": EOF
```
I think it's good enough.

connect/extensions.go
skazi0: Note the comment. It won't be just "binary" that we need to print here.

connect/zypper_test.go
skazi0: As we're simulating command execution, it's there to explicitly show that the command exits with zero. I added it to other cases to make it clear.

connect/api_test.go
djoreilly: okay with me.

connect/extensions.go
djoreilly: okay

connect/zypper_test.go
skazi0: I changed it to `os.Stdout.Write(readTestFile("products-no-base.xml", t))`.
I think this is the best pattern in case we need some binary test files at some point.

connect/zypper_test.go
djoreilly: thanks

connect/zypper_test.go
djoreilly: yep that's even better

connect/connection.go.orig
```diff
@@ -0,0 +1,68 @@
+package connect
```
djoreilly, line 1: using `git add .` ?

djoreilly (changes requested):
please delete connect/connection.go.orig

connect/connection.go.orig
skazi0: yeah... lazy Friday mode :o)

djoreilly (approved):
lgtm

PR #6, opened 2021-06-21
Complete credentials file handling

djoreilly:
Add methods to read, write and remove the system credentials
and service credentials files. These files are absolute from
/ or from the directory specified with --root.

### DIFF
--- a/connect/api_test.go
+++ b/connect/api_test.go
@@ -9,6 +9,11 @@ import (
 
 func TestGetActivations(t *testing.T) {
 	response := readTestFile("activations.json", t)
+	CFG.FsRoot = t.TempDir()
+	if err := writeSystemCredentials("user1", "pass1"); err != nil {
+		t.Fatalf("Unexpected error: %s", err)
+	}
+
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 		w.Write(response)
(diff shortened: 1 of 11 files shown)


connect/api_test.go
```diff
@@ -41,7 +45,11 @@ func TestGetActivationsRequest(t *testing.T) {
 		url        = "/connect/systems/activations"
 		gotRequest *http.Request
 	)
-
+	CFG.FsRoot = t.TempDir()
+	if err := writeSystemCredentials(user, password); err != nil {
+		t.Errorf("Unexpected error: %s", err)
+		return
```
djoreilly, line 51: Will change these to use t.Fatalf() and then no need for return

skazi0 (approved):
lgtm

connect/credentials.go
```diff
@@ -39,6 +40,11 @@ func serviceCredentialsFile(service string) string {
 	return filepath.Join(CFG.FsRoot, defaulCredentialsDir, service)
 }
 
+func (c Credentials) authHeader() string {
+	auth := c.Username + ":" + c.Password
+	return "Basic " + base64.StdEncoding.EncodeToString([]byte(auth))
```
skazi0, line 45: why do you generate the header by hand instead of `req.SetBasicAuth()` we had before?

skazi0 (changes requested):

connect/credentials.go
djoreilly: > why do you generate the header by hand instead of `req.SetBasicAuth()` we had before?

Good point. I don't remember. I will change it back. Thanks.

connect/api_test.go
djoreilly: Actually a helper function might be better. Like:
createTestCredentials(user, pass string, t *testing.T)

skazi0 (approved):
lgtm

PR #7, opened 2021-06-21
Strip trailing newline from execute() output

djoreilly:


skazi0 (changes requested):
Why do we need that? Maybe add some details to the commit message?

djoreilly:
> Why do we need that? Maybe add some details to the commit message?

I updated the commit message. Thanks.

skazi0 (approved):

PR #8, opened 2021-06-21
Add the regcode, product and instance-data cli options

djoreilly:
This commit just adds these flags to the cli, and checks for
invalid combinations like the Ruby version.

### DIFF
--- a/cmd/suseconnect.go
+++ b/cmd/suseconnect.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/SUSE/connect-ng/connect"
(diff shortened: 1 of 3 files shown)


connect/config.go
```diff
@@ -18,12 +18,15 @@ const (
...
-	Namespace string
-	FsRoot    string
+	Path             string
+	BaseURL          string
+	Language         string
+	Insecure         bool
+	Namespace        string
+	FsRoot           string
+	Token            string
+	Product          string
```
djoreilly, line 28: I think this should type Product{} not a string.

cmd/suseconnect.go
```diff
@@ -94,6 +115,21 @@ func main() {
 	} else if deregister {
 		err := connect.Deregister()
 		exitOnError(err)
+	} else {
+		if instanceDataFile != "" && connect.URLDefault() {
```
skazi0, line 119: why is this nested and not just another `else if`?

djoreilly: > why is this nested and not just another `else if`?

I'm trying to follow the flow of the Ruby version:
https://github.com/SUSE/connect/blob/master/lib/suse/connect/cli.rb#L38-L51

skazi0: hmm... I guess that's fine then.

skazi0 (approved):
lgtm

cmd/suseconnect.go
djoreilly: I agree it's not nice. Maybe after we have parity of functionality we can come back and improve the whole cli option processing.

PR #9, opened 2021-06-23
Implement hwinfo

djoreilly:
This will be used on x86_64 and arm64 to find the CPU and socket
count, and "Hypervisor vendor" on VMs. A map[string]string is
returned. The callers can lookup relevant keys and convert the
value to another type as needed.

### DIFF
--- a/connect/api.go
+++ b/connect/api.go
@@ -107,3 +107,42 @@ func deregisterSystem() error {
 	_, err := callHTTP("DELETE", "/connect/systems", nil, nil, authSystem)
 	return err
 }
+
+// updateSystem updates the system's hardware information on SCC
+// https://scc.suse.com/connect/v4/documentation#/systems/put_systems
+// The body parameter is produced by makeSysInfoBody()
+func updateSystem(body []byte) error {
+	_, err := callHTTP("PUT", "/connect/systems", body, nil, authSystem)
+	return err
+}
+
+// makeSysInfoBody returns the JSON payload needed for the announce/update system calls
+func makeSysInfoBody(distroTarget, namespace string, instanceData []byte) ([]byte, error) {
+	var payload struct {
+		Hostname     string `json:"hostname"`
+		DistroTarget string `json:"distro_target"`
+		InstanceData string `json:"instance_data,omitempty"`
+		Namespace    string `json:"namespace,omitempty"`
+		Hwinfo       hwinfo `json:"hwinfo"`
+	}
+	if distroTarget != "" {
+		payload.DistroTarget = distroTarget
+	} else {
+		var err error
+		payload.DistroTarget, err = zypperDistroTarget()
+		if err != nil {
+			return nil, err
+		}
+	}
+	payload.InstanceData = string(instanceData)
+	payload.Namespace = namespace
+
+	hw, err := getHwinfo()
+	if err != nil {
+		return nil, err
+	}
+	payload.Hostname = hw.Hostname
+	payload.Hwinfo = hw
+
+	return json.Marshal(payload)
+}
--- a/connect/client.go
+++ b/connect/client.go
@@ -2,7 +2,21 @@ package connect
 
 import (
 	"fmt"
+	"os"
 )
+
+// Register announces the system, activates the
+// product on SCC and adds the service to the system
+func Register() error {
+	printInformation("register")
+	err := announceOrUpdate()
+	if err != nil {
+		return err
+	}
+
+	// TODO remainder of Register()
+	return nil
+}
 
 // Deregister deregisters the system
 func Deregister() error {
@@ -96,6 +110,35 @@ func removeOrRefreshService(service Service) error {
 	}
 	fmt.Println("-> Removing service from system ...")
 	return removeService(service.Name)
+}
+
+// UpdateSystem resend the system's hardware details on SCC
+func UpdateSystem(distroTarget, instanceDataFile string) error {
+	fmt.Printf("\nUpdating system details on %s ...\n", CFG.BaseURL)
+	var instanceData []byte
+	if instanceDataFile != "" {
+		var err error
+		instanceData, err = os.ReadFile(instanceDataFile)
+		if err != nil {
+			return err
+		}
+	}
+	sysInfoBody, err := makeSysInfoBody(distroTarget, CFG.Namespace, instanceData)
+	if err != nil {
+		return err
+	}
+	return updateSystem(sysInfoBody)
+}
+
+// announceOrUpdate Announces the system to the server, receiving and storing its
+// credentials. When already announced, sends the current hardware details to the server
+func announceOrUpdate() error {
+	if IsRegistered() {
+		return UpdateSystem("", "")
+	}
+
+	// TODO remainder of announceOrUpdate()
+	return nil
 }
 
 // IsRegistered returns true if there is a valid credentials file
--- a/connect/hwinfo.go
+++ b/connect/hwinfo.go
@@ -119,3 +119,18 @@ func privateIP(ip net.IP) bool {
 	}
 	return false
 }
+
+func hostname() string {
+	name, err := os.Hostname()
+	// TODO the Ruby version has this "(none)" check - why?
+	if err == nil && name != "" && name != "(none)" {
+		return name
+	}
+	Debug.Print(err)
+	ip, err := getPrivateIPAddr()
+	if err != nil {
+		Debug.Print(err)
+		return ""
+	}
+	return ip
+}
--- a/connect/hwinfo.go
+++ b/connect/hwinfo.go
@@ -3,7 +3,13 @@ package connect
 import (
 	"bufio"
 	"bytes"
+	"regexp"
 	"strings"
+)
+
+var (
+	cloudEx = `Version: .*(amazon)|Manufacturer: (Amazon)|Manufacturer: (Google)|Manufacturer: (Microsoft) Corporation`
+	cloudRe = regexp.MustCompile(cloudEx)
 )
 
 func lscpu() (map[string]string, error) {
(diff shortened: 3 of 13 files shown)


djoreilly:
There are still more commits to come on this branch.

connect/hwinfo.go
```diff
@@ -0,0 +1,30 @@
+package connect
+
+import (
+	"bufio"
+	"bytes"
+	"strings"
+)
+
+func lscpu() (map[string]string, error) {
+	output, err := execute([]string{"lscpu"}, false, nil)
```
skazi0, line 26: `lscpu` has localized outputs. We should probably set some env variables to make sure we get expected output for parsing.

connect/hwinfo.go
```diff
@@ -88,3 +89,42 @@ func uuid() (string, error) {
 	}
 	return string(output), nil
 }
+
+// getPrivateIPAddr returns the first private IP address on the host
+func getPrivateIPAddr() (string, error) {
+	output, err := execute([]string{"ip", "-4", "--oneline", "--brief", "address"}, false, nil)
```
skazi0, line 95: Why not use the built in `net.InterfaceAddrs()` like the ruby version did with `Socket.ip_address_list` ?

djoreilly: > Why not use the built in `net.InterfaceAddrs()` like the ruby version did with `Socket.ip_address_list` ?

sure - that  way looks simpler. I will change it

connect/hwinfo.go
```diff
@@ -130,3 +130,26 @@ func hostname() string {
 	}
 	return ip
 }
+
+func readValues(arg string) (map[string]string, error) {
+	output, err := execute([]string{"read_values", arg}, false, nil)
```
skazi0, line 135: It would be good to add some info to the code and/or commit message about this being specific to S390.

connect/hwinfo.go
```diff
@@ -28,3 +106,114 @@ func lscpu2map(b []byte) map[string]string {
...
+	Debug.Print(err)
+	ip, err := getPrivateIPAddr()
+	if err != nil {
+		Debug.Print(err)
+		return ""
+	}
+	return ip
+}
+
+// readValues calls read_values from SUSUE/s390-tools
```
skazi0, line 196: typo

djoreilly: well spotted. Will fix with next push

connect/hwinfo.go
```diff
@@ -53,3 +62,29 @@ func findCloudProvider(b []byte) string {
...
+		content, err := os.ReadFile("/sys/hypervisor/uuid")
+		if err != nil {
+			return "", err
+		}
+		return string(content), nil
+	}
+	output, err := execute([]string{"dmidecode", "-s", "system-uuid"}, false, nil)
+	if err != nil {
+		return "", err
+	}
```
djoreilly, line 151: TODO need to check for ['Not Settable', 'Not Present']
https://github.com/SUSE/connect/blob/master/lib/suse/connect/hwinfo/x86.rb#L34

connect/api.go
```diff
@@ -57,3 +58,46 @@ func GetProduct(productQuery Product) (Product, error) {
 	}
 	return remoteProduct, nil
 }
+
+// updateSystemAPI resend the system's hardware info to SCC.
+// https://github.com/SUSE/connect/wiki/SCC-API-%28Implemented%29#update-system
+func updateSystemAPI(distroTarget, namespace string, instanceData []byte) error {
```
djoreilly, line 64: Maybe putSystem() is better here.

skazi0: why not just `updateSystem()` ?

skazi0: btw, the link in the comment doesn't seem to be valid anymore

connect/api.go
```diff
@@ -57,3 +58,46 @@ func GetProduct(productQuery Product) (Product, error) {
...
+	if err != nil {
+		return err
+	}
+	body := bytes.NewReader(payload)
+	_, err = callHTTP("PUT", "/connect/systems", body, nil, authSystem)
+	return err
+}
+
+// updatePayload returns the JSON body needed for the announce and update calls
+func updatePayload(distroTarget, namespace string, instanceData []byte) ([]byte, error) {
```
skazi0, line 75: this should be something like `systemUpdatePayload()` as there are other update APIs like `PUT /connect/systems/products`

connect/api.go
djoreilly: > why not just `updateSystem()` ?

because I have already taken that name in client.go, so it needs to something different. I like putSystem()  here, and it follows a convention we already have with GetProduct() and GetActivations()

djoreilly: > > why not just `updateSystem()` ?
> 
> because I have already taken that name in client.go, so it needs to something different. I like putSystem() here, and it follows a convention we already have with GetProduct() and GetActivations()

thanks - I should have check that

connect/api.go
```diff
@@ -57,3 +58,46 @@ func GetProduct(productQuery Product) (Product, error) {
...
 }
+
+// updateSystemAPI resend the system's hardware info to SCC.
+// https://github.com/SUSE/connect/wiki/SCC-API-%28Implemented%29#update-system
+func updateSystemAPI(distroTarget, namespace string, instanceData []byte) error {
+	payload, err := updatePayload(distroTarget, namespace, instanceData)
+	if err != nil {
+		return err
+	}
+	body := bytes.NewReader(payload)
```
djoreilly, line 69: I think it would be more natural and convenient for callHTTP() to receive a []byte instead of io.Reader for the body. I will propose a change for that

connect/hwinfo.go
djoreilly: done

connect/api.go
djoreilly: done

connect/api.go
djoreilly: I made some changes around here. Maybe better?

connect/connection.go
```diff
@@ -58,14 +59,14 @@ func successCode(code int) bool {
 	return code >= 200 && code < 300
 }
 
-func callHTTP(verb, path string, body io.Reader, query map[string]string, auth authType) ([]byte, error) {
+func callHTTP(verb, path string, body []byte, query map[string]string, auth authType) ([]byte, error) {
```
skazi0, line 62: Please take this change out to a separate PR to make parallel work easier and limit future rebasing.

connect/client.go
```diff
@@ -24,3 +63,25 @@ func IsRegistered() bool {
...
+		fmt.Println("Registering system to ", server)
+	} else {
+		fmt.Println("Deregistering system from ", server)
+	}
+	if CFG.FsRoot != "" {
+		fmt.Println("Rooted at: ", CFG.FsRoot)
+	}
+
+	// TODO
+	// if CFG.Email != "" {
```
skazi0, line 84: any idea where this email comes from?

connect/client.go
```diff
@@ -24,3 +63,25 @@ func IsRegistered() bool {
 func URLDefault() bool {
 	return CFG.BaseURL == defaultBaseURL
 }
+
+func printInformation(action string) {
```
skazi0, line 67: This would also be a good candidate for a separate PR. I could use it directly in my Deregister code.

connect/api.go
skazi0: looks good

connect/api.go
skazi0: nice. thx!

connect/connection.go
djoreilly: will do

connect/client.go
djoreilly: need to research that - the cli I think

connect/client.go
djoreilly: I will split it out

connect/connection.go
djoreilly: done

connect/client.go
djoreilly: done

connect/client.go
skazi0: yeah, now I see that. I need to get my vision checked :o)

djoreilly:
rebased

djoreilly:
Need to rebase this and make changes for the new naming convention

connect/api.go
```diff
@@ -84,3 +84,41 @@ func deactivateProduct(product Product) (Service, error) {
 	err = json.Unmarshal(resp, &remoteService)
 	return remoteService, err
 }
+
+// updateSystem updates the system's hardware info on SCC
+// https://scc.suse.com/connect/v4/documentation#/systems/put_systems
```
skazi0, line 89: please add a similar comment about the format of this data and `makeSysInfoBody()`

skazi0 (approved):
looks good so far. great work!

djoreilly:
rebased

connect/api.go
djoreilly: done

connect/hwinfo.go
```diff
@@ -0,0 +1,262 @@
...
+	Hypervisor    string `json:"hypervisor"`
+	Arch          string `json:"arch"`
+	UUID          string `json:"uuid"`
+	CloudProvider string `json:"cloud_provider"`
+}
+
+func getHwinfo() (hwinfo, error) {
+	hw := hwinfo{}
+	var err error
+	hw.Arch, err = arch()
```
skazi0, line 37: with `err` defined above, these could be folded into the `if`s below.

skazi0: i.e. 
```
if hw.Arch, err = arch(); err != nil {
    return hwinfo{}, err
}

File: connect/hwinfo.go [hunk 14]
```diff
@@ -0,0 +1,262 @@
...
+	"bytes"
+	"net"
+	"os"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var (
+	cloudEx = `(?m)Version: .*(amazon)|Manufacturer: (Amazon)|Manufacturer: (Google)|Manufacturer: (Microsoft) Corporation`
```
skazi0, line 14: I don't think it should be a multiline match.

skazi0 (approved):
lgtm. just some nit picking. feel free to ignore.

connect/hwinfo.go
djoreilly: yeah, I'll take it out. I can't remember what I was trying

djoreilly: done

connect/hwinfo.go
djoreilly: done

skazi0 (approved):

PR #10, opened 2021-06-25
Add helper function for tests to read files from testdata

djoreilly:
Many tests read files from ../testdata. This patch moves the
repeatitive error checking and calls to t.Fatal() out to a
helper function. It can be called from any *_test.go, and that
filename and line number will be printed if there is an error.

For example: if activations.json got deleted:

--- FAIL: TestGetActivations (0.00s)
    helpers_test.go:14:
        api_test.go:11: open ../testdata/activations.json: no such file or directory

### DIFF
--- a/connect/api_test.go
+++ b/connect/api_test.go
@@ -4,18 +4,14 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
-	"os"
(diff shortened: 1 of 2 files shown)


skazi0 (approved):
Nice. Thx!

PR #12, opened 2021-06-28
Add files to build rpm

JanZerebecki:


suseconnect-ng.spec
```diff
@@ -0,0 +1,76 @@
...
+%prep
+%setup -q -n connect-ng-%{version}
+
+%build
+find %_builddir/..
+%goprep %{import_path}
+find %_builddir/..
+go list -m all
+%gobuild cmd
+go build -buildmode=c-shared -o %_builddir/go/src/github.com/SUSE/connect-ng/ext/libsuseconnect.so ext/main.go
```
djoreilly, line 50: This extension library will not be needed for the time being. It's purpose in the PoC was to demonstrate how Ruby could call a library written in Go. It exposes no useful functions at the moment.

djoreilly (commented):
Looks ok to me, but I have no experience building RPMs for Go programs

suseconnect-ng.spec
JanZerebecki: Yes, the result is not used, I just build it for CI purposes.

djoreilly (approved):
lgtm.

skazi0 (approved):
lgtm, but I also haven't built any go packages so my experience is limited.

PR #14, opened 2021-06-29
Add text formatting util functions

skazi0:
Ruby logger has some functions to get bold and colored outputs.
Regular wrapper functions were added to mimic the original output.

### DIFF
--- a/connect/main.go
+++ b/connect/main.go
@@ -1,6 +1,7 @@
 package connect
 
 import (
+	"fmt"
 	"io"
 	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


connect/utils.go
```diff
@@ -0,0 +1,15 @@
+package connect
```
djoreilly, line 1: Can we put these in main.go instead? We already have some logging stuff there. That way we avoid clogging the directory with too many files.

connect/utils.go
```diff
@@ -0,0 +1,15 @@
...
+
+import "fmt"
+
+func logGreen(text string) string {
```
djoreilly, line 5: not sure about the "log" prefix, as technically it does not log, but just converts the text color. How about toGreen(), toRed() and toBold() ?

djoreilly (changes requested):

connect/utils.go
skazi0: I wanted to keep the naming from original code but these are not very important. I renamed to `greenText()` and `redText()`.

connect/utils.go
skazi0: moved

djoreilly:
From where will these be called? I think all the screen output we are doing is in cmd/suseconnect.go?  which is another package, which means we could have to call like connect.greenText() which is a bit unwieldy. Could just add these one-liners to the bottom of cmd/suseconnect.go instead?

djoreilly:
oh - I see them in client.rb and not cli.rb. This lgtm then

djoreilly (approved):

PR #15, opened 2021-06-30
Add Service struct and UpgradeProduct API

skazi0:
Service will hold the service description as returned by the API.
It also replaced local struct in Activation.

Product struct was modified to be used as PUT payload for added upgrade
API call.

### DIFF
--- a/connect/api.go
+++ b/connect/api.go
@@ -52,8 +52,21 @@ func GetProduct(productQuery Product) (Product, error) {
 		return remoteProduct, err
 	}
 	err = json.Unmarshal(resp, &remoteProduct)
+	return remoteProduct, err
+}
+
+func upgradeProduct(product Product) (Service, error) {
+	// NOTE: this can add some extra attributes to json payload which
+	//       seem to be safely ignored by the API.
+	payload, err := json.Marshal(product)
+	remoteService := Service{}
 	if err != nil {
-		return remoteProduct, err
+		return remoteService, err
 	}
-	return remoteProduct, nil
+	resp, err := callHTTP("PUT", "/connect/systems/products", payload, nil, authSystem)
+	if err != nil {
+		return remoteService, err
+	}
+	err = json.Unmarshal(resp, &remoteService)
+	return remoteService, err
 }
--- a/connect/service.go
+++ b/connect/service.go
@@ -1,0 +1,10 @@
+package connect
+
+// Service represents an installed service or service information from API
+type Service struct {
+	ID            int     `json:"id"`
+	URL           string  `json:"url"`
+	Name          string  `json:"name"`
+	Product       Product `json:"product"`
+	ObsoletedName string  `json:"obsoleted_service_name"`
+}
--- a/connect/activation.go
+++ b/connect/activation.go
@@ -11,20 +11,12 @@ type Activation struct {
 	Type      string    `json:"type"`
 	StartsAt  time.Time `json:"starts_at"`
 	ExpiresAt time.Time `json:"expires_at"`
-	Service   struct {
-		Product struct {
-			Name       string `json:"name"`
-			Version    string `json:"version"`
-			Arch       string `json:"arch"`
-			Identifier string `json:"identifier"`
-			Free       bool   `json:"free"`
-		} `json:"product"`
-	} `json:"service"`
+	Service   Service   `json:"service"`
 }
 
 func (a Activation) toTriplet() string {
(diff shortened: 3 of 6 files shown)


connect/service.go
```diff
@@ -0,0 +1,11 @@
+package connect
+
+// Service represents an installed service or service information from API
+type Service struct {
+	ID      int     `json:"id"`
+	URL     string  `json:"url"`
+	Name    string  `json:"name"`
+	Product Product `json:"product"`
+	// optional
```
djoreilly, line 9: what does optional mean here?

connect/service.go
```diff
@@ -0,0 +1,11 @@
+package connect
```
djoreilly, line 1: I wonder do we need a new file just for this struct.

djoreilly (changes requested):

connect/api.go
```diff
@@ -52,8 +53,21 @@ func GetProduct(productQuery Product) (Product, error) {
 		return remoteProduct, err
 	}
 	err = json.Unmarshal(resp, &remoteProduct)
+	return remoteProduct, err
+}
+
+func UpgradeProduct(product Product) (Service, error) {
```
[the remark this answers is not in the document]
skazi0: what about other functions (e.g. GetProduct())?

connect/api.go
```diff
@@ -52,8 +53,21 @@ func GetProduct(productQuery Product) (Product, error) {
...
+	// NOTE: this can add some extra attributes to json payload which
+	//       seem to be safely ignored by the API.
```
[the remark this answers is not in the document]
skazi0: No. This comes from the marshaling. for strings we could do "omitempty". For bools there will always be something added to json. We could do some more sophisticated interface based way of generating these jsons but for this case those extra attributes seem to make no difference.

connect/service.go
skazi0: It comes from the docs. I've never seen it missing from the response but maybe it could be.

connect/service.go
skazi0: It keeps the structure pretty predictable. We can put some utility functions for service here when needed.

connect/api.go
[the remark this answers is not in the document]
djoreilly: I think it should be getProduct. I think if the ruby counterpart is not called from yast.rb, then we should not export it

connect/api.go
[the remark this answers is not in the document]
djoreilly: ah ok fair enough

connect/service.go
djoreilly: ok. I don't think the comment is useful

connect/service.go
djoreilly: ok

connect/api.go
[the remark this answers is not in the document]
skazi0: How can I test it then? 
```
# github.com/SUSE/connect-ng/connect [github.com/SUSE/connect-ng/connect.test]
/home/skazi/SUSE/go-connect/connect/api_test.go:139:18: undefined: upgradeProduct
/home/skazi/SUSE/go-connect/connect/api_test.go:162:12: undefined: upgradeProduct
FAIL	github.com/SUSE/connect-ng/connect [build failed]
FAIL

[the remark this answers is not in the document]
Comment (skazi0) [reply connect/api.go]: I'll rename the other functions in a separate commit/PR.

File: connect/service.go (hunk 1 above)
Comment (skazi0) [reply connect/service.go]: removed

File: connect/api.go [hunk 5]
```diff
@@ -52,8 +52,21 @@ func GetProduct(productQuery Product) (Product, error) {
 		return remoteProduct, err
 	}
 	err = json.Unmarshal(resp, &remoteProduct)
+	return remoteProduct, err
+}
+
+func upgradeProduct(product Product) (Service, error) {
```
djoreilly, line 58: ok, so in client.go we can create a UpgradeProduct() which simply just calls to here. That's a convention we could follow and it is closer to the ruby. What do you think?

skazi0: do we need it now? or shall we add it when it needs to be used from outside the package?

djoreilly: I think deregister calls it. So we will need it for that
https://github.com/SUSE/connect/blob/master/lib/suse/connect/client.rb#L65

djoreilly (approved):

PR #24, opened 2021-07-02
Add announceSystem() api call

djoreilly:
The payload gets passed as a parameter, unlike the Ruby version.

### DIFF
--- a/connect/api.go
+++ b/connect/api.go
@@ -10,6 +10,24 @@ const (
 	APIVersion = "v4"
 )
 
+// announceSystem announces a system to SCC
+// https://scc.suse.com/connect/v4/documentation#/subscriptions/post_subscriptions_systems
+// The body parameter is produced by makeSysInfoBody()
+func announceSystem(body []byte) (string, string, error) {
+	resp, err := callHTTP("POST", "/connect/subscriptions/systems", body, nil, authToken)
+	if err != nil {
+		return "", "", err
+	}
+	var creds struct {
+		Login     string `json:"login"`
+		Passoword string `json:"password"`
+	}
+	if err = json.Unmarshal(resp, &creds); err != nil {
+		return "", "", err
+	}
+	return creds.Login, creds.Passoword, nil
+}
+
 func upToDate() bool {
 	// REVIST 404 case - see original
 	// Should fail in any case. 422 error means that the endpoint is there and working right
--- a/connect/api_test.go
+++ b/connect/api_test.go
@@ -7,6 +7,26 @@ import (
 	"testing"
 )
 
+func TestAnnounceSystem(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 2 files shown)


connect/api.go
```diff
@@ -10,6 +11,20 @@ const (
 	APIVersion = "v4"
 )
 
+func announceSystem(body []byte) (string, string, error) {
```
skazi0, line 14: Most of our API calls use some struct as parameter. Why not pass `hwinfo` and string hostname here?

connect/api.go
```diff
@@ -10,6 +11,20 @@ const (
...
+	resp, err := callHTTP("POST", "/connect/subscriptions/systems", body, nil, authToken)
+	if err != nil {
+		return "", "", err
+	}
+	m := make(map[string]string)
```
skazi0, line 19: I think it would look cleaner with small local struct with tags and unmarshal. Something like we have in `parseProductsXML`.

skazi0 (changes requested):
comments inline

connect/api.go
djoreilly: announce_system and update_system both build the exact same payload:
https://github.com/SUSE/connect/blob/master/lib/suse/connect/api.rb#L55-L82

This is even longer in Go. I have extracted that to one place
https://github.com/SUSE/connect-ng/pull/9/commits/17237d5f655ecbab8dc5df3ebf4c50b23e221a3d#diff-1fefdd8a034af76909710e05c9949c26981b14a4f065c9cec8b9b6811bc14d9cR72

skazi0: this could still be called inside `announceSystem()` and `updateSystem()` accordingly.

djoreilly: why?

skazi0: if you call `makeSysInfoBody()` inside these functions then you can pass relevant data to them instead of raw bytes.

djoreilly: sure. But makeSysInfoBody() makes calls to the system, and I want to unit test announceSystem()

skazi0: ah. the old mocking problem again. can you do it like with `var execCommand = exec.Command` ?

djoreilly: there are many calls. Why is receiving a slice header a problem here?

skazi0: because looking at the function you have no idea what kind of input it expects.

connect/api.go
djoreilly: done

connect/api.go
djoreilly: I added a comment. The caller AnnounceSystem() in client.go will be calling makeSysInfoBody()

skazi0: OK.

skazi0 (dismissed):

connect/api.go
skazi0: one more thing... maybe mention `makeSysInfoBody()` in the comment as well.

skazi0 (approved):

PR #26, opened 2021-07-02
Add Product.IsEmpty()

skazi0:
In original ruby code it was just 'if product', in golang we need to
make this test more explicit.


djoreilly (approved):
lgtm

PR #27, opened 2021-07-02
Add deactivateProduct() API call

skazi0:
This is used in both registration and deregistration.
Additional Product field was added to match original implementation.

Some behavior differences wrt original implementation were noted in
DIFFERENCES.md

### DIFF
--- a/connect/api_test.go
+++ b/connect/api_test.go
@@ -166,3 +166,67 @@ func TestUpgradeProductError(t *testing.T) {
 		}
 	}
 }
+
+func TestDeactivateProduct(t *testing.T) {
+	createTestCredentials("", "", t)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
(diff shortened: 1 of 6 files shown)


connect/api_test.go
```diff
@@ -166,3 +166,44 @@ func TestUpgradeProductError(t *testing.T) {
 		}
 	}
 }
+
+func TestDeactivateProductSMT(t *testing.T) {
+	createTestCredentials("", "", t)
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		w.Write(readTestFile("service_inactive_smt.json", t))
```
djoreilly, line 174: this file is missing

djoreilly (changes requested):

connect/api_test.go
skazi0: added

skazi0:
I added another test sample for SCC vs SMT comparison.

djoreilly (approved):
lgtm

PR #28, opened 2021-07-02
Make requirement for go 1.16 explicit.

JanZerebecki:
Make requirement for go 1.16 explicit.

This requirement is due to requiring the module embed.

The result is that it will only build on SLE_15_SP1, SLE_15_SP3
openSUSE_Factory, openSUSE_Leap_15.3, and openSUSE_Tumbleweed. This
removes the possibility to build on SLE_15 (SP0), and SLE_15_SP2 and
anything before SLE_15, and before openSUSE_Leap_15.3.

Also run go list all during the build without -m to be compatible with
earlier go versions to aid debugging and to also list all packages
instead of only modules.

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -28,6 +28,7 @@ Group:          System/Management
 Source:         connect-ng-%{version}.tar.xz
 Source1:        %name-rpmlintrc
 BuildRequires:  golang-packaging
+BuildRequires:  go >= 1.16
 
 %description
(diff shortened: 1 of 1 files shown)


djoreilly (approved):

JanZerebecki:
Note the reason the build on SP1 here worked is that https://build.opensuse.org/projects/systemsmanagement:SCC/meta includes devel:languages:go for SP1, otherwise it also wouldn't build there.

PR #29, opened 2021-07-02
obs: use an in between branch

JanZerebecki:
This branch as many repositories removed instead of only disabled to
make them not be reenabled on branches building PRs.

It also has the _service entries changed from manual to serveronly,
otherwise they won't be rerun by OBS.


JanZerebecki:
This will only have an effect once it is merged.

skazi0 (approved):

PR #31, opened 2021-07-05
Add zypper service commands needed for registration

djoreilly:
Adds addService(), refreshService() and enableServiceAutorefresh()

### DIFF
--- a/connect/zypper.go
+++ b/connect/zypper.go
@@ -109,6 +109,28 @@ func zypperDistroTarget() (string, error) {
 	return string(output), nil
 }
 
+func addService(serviceURL, serviceName string, refresh bool) error {
+	// Remove old service which could be modified by a customer
+	if err := removeService(serviceName); err != nil {
+		return err
+	}
+	args := []string{"--non-interactive", "addservice", "-t", "ris", serviceURL, serviceName}
+	_, err := zypperRun(args, true, []int{zypperOK})
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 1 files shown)


djoreilly:
need to add install_release_package

skazi0 (approved):
lgtm

connect/zypper.go
```diff
@@ -120,12 +142,48 @@ func removeService(serviceName string) error {
...
 	_, err := zypperRun(args, true, []int{zypperOK})
 	return err
 }
 
+func installReleasePackage(identifier string) error {
+	if identifier == "" {
+		return nil
+	}
+	args := []string{"rpm", "-q", identifier + "-release"}
+	if _, err := execute(args, false, nil); err != nil {
```
djoreilly, line 168: this is wrong - it should return if there is no error, ie the rpm is already installed

djoreilly: done

skazi0: ah, right, the go-style error handling is very error prone wrt negative tests. maybe rename this to `ok` to make it clear what we're up to?

djoreilly: no it was my mistake reading the ruby "unless" statement. I added a comment with the latest push

skazi0: `unless` is very error prone as well :o)
people are not good at inverse logic.

skazi0 (approved):
lgtm

PR #32, opened 2021-07-05
Always call zypper with --non-interactive arg

djoreilly:
Then it can be done one place.

### DIFF
--- a/internal/connect/zypper.go
+++ b/internal/connect/zypper.go
@@ -32,7 +32,7 @@ const (
 )
 
 func zypperRun(args []string, quiet bool, validExitCodes []int) ([]byte, error) {
-	cmd := []string{zypperPath}
+	cmd := []string{zypperPath, "--non-interactive"}
 	if CFG.FsRoot != "" {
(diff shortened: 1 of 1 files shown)


skazi0:
The `targetos` call doesn't have `--non-interactive` in original code. Have you checked if there's any reason for that?

djoreilly:
> The `targetos` call doesn't have `--non-interactive` in original code. Have you checked if there's any reason for that?

I have and it works with the same with --non-interactive. I cannot see a reason why it is different. I will check the git history and try to find if there is one.

djoreilly:
I checked the git history and `--non-interactive` was never used with `targetos` . I guess because it is not needed with `targetos` as the command does not have any possible interaction.

https://github.com/SUSE/connect/commit/e05e040cd2f809fb43464b4ba2499075dd78b4da#diff-947d0f660915c13c25028887223e8a202b48d34bb7f47c9780c013ce3b819562

djoreilly:
closing. The zypper migration plugin requires interactive sometimes

PR #35, opened 2021-07-07
Fix instance data file path

djoreilly:
It should be relative to config root. Also extract reading the
file to a function which will also be used by AnnounceSystem().


skazi0 (approved):
Lgtm

PR #42, opened 2021-07-09
Fix callHTTP() so connections are reused

djoreilly:
For connection reuse, the http.Client needs to be created
once and reused on subsequent requests. This is recommended
in "go doc http client".


skazi0 (dismissed):
needs to be rebased (DIFFERENCES.md) but the code works fine

djoreilly:
rebased

skazi0 (approved):

PR #44, opened 2021-07-12
Add rollback CLI option

skazi0:


skazi0:
I rebased and moved the code back to `connect` package. We can extract it to a separate one if needed in the future.

DIFFERENCES.md
```diff
@@ -17,3 +17,9 @@
   Go version will handle this as "API error" while ruby version will fail to
   parse the response and handle this as "JSON error".
 - The Go HTTP client tries to reuse connections with keep-alive.
+- Docs for GET APIs call for URL-encoded query params. Ruby version sends a
```
djoreilly, line 20: Is this related to rollback somehow?

connect/api.go
```diff
@@ -146,6 +150,24 @@ func deregisterSystem() error {
 	return err
 }
 
+func synchronizeProducts(products []Product) ([]Product, error) {
```
djoreilly, line 153: kinda long name, maybe syncProducts() or synchronize() like the Ruby version

suseconnect/usage.txt
```diff
@@ -40,6 +40,9 @@ Manage subscriptions at https://scc.suse.com
         --write-config       Write options to config file at /etc/SUSEConnect.
         --cleanup            Remove old system credentials and all zypper
                              services installed by SUSEConnect.
+        --rollback           Revert the registration state in case of a failed
+                             migration.
+
```
djoreilly, line 46: The Ruby version has just one blank line before `Common options:`

djoreilly (changes requested):
Just some small comments. Yes we can revisit the packaging later when needed.

DIFFERENCES.md
skazi0: I noticed this while comparing debug traces of old and new version. IIRC this was visible in `showProduct()` GET call.

connect/api.go
skazi0: `synchronize()` looks a bit more generic to me. `syncProducts()` is fine.

suseconnect/usage.txt
skazi0: good catch! fixed.

connect/api.go
skazi0: done

DIFFERENCES.md
djoreilly: ok, yes I checked and showProduct() is the only api call that uses toQuery(). I think we should specify the difference is between Go showProduct() and Ruby show_product(), because as it stands the reader would have to go compare the API doc, Go code and Ruby code, to find which call(s) this is about

DIFFERENCES.md
```diff
@@ -17,3 +17,9 @@
...
+  JSON query in body (like for other verbs). Go implementation follows docs.
+- The connect.synchronizeProducts() returns deserialized slice of Product
```
djoreilly, line 22: connect.syncProducts() - but not sure that this is a difference we need to document. The function is not exported.

djoreilly (changes requested):

DIFFERENCES.md
skazi0: added note about `showProduct()`

DIFFERENCES.md
skazi0: I'd rather keep it for developers as a sign of by-design change in behavior.

djoreilly (approved):
lgtm

PR #45, opened 2021-07-12
Fix list-extensions format

skazi0:
If there were more than one top-level extension, there were some
unwanted empty lines added. After this change the output should look
exactly like in ruby version.


djoreilly (approved):
lgtm

PR #51, opened 2021-07-16
S390 hwinfo

djoreilly:
Some fixes for s390 after testing.


internal/connect/hwinfo.go
```diff
@@ -100,8 +100,9 @@ func hwinfoS390(hw *hwinfo) error {
 	if err != nil {
 		return err
 	}
-	hw.UUID = string(rvuOut)
-	if hw.UUID == "" {
+	if !strings.HasPrefix(string(rvuOut), "help:") {
```
JanZerebecki, line 103: Please instead fix this in the uuid() function.

JanZerebecki (commented):
See the inline comment.

Could an integration test that runs during go test for the arch specific code have caught some of what this PR fixes? Manual testing is bothersome. E.g. a `:` is not a valid char in a UUID.

Otherwise looks good.

skazi0 (approved):
lgtm

djoreilly:
> Could an integration test that runs during go test for the arch specific code have caught some of what this PR fixes? Manual testing is bothersome. E.g. a `:` is not a valid char in a UUID.

Maybe the image running the integration tests has dmidecode installed? If so it would not see the error.

internal/connect/hwinfo.go
djoreilly: The way uuid is found for x86 and arm is completely different than s390. So I decided to put them in different functions. I have added some comments to make this more clear. Also now it checks for a valid uuid and returns "" for anything else.

skazi0 (dismissed):

internal/connect/hwinfo.go
```diff
@@ -190,6 +183,27 @@ func uuid() (string, error) {
...
+	if isUUID(uuid) {
+		return uuid
+	}
+	Debug.Print("Not implemented. Unable to determine UUID for s390x. Set to \"\"")
+	return ""
+}
+
+// isUUID returns true if s is a valid v4 uuid
+func isUUID(s string) bool {
+	exp := `^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-4[a-fA-F0-9]{3}-[8|9|aA|bB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$`
```
djoreilly, line 202: I think this regexp needs to be relaxed a bit. It's not uuid4, but RFC 4122. 
I think just `^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$` is enough

internal/connect/hwinfo.go
JanZerebecki: Looks good now.

djoreilly:
changed regexp for the uuid - it may not be a uuid4

JanZerebecki:
> > Could an integration test that runs during go test for the arch specific code have caught some of what this PR fixes? Manual testing is bothersome. E.g. a `:` is not a valid char in a UUID.
> 
> Maybe the image running the integration tests has dmidecode installed? If so it would not see the error.

Neither SLE nor openSUSE has dmidecode on s390, as the package demidecode is explicitly not built for that arch. Though maybe some other package provides that binary there, didn't check.

The cucumber tests from ruby connect do no run on any arch other than x86_64 at all.

The go tests, that are run during rpm build, could have integration tests as long as they do not require networking to another host. E.g. running the hwinfo in full, but not sending it to an scc would work. That is currently the only way where integration tests already run for s390.

JanZerebecki (approved):

skazi0 (approved):
lgtm

djoreilly:
> The go tests, that are run during rpm build, could have integration tests as long as they do not require networking to another host. E.g. running the hwinfo in full, but not sending it to an scc would work. That is currently the only way where integration tests already run for s390.

The hwinfo part can only be invoked during an announce or update, and both of these need to connect to the server to complete successfully.

JanZerebecki:
You are talking about running the cli. I'm talking about calling the internal function in a go test. Is there any reason against the later?

skazi0:
@JanZerebecki this would need to be some special type of test. Unit tests should work in a fixed environment and should not actually rely on the system where it's executed.

djoreilly:
@JanZerebecki currently I don't know a solution for that. I'll merge this PR now, but we should research that kind of testing.

PR #54, opened 2021-07-20
Simplify list-extensions template

skazi0:
The template was nested to mimic the original template used to print
tree structure of extensions. As go templates are more limited than
ruby's erb, some pre-formatting done in the go code.
With that in place, there's no point in having the template recursive
since all of the indentations are generated in go.
After this commit, pre-formatting returns flat list of extensions
prepared for printing and only some basic formatting is added in the
template. This makes the template easier to understand and more robust.


djoreilly (approved):
lgtm

PR #56, opened 2021-07-20
Try to use localized error from server response

djoreilly:
When the server sends an error response, use the "localized_error"
if provided, otherwise use the "error" key.


djoreilly:
https://github.com/SUSE/connect/blob/master/lib/suse/connect/errors.rb#L37

skazi0 (approved):
nice

JanZerebecki (approved):

PR #58, opened 2021-07-22
Add --version

JanZerebecki:


sayalilunkad:
lgtm

djoreilly (changes requested):
`version.txt` is not part of the repo, so test fails

$ go test ./internal/connect/
internal/connect/version.go:9:13: pattern version.txt: no matching files found

suseconnect/usage.txt
```diff
@@ -42,6 +42,7 @@ Manage subscriptions at https://scc.suse.com
                              services installed by SUSEConnect.
         --rollback           Revert the registration state in case of a failed
                              migration.
+        --version            Print program version.
```
djoreilly, line 45: In the Ruby SUSEConnect,  `--version` is between `--root` and `--debug` - but I'm not sure it really is a `Common option`

Makefile
```diff
@@ -3,24 +3,33 @@ all: test build build-so
...
+	       git log --no-show-signature -n1 --date='format:%Y%m%d' --pretty="format:$${parent}~git$${offset}.%h" > internal/connect/version.txt
+	cat -v internal/connect/version.txt
+	@echo
+
+build: out internal/connect/version.txt
 	go build -v -o out/ github.com/SUSE/connect-ng/suseconnect
 
-test:
+test: internal/connect/version.txt
 	go test -v ./internal/connect
```
djoreilly, line 18: should the file be removed after? Or should it be in gitignore?

internal/connect/version.go
```diff
@@ -0,0 +1,19 @@
...
+	_ "embed"
+	"strings"
+)
+
+var (
+	//go:embed version.txt
+	version        string
+)
+
+func GetFullVersion() string {
```
djoreilly, line 13: what uses this?

JanZerebecki: Nothing.

djoreilly:
Also we send the version in the `User-Agent` header in the calls to the server. So `addHeaders()` in `connection.go` needs to be changed to use `GetShortenedVersion()` and the `Version` can be removed from `main.go`

Makefile
JanZerebecki: Added to .gitignore.

suseconnect/usage.txt
JanZerebecki: Yes, as implemented it is not a common option but more a command, where only one of them will be executed. The ruby version also calls exit directly before making use of any other command:
https://github.com/SUSE/connect/blob/f874c2d0b801402f34b8d60eb75fbef40937603e/lib/suse/connect/cli.rb#L210

internal/connect/version.go
JanZerebecki: Removed, we can add it once we need it.

JanZerebecki:
> `version.txt` is not part of the repo, so test fails
> 
> $ go test ./internal/connect/
> internal/connect/version.go:9:13: pattern version.txt: no matching files found

You'll need to use make test.

> Also we send the version in the `User-Agent` header in the calls to the server. So `addHeaders()` in `connection.go` needs to be changed to use `GetShortenedVersion()` and the `Version` can be removed from `main.go`

Done.

djoreilly (approved):
lgtm

PR #59, opened 2021-07-23
Fix list-extensions printing "Not available" when using SCC

djoreilly:
Only SMT/RMT servers return the Available field in their responses.
SCC does not, so and the zero-value false was used. Fix this by
checking if SCC is being used and if so set Available to true.

### DIFF
--- a/internal/connect/extensions.go
+++ b/internal/connect/extensions.go
@@ -67,6 +67,9 @@ func preformatExtensions(extensions []Product, activations map[string]Activation
 
 	var ret []displayExtension
 	for _, e := range sorted {
+		if URLDefault() {
+			e.Available = true // Only SMT/RMT return this field
+		}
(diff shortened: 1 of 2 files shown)


JanZerebecki (approved):

JanZerebecki:
Thank you, that explains a part about URLDefault() I missed. This works as a workaround for now, but should be able to make a more correct solution.

I think it should be fixed during Unmarshal instead. There is another use of this in client.go and both would become easier to understand then.

We shouldn't really be hard coding a special case for a specific server implementation. We currently have the schema of Available wrong. It is true if missing and only false if explicitly false. To be able to do that we might need to create a custom Unmarshal implementation for Product.

djoreilly:
Agree. And we should add tests that use data like what comes from SCC.

PR #61, opened 2021-07-27
Add Requires that weren't explicit (bsc#1188646)

JanZerebecki:


suseconnect-ng.spec
```diff
@@ -32,6 +32,26 @@ BuildRequires:  go >= 1.16
...
+Requires:       coreutils
+# ExclusiveArch from this package
+%ifarch %ix86 ia64 x86_64 %arm aarch64
+Requires:       dmidecode
+%endif
+# ExclusiveArch from this package
+%ifarch s390x
+Requires:       s390-tools
+%endif
+Requires:       systemd
```
djoreilly, line 49: why is systemd required?

suseconnect-ng.spec
```diff
@@ -32,6 +32,26 @@ BuildRequires:  go >= 1.16
 Conflicts:      SUSEConnect
 Obsoletes:      SUSEConnect
 Provides:       SUSEConnect
+%if 0%{?fedora} || 0%{?rhel} || 0%{?centos_version}
```
djoreilly, line 35: can SUSECoonnect be installed on these OSs?

djoreilly (commented):
Just a couple of questions

suseconnect-ng.spec
JanZerebecki: Yes, in theory. Not sure if anyone did use it, maybe for an in place migration to a SUSE OS. I copied this from the ruby version.

suseconnect-ng.spec
JanZerebecki: hwinfo.go executes systemd-detect-virt which is in that rpm.

djoreilly (approved):
lgtm

PR #65, opened 2021-08-20
Add API call to get online migration paths

djoreilly:
This will be used by the zypper migration plugin.


JanZerebecki (approved):

skazi0:
I'll test this when I get to this point in my implementation.
FTR: you can see current state here: https://github.com/SUSE/connect-ng/pull/67
Feel free to comment there already.

PR #67, opened 2021-08-25
Add zypper migration plugin implementation.

skazi0:


internal/connect/zypper.go
```diff
@@ -199,3 +203,31 @@ func setReleaseVersion(version string) error {
...
+		flags = append(flags, "--quiet")
+	}
+	if version != "" {
+		flags = append(flags, "--releasever", version)
+	}
+	args = append(flags, args...)
+	output, err := zypperRun(args, quiet, []int{zypperOK})
+	// TODO: print output as it goes not post-factum and do it for all zypper calls!
+	if len(output) > 0 {
+		fmt.Print(string(output))
```
djoreilly, line 230: should this be the quiet logger instead of `fmt`?

internal/connect/api_test.go
```diff
@@ -250,3 +250,25 @@ func TestDeactivateProductError(t *testing.T) {
...
+	migrations, err := productMigrations(nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(migrations) != 2 {
+		t.Fatalf("len(migrations) == %d, expected 2", len(migrations))
+	}
+	newestBase := migrations[0][0].ToTriplet()
+	expected := "SLES/15.4/x86_64"
+	if newestBase != expected {
```
djoreilly, line 271: this part of the test was added to check the sort worked, but `productMigrations()` doesn't sort anymore

suseconnect/migration.go
```diff
@@ -0,0 +1,562 @@
...
+	// pass root to connect config
+	if fsRoot != "" {
+		connect.CFG.FsRoot = fsRoot
+		// if we update a chroot system, we cannot create snapshots of it
+		noSnapshots = true
+	}
+
+	if !isSnapperConfigured() {
+		noSnapshots = true
+		connect.Debug.Println("Snapper not configured")
```
djoreilly, line 110: maybe we could use shortcuts for these loggers? Something like
`var Debug = connect.Debug`

suseconnect/migration.go
```diff
@@ -0,0 +1,562 @@
...
+
+	// preprocess the migrations lists
+	migrations := make([]connect.MigrationPath, 0)
+	unavailableMigrations := make([]connect.MigrationPath, 0)
+	for _, m := range allMigrations {
+		mAvailable := true
+		for _, p := range m {
+			mAvailable = mAvailable && p.Available
+		}
+		// sort migrations to put already installed products last and base products first
```
djoreilly, line 210: could the sorting be done in function with some unit tests?

djoreilly (commented):
Some comments inline. I didn't get to review it all - hopefully will get more time later today.

internal/connect/main.go
```diff
@@ -17,10 +17,13 @@ var (
 	CFG Config
 	// Debug logger for debugging output
 	Debug *log.Logger
+	// QuietOut is used to simplify --quiet option
+	QuietOut *log.Logger
```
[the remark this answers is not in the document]
skazi0: hmmm... sounds good.

internal/connect/api_test.go
skazi0: ok. I'll remove it.

internal/connect/main.go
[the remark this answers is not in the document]
skazi0: no.
wait.
I added it here again to get the "Executing ..." output from zypper calls automatically if not quiet.
if we reverse the logic and name this logger "Info" this would mean that it's enabled by default unless `--quiet` is used.
this results in "Executing ..." being printed in `SUSEConnect` case for all zypper calls.
I think `QuietOut` is still a bit more clear if you consider above.

internal/connect/zypper.go
```diff
@@ -199,3 +203,31 @@ func setReleaseVersion(version string) error {
 	_, err := zypperRun(args, true, []int{zypperOK})
 	return err
 }
+
+// RefreshRepos runs zypper to refresh all repositories
+func RefreshRepos(version string, force bool, quiet bool, verbose bool, nonInteractive bool) error {
+	args := []string{"ref"}
+	flags := []string{}
```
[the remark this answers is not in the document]
skazi0: the idea was that `flags` come before `args` but it could be reworked to have only one list if you think it's not clear.

internal/connect/zypper.go
skazi0: this should not be needed at all. we have to find some way to nicely print the output as the comment says. this is very important for long runs like `zypper dup`.

suseconnect/migration.go
skazi0: done. for now with `Debug` and `QuietOut` but these could be unexported as they won't be used from outside.

suseconnect/migration.go
```diff
@@ -0,0 +1,566 @@
...
+	flag.IntVar(&migrationNum, "migration", 0, "")
+	flag.StringVar(&fsRoot, "root", "", "")
+
+	flag.Parse()
+	// this is only to keep the flag parsing logic simple and avoid double
+	// negations/negatives below
+	selfUpdate := !noSelfUpdate
+
+	if verbose {
+		connect.EnableDebug()
```
djoreilly, line 96: I don't see the original doing this. It adds a lot of detailed output. But I think we do need some way to enable debug logging on the suseconnect library.

internal/connect/zypper.go
skazi0: Yes, that's exactly what we need for zypper-migration-plugin. We might need to extend the current execute logic or make it somehow "pluggable" with different "executors".

internal/connect/main.go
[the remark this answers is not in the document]
skazi0: then the name `Info` can't be used as a direct replacement for `QuietOut`. maybe we'll find some better alternative later.

suseconnect/migration.go
skazi0: this was me trying to do a "clever trick" and map `--verbose` to `--debug` which is not directly present in `zypper-migration-plugin`. the other option would be to add something like `VerboseOut` to match the `QuietOut`.

suseconnect/migration.go
skazi0: done.

suseconnect/migration.go
djoreilly: This spews pages and pages of detailed stuff, and makes it very hard to see the messages from `--verbose`. Also my scrollback buffer did not go back far enough.

I think we should add a `--debug` option to the zypper migration plugin CLI to enable this level of logging.

internal/connect/main.go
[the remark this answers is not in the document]
djoreilly: If the name is changed to `Info`, the logger will still be disabled when `SUSEConnect` runs with the current logic. Anyway it's not so important - just `QuietOut` is a peculiar name.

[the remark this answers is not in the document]
skazi0: It would but I think it would be confusing. By default, most tools output "info" level messages without any additional options; so if we name it `Info` and make it not print anything by default it would be strange (at least to me).

internal/connect/zypper.go
djoreilly: Okay so the Ruby zypper migration plugin uses `system` to directly call zypper in several places, so that output is displays as it comes from zypper.

My preference would be to create a new function like zypperLongRun or something and use it for these cases. I'm not keen about doing clever stuff that makes the current zypperRun/execute() more complex

suseconnect/migration.go
skazi0: Should `--debug` imply `--verbose` ?

skazi0: I changed the `--verbose` flag to support multi-use. Single `--verbose` is like before. Double `--verbose --verbose` enables connect's debug output.

internal/connect/zypper.go
skazi0: Maybe we could use the new `execute()` func variable to change this functionality in zypper-migration case? Only two `snapper` calls use the backticks to grab output from the command. The rest is called via `system`. I'm not sure about the commands called inside connect package.

suseconnect/migration.go
```diff
@@ -1,5 +1,28 @@
 package main
 
+// TODO LIST
```
djoreilly, line 3: thanks - breaking down big blocks to smaller functions with structure should help

internal/connect/main.go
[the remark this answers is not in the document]
djoreilly: okay thanks. I understand now.

internal/connect/zypper.go
djoreilly: Or maybe we could just change the existing execute() to always use pipes, read the pipe, print (if asked) and accumulate the output in a buffer as it comes. Then return the output at the end.

suseconnect/migration.go
```diff
@@ -457,7 +367,7 @@ func isSnapperConfigured() bool {
 	return false
 }
 
-func checkSystemProducts() ([]connect.Product, error) {
+func checkSystemProducts(rollbackOnFailure bool, printInstalledProducts bool) ([]connect.Product, error) {
```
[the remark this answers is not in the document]
skazi0: Yeah, I got to the same idea when I looked at this code after I finished. There are currently only two cases where this code is used and they use mutually exclusive set of options so it might as well be restructured to have these parts outside.

suseconnect/migration.go
skazi0: Unfortunately `flag` doesn't seem to support GNU style args. There's https://github.com/spf13/pflag but I don't think it makes sense to add a dependency just for this corner case. I'll change it to `--debug` if you think it's more readable.

skazi0: done.

suseconnect/migration.go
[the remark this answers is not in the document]
skazi0: I moved the printing part out.
I left the rollback part in as I think it is kind of part of the product checking functionality (even if optional).

internal/connect/status.go
```diff
@@ -107,3 +107,32 @@ func getStatusText(statuses map[string]Status) (string, error) {
...
+	products, err := installedProducts()
+	if err != nil {
+		return products, err
+	}
+	installedIDs := make(map[string]struct{}, 0)
+	for _, prod := range products {
+		installedIDs[prod.ToTriplet()] = struct{}{}
+	}
+
+	if IsRegistered() {
```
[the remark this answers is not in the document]
skazi0: In original code this is something like `uniq(installed_products + activated_products)`.
I tried to implement the same logic here.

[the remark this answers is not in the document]
djoreilly: ah okay - I see that in status.rb

internal/connect/status.go
```diff
@@ -107,3 +107,32 @@ func getStatusText(statuses map[string]Status) (string, error) {
...
+
+// SystemProducts returns sum of installed and activated products
+// Products from zypper have priority over products from
+// activations as they have summary field which is missing
+// in the latter.
+func SystemProducts() ([]Product, error) {
+	products, err := installedProducts()
+	if err != nil {
+		return products, err
+	}
```
djoreilly, line 119: maybe `if !IsRegistered() { return products, nil }` here instead of the check+indent below?

skazi0: done

internal/connect/system.go
```diff
@@ -3,23 +3,56 @@ package connect
...
+}
+
 func execute(cmd []string, quiet bool, validExitCodes []int) ([]byte, error) {
 	Debug.Printf("Executing: %s Quiet: %v\n", cmd, quiet)
 	var stderr, stdout bytes.Buffer
 	comm := execCommand(cmd[0], cmd[1:]...)
-	comm.Stdout = &stdout
-	comm.Stderr = &stderr
+	if systemEcho {
+		comm.Stdout = newEchoBufferWriter(os.Stdout, &stdout)
```
djoreilly, line 45: ah nice - I missed that bit in the doc about goroutines reading pipes if these are not os.File.

But I think we could use`io.MultiWriter` and there should be no need for `echoBufferWriter` ?

djoreilly (changes requested):

internal/connect/system.go
skazi0: as far as I understand `io.MultiWriter` doesn't interleave the lines but concatenates whole outputs, e.g. first everything from stdout, then everything from stderr.

djoreilly: It does what echoBufferWriter does, but in a more general way.

https://github.com/golang/go/blob/50c69cc3a9dfe4f268eebcf3dcbf7f3a93635842/src/io/multi.go#L96-L99
https://github.com/golang/go/blob/50c69cc3a9dfe4f268eebcf3dcbf7f3a93635842/src/io/multi.go#L58-L60

I tried it and it worked
```
	comm.Stdout = io.MultiWriter(os.Stdout, &stdout)
	comm.Stderr = io.MultiWriter(os.Stderr, &stderr)
```

skazi0: Ah, you're right. I was confused by `io.MultiReader` which acts as I described.

skazi0: updated

djoreilly (commented):
Just some comments about zypperDupArgs

suseconnect/migration.go
```diff
@@ -0,0 +1,628 @@
...
+	flag.Visit(func(f *flag.Flag) {
+		// skip not wanted
+		if _, found := wanted[f.Name]; !found {
+			return
+		}
+		// special case (update var and skip flag)
+		if strings.Contains(f.Name, "allow-vendor-change") {
+			avc = !strings.Contains(f.Name, "no-")
+			return
+		}
```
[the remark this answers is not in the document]
skazi0: This is tricky. We can't get exact behavior as the original here. The problem is when someone passes both `--no-allow-vendor-change` and `--allow-vendor-change`. In original code the last argument is used but `flag.Visit()` loops over the args as they are defined not as they are passed so we need to give higher priority to one of these options.

suseconnect/migration.go
```diff
@@ -0,0 +1,628 @@
...
+		// multiArg? loop over values
+		if val, ok := f.Value.(*multiArg); ok {
+			for _, v := range *val {
+				args = append(args, "--"+f.Name, v)
+			}
```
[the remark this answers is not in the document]
skazi0: The previous `return`s are explicit "skip this flag" cases. Here we don't know if some more processing will be added after this `if` or not. I'd rather keep it that way to avoid future mistakes.

suseconnect/migration.go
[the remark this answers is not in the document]
djoreilly: ok, I see now. 

`flag.Visit()` loops in in lexicographical order, so `no-allow-vendor-change` will be last, and we want that to be the priority anyway. 

I think this does the same and maybe clearer?
```
		if f.Name == "allow-vendor-change" {
			avc = true
			return
		} else if f.Name == "no-allow-vendor-change" {
			avc = false
			return
		}
```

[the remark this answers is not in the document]
skazi0: I changed this part a bit. Is it clearer now?

suseconnect/migration.go
```diff
@@ -0,0 +1,628 @@
...
+	wanted := map[string]struct{}{"auto-agree-with-licenses": {}, "l": {},
+		"allow-vendor-change": {}, "no-allow-vendor-change": {},
+		"debug-solver": {}, "recommends": {}, "no-recommends": {},
+		"replacefiles:": {}, "details": {}, "download": {},
+		"download-only": {}, "from": {}, "repo": {}}
+
+	args := []string{}
+
+	// special case (from original)
+	// mutually exclusive: allow-vendor-change/no-allow-vendor-change
```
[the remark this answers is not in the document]
skazi0: added more explanation and a note in DIFFERENCES

skazi0:
rebased and updated wrt conflicting flag checking

suseconnect/migration.go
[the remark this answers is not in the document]
djoreilly: yes thanks

suseconnect/migration.go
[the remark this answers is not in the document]
djoreilly: ok makes sense

DIFFERENCES.md
```diff
@@ -26,3 +26,16 @@
   This is a bug and it's already fixed in the Go version.
 - With --debug the Go version sends all debug output to stderr. The Ruby
   version sends http debug to stderr, and other debug to stdout.
+- In zypper-migration plugin, bool flag arguments matching default value are
+  ignored. This changes behavior for cases where both positive and negative
+  flag is passed, e.g.: `zypper migration --verbose --no-verbose` disables
```
djoreilly, line 31: so this should not be possible anymore with the contradicts checker

internal/connect/system.go
```diff
@@ -3,18 +3,34 @@ package connect
...
+}
+
 // Assign function for running external commands to a variable so it can be mocked by tests.
 var execute = func(cmd []string, validExitCodes []int) ([]byte, error) {
 	Debug.Printf("Executing: %s\n", cmd)
 	var stderr, stdout bytes.Buffer
 	comm := exec.Command(cmd[0], cmd[1:]...)
-	comm.Stdout = &stdout
-	comm.Stderr = &stderr
+	if systemEcho {
```
djoreilly, line 27: so maybe we could add this here
`fmt.Fprintf(os.Stdout,"\nExecuting '%s'\n\n", strings.Join(cmd, " "))`

Then there would be no need for the `QuietOut.Printf` in zypper.go, and the QuietOut logger could be removed from the connect package, as it's specific to the zypper plugin

suseconnect/migration.go
```diff
@@ -0,0 +1,653 @@
...
+)
+
+// logger shortcuts
+var (
+	Debug      *log.Logger = connect.Debug
+	QuietOut   *log.Logger = connect.QuietOut
+	VerboseOut             = log.New(io.Discard, "", 0)
+)
+
+// this Value type is used to hold values of args which could be passed multiple times
```
djoreilly, line 48: implements the flag.Value interface ... used to hold ...

suseconnect/migration.go
```diff
@@ -0,0 +1,653 @@
...
+	// this is only to keep the flag parsing logic simple and avoid double
+	// negations/negatives below
+	selfUpdate := !noSelfUpdate
+
+	if verbose {
+		VerboseOut.SetOutput(os.Stdout)
+	}
+
+	if debug {
+		connect.EnableDebug()
```
djoreilly, line 126: Maybe instead of providing this option, we could just always configure `connect.Debug` to append to `/var/log/zypper-migration.log`. Then if something goes wrong, support should be able to determine why from the details of the zypper and api calls.

suseconnect/migration.go
```diff
@@ -0,0 +1,653 @@
...
+		connect.QuietOut.Println(msg)
+		service, err := connect.UpgradeProduct(p)
+		if err != nil {
+			return baseProductVersion, fmt.Errorf("%s: %v", msg, err)
+		}
+
+		if service.ObsoletedName != "" {
+			msg := "Removing service " + service.ObsoletedName
+			VerboseOut.Println(msg)
+			connect.MigrationRemoveService(service.ObsoletedName)
```
djoreilly, line 501: check error?

suseconnect/migration.go
```diff
@@ -0,0 +1,653 @@
...
+		//               break
+		//             end
+		//           end
+		//         end
+		//       end
+		//     end
+
+		msg = "Adding service " + service.Name
+		VerboseOut.Println(msg)
+		connect.MigrationAddService(service.URL, service.Name)
```
djoreilly, line 535: check error?

djoreilly (commented):
Some comments inline.

internal/connect/system.go
skazi0: hmm... you mean to print it only when `systemEcho` is enabled?

suseconnect/migration.go
skazi0: done

suseconnect/migration.go
skazi0: done

suseconnect/migration.go
skazi0: done

DIFFERENCES.md
skazi0: removed

internal/connect/system.go
djoreilly: yes

djoreilly: ok

internal/connect/zypper.go
```diff
@@ -244,3 +254,99 @@ func DistUpgrade(version string, quiet, verbose, nonInteractive bool, extraArgs
...
+func createZypperRestoreScript(scriptPath, tarballPath string, paths []string) error {
+	var data struct {
+		Paths   []string
+		Tarball string
+	}
+	// remove leading "/" from paths to allow using them from different root
+	for _, p := range paths {
+		data.Paths = append(data.Paths, strings.TrimLeft(p, "/"))
+	}
+	data.Tarball = strings.TrimLeft(tarballPath, "/")
```
djoreilly, line 300: so the restore script uses a relative path. But the backup always puts it in `/var/adm/backup/system-upgrade/repos.tar.gz`

djoreilly (changes requested):

suseconnect/migration.go
```diff
@@ -554,8 +557,8 @@ func migrateSystem(migration connect.MigrationPath) (string, error) {
 // returns fs_inconsistent flag
 func applyMigration(migration connect.MigrationPath, quiet, verbose, nonInteractive bool, dupArgs []string) (bool, error) {
 	fsInconsistent := false
-	// TODO
-	//   zypp_backup(options[:root] ? options[:root]: "/")
+
+	connect.ZypperBackup()
```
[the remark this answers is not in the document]
skazi0: true. the other one was there temporarily for testing the backup function. I removed it.

internal/connect/zypper.go
```diff
@@ -244,3 +254,99 @@ func DistUpgrade(version string, quiet, verbose, nonInteractive bool, extraArgs
...
+	if err != nil {
+		return err
+	}
+	// allow execution of script
+	os.Chmod(scriptPath, 0744)
+	return nil
+}
+
+// ZypperBackup creates backup of zypper configuration files
+func ZypperBackup() error {
```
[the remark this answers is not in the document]
skazi0: hmm. I don't know. `zypp` and `zypper` are used interchangeably in some contexts. I'd rather use "zypper" everywhere for consistency.

suseconnect/migration.go
```diff
@@ -379,9 +381,10 @@ func migrationMain() {
 	if err != nil {
 		QuietOut.Print("\nPerforming repository rollback...\n")
 
-		// TODO
 		// restore repo configuration from backup file
-		//     zypp_restore
+		// NOTE: original ignores failures of this command
```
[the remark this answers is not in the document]
skazi0: added

internal/connect/zypper.go
```diff
@@ -244,3 +254,99 @@ func DistUpgrade(version string, quiet, verbose, nonInteractive bool, extraArgs
...
+	_, err := execute(command, []int{0})
+
+	// tarball can contain sensitive data, so prevent read to non-root
+	// do it for sure even if tar failed as it can contain partial content
+	if fileExists(tarballPath) {
+		os.Chmod(tarballPath, 0600)
+	}
+
+	if err != nil {
+		return fmt.Errorf("failed to create backup")
```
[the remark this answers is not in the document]
skazi0: done. thx!

internal/connect/zypper.go
skazi0: Yeah, there are some bugs around `--root` in the original. I've already fixed one here: https://github.com/SUSE/zypper-migration/pull/14
I'll re-check it again.

skazi0: I updated the code to store both script and tarball under given `--root`. Tested and seem to be working fine.
I added a note in DIFFERENCES as this seems to be broken in the original code.

internal/connect/zypper.go
```diff
@@ -1,8 +1,13 @@
 package connect
```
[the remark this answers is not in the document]
skazi0: done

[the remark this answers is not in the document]
skazi0: I kept the template in a separate file. It's a bash script and you can get syntax coloring etc.

internal/connect/zypper.go
djoreilly: ok thanks

internal/connect/zypper.go
```diff
@@ -32,6 +33,11 @@ const (
 	zypperInfoReposSkipped    = 106 // Some repository had to be disabled temporarily because it failed to refresh
 )
 
+var (
+	//go:embed zypper-restore.tmpl
+	restoreTemplate string
```
djoreilly, line 38: this should be in `zypperbackup.go` where it's actually used

djoreilly (changes requested):

internal/connect/zypper.go
skazi0: done

suseconnect/migration.go
```diff
@@ -466,11 +468,106 @@ func sortMigrationProducts(m connect.MigrationPath, installedIDs connect.StringS
...
+		rightParts = append(rightParts, "0")
+	}
+
+	// lenghts are equal so we can use one index
+	for i := range leftParts {
+		var l, r int
+		var err error
+		// take and convert i-th part of left
+		if l, err = strconv.Atoi(leftParts[i]); err != nil {
+			panic(err)
```
djoreilly, line 495: not sure about panic() here. Would just printing a useful error message and exiting be better?

suseconnect/migration.go
```diff
@@ -466,11 +468,106 @@ func sortMigrationProducts(m connect.MigrationPath, installedIDs connect.StringS
...
+	repos, err := connect.Repos()
+	if err != nil {
+		return err
+	}
+	for _, availableProduct := range productPackages {
+		// skip non-obsolete products
+		if compareEditions(availableProduct.Edition, p.Edition()) >= 0 {
+			continue
+		}
+		// filter out "(System Packages)" and already disabled repos
```
djoreilly, line 525: what are "(System Packages)" ?

djoreilly (commented):
some comments in line

suseconnect/migration.go
skazi0: These are packages which don't belong to any repo. They can be manually installed or be left over after some upgrade.

skazi0: The `zypper` man page says this:
```
       The set of installed packages on a system is sometimes denoted as
       repository @System or System Packages. In contrast to available
       repositories providing packages which can be installed, @System
       provides packages which can only be deleted. Installed packages which
       are not also provided by at least one of the available repositories are
       often denoted as being unwanted, orphaned or dropped.

File: suseconnect/migration.go (hunk 29 above)
Comment (skazi0) [reply suseconnect/migration.go:495]: ruby version doesn't do any checking explicitly so I assume it would just die with default stack trace.

File: suseconnect/migration.go [hunk 31]
```diff
@@ -466,11 +468,106 @@ func sortMigrationProducts(m connect.MigrationPath, installedIDs connect.StringS
...
+	for len(rightParts) < len(leftParts) {
+		rightParts = append(rightParts, "0")
+	}
+
+	// lenghts are equal so we can use one index
+	for i := range leftParts {
+		var l, r int
+		var err error
+		// take and convert i-th part of left
+		if l, err = strconv.Atoi(leftParts[i]); err != nil {
```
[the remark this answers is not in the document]
skazi0: this is tricky question. this column can contain letters (for regular packages) but I don't think we have letters in *product* packages.

suseconnect/migration.go
djoreilly: ah okay thanks

suseconnect/migration.go
djoreilly: maybe print message and return 0 here, so cleaning up this repo is skipped?

internal/connect/zypper.go
```diff
@@ -275,3 +275,40 @@ func Repos() ([]Repo, error) {
...
+
+// Package holds package info as returned by `zypper search`
+type Package struct {
+	Name    string `xml:"name,attr"`
+	Edition string `xml:"edition,attr"` // VERSION[-RELEASE]
+	Arch    string `xml:"arch,attr"`
+	Repo    string `xml:"repository,attr"`
+}
+
+func parseSearchResultXML(xmlDoc []byte) ([]Package, error) {
```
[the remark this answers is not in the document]
skazi0: done

suseconnect/migration.go
```diff
@@ -150,41 +147,35 @@ func migrationMain() {
...
-		//         exit 1
-		//       end
-
-		//       print "\nRestarting the migration script...\n" unless options[:quiet]
-		//       exec $0, *save_argv
-		//     end
-		//     exit 1
-		//   end
+		if err := connect.PatchCheck(quiet, verbose, nonInteractive, false); err != nil {
+			if err.(connect.ZypperError).ExitCode >= 100 {
```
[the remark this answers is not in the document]
skazi0: I thought about this for a sec... but I would need to return error anyway for error cases, right?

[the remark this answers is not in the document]
skazi0: yeah, done that. recheck please

internal/connect/zypper.go
```diff
@@ -312,3 +312,21 @@ func DisableRepo(name string) error {
...
 	return err
 }
+
+// PatchCheck checks if there are any patches pending to be installed.
+// nil return value means that there are none
+// zypperInfoUpdateNeeded or zypperInfoSecUpdateNeeded indicate pending patches
+func PatchCheck(quiet, verbose, nonInteractive, noRefresh bool) error {
+	flags := zypperFlags("", quiet, verbose, nonInteractive, noRefresh)
+	args := append(flags, "patch-check", "--updatestack-only")
+	_, err := zypperRun(args, []int{zypperOK, zypperInfoUpdateNeeded, zypperInfoSecUpdateNeeded})
```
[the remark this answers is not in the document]
skazi0: done

suseconnect/migration.go
```diff
@@ -150,41 +147,36 @@ func migrationMain() {
...
-		//         exit 1
-		//       end
-
-		//       print "\nRestarting the migration script...\n" unless options[:quiet]
-		//       exec $0, *save_argv
-		//     end
-		//     exit 1
-		//   end
+		if pending, err := connect.PatchCheck(true, quiet, verbose, nonInteractive, false); err != nil {
+			fmt.Println(err)
```
djoreilly, line 155: maybe add a bit of context to the message, to help people find where it came from

suseconnect/migration.go
```diff
@@ -150,41 +147,36 @@ func migrationMain() {
...
+			os.Exit(1)
+		} else if pending {
+			// install pending updates and restart
+			// TODO: check error?
+			connect.Patch(true, quiet, verbose, nonInteractive, true)
+			// stop infinite restarting
+			// check that the patches were really installed
+			if pending, err := connect.PatchCheck(true, true, false, true, true); pending || err != nil {
+				fmt.Println("patch failed, exiting.")
```
djoreilly, line 164: Maybe `Patch self-update failed, exiting` and add the `err` if any

djoreilly, line 164: Maybe `Patch self-update failed, exiting` and add the `err` if any

suseconnect/migration.go
```diff
@@ -150,41 +147,36 @@ func migrationMain() {
...
+				os.Exit(1)
+			}
+			QuietOut.Print("\nRestarting the migration script...\n")
+			// this should replace current process with a new one but stop on error
+			// just in case
+			if err := syscall.Exec(os.Args[0], os.Args, []string{}); err != nil {
```
djoreilly, line 170: just thinking ... we only need to restart if the suseconnect binary was updated? Could we find that by running `suseconnect --version` and compare with ` GetShortenedVersion()`

suseconnect/migration.go
skazi0: done

suseconnect/migration.go
djoreilly: but as it stands, it does not say why it failed. Were there still pending updates, or was there an error, and if there was an error what was it.

suseconnect/migration.go
djoreilly: yeah - better not overthink this

suseconnect/migration.go
```diff
@@ -0,0 +1,731 @@
...
+		if interrupted {
+			return baseProductVersion, fmt.Errorf("%s: %v", msg, ErrInterrupted)
+		}
+	}
+	return baseProductVersion, nil
+}
+
+// containsProduct returns true if given slice of products contains one with given name.
+func containsProduct(products []connect.Product, name string) bool {
+	for i := range products {
```
djoreilly, line 594: I guess there is no need for an index here? Just `for _, p := range products ...`

suseconnect/migration.go
```diff
@@ -0,0 +1,731 @@
...
+// updates system records in SCC/SMT
+// adds/removes services to match target state
+// disables obsolete repos
+// returns base product version string
+func migrateSystem(migration connect.MigrationPath, forceDisableRepos bool) (string, error) {
+	var baseProductVersion string
+
+	for _, p := range migration {
+		msg := "Upgrading product " + p.FriendlyName
+		connect.QuietOut.Println(msg)
```
djoreilly, line 555: just `QuietOut.Println(msg)` ?

djoreilly (changes requested):
I had a read through the whole patchset and just noticed a couple of minor things.

suseconnect/migration.go
skazi0: done

djoreilly (approved):
lgtm - thanks!

PR #69, opened 2021-08-30
Use versioned obsoletes and provides

JanZerebecki:
The old versions had no epoch, use epoch 2 version 0 so that the
obsoleted packages still have enough space for newer versions that this
package still obsoletes.

This is to fix the warning:
W: self-obsoletion SUSEConnect obsoletes SUSEConnect
The package obsoletes itself.  This is known to cause errors in various tools
and should thus be avoided, usually by using appropriately versioned Obsoletes
and/or Provides and avoiding unversioned ones.

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -30,10 +30,10 @@ Source1:        %name-rpmlintrc
 BuildRequires:  golang-packaging
 BuildRequires:  go >= 1.16
 BuildRequires:  zypper
-Obsoletes:      SUSEConnect
-Provides:       SUSEConnect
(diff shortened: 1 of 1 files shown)


JanZerebecki:
I talked to Jiří and he said before submitting we first need to test if the upgrade works: install SUSEConnect and zypper-migration-plugin from 15-SP3 and then run 'zypper up' with the repo containing the new suseconnect-ng. If that works then we can submit.

JanZerebecki:
Frederic and Jiří talked about the rpmlint warning and they agreed that we should ignore it. Asked me to not do this change, thus closing.

PR #70, opened 2021-09-06
Simplify mocking of external commands

djoreilly:
Our execute() function encapsulates the calls to exec.Command
and exec.Run. So it's much easier to substitute it at test time.

We don't use testing.Parallel(), so there is no danger modifying
the execute package variable.

Also any test that invokes code that calls execute() should first
replace the execute function, so no tests should be calling the
original execute function. Therefore there is no need to restore
it at the end of a test that replaces it.

### DIFF
--- a/internal/connect/system.go
+++ b/internal/connect/system.go
@@ -8,19 +8,14 @@ import (
 	"strings"
 )
 
-var execCommand = exec.Command
-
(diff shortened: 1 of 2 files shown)


skazi0 (approved):
lgtm

PR #71, opened 2021-09-08
Remove quiet param from execute() and zypperRun()

djoreilly:
The quiet param was copied from the Ruby SUSEConnect in the rewrite.
But it seems pointless as callers ignore the returned output anyway.

Removing it improves readability.


skazi0 (approved):
lgtm

PR #72, opened 2021-09-10
Add basic string set implementation

djoreilly:
Go does not provide a set implementation, nor does its  standard
library. We use the common idiom of maps with strings as keys and
empty structs as values. But it's tedious to read and write.
This commit abstracts away the implementation details.

### DIFF
--- a/internal/connect/stringset.go
+++ b/internal/connect/stringset.go
@@ -1,0 +1,40 @@
+package connect
+
+// StringSet is used to implement a set of strings
+type StringSet struct {
+	m map[string]struct{}
+}
+
+// NewStringSet constructor for StringSet.
+// The set can be initialized with zero or more strings.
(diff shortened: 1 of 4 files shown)


internal/connect/stringset.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Delete s from set
+func (ss StringSet) Delete(s string) {
+	delete(ss.m, s)
+}
+
+// Contains returns true if s is in the set
+func (ss StringSet) Contains(s string) bool {
+	if _, ok := ss.m[s]; ok {
```
skazi0, line 33: can it be simplified to something like this?
```
_, ok := ss.m[s];
return ok
```

djoreilly: done

skazi0 (approved):
nice. thx!

PR #73, opened 2021-09-27
Allow loading Config from any path

djoreilly:
Currently it's hard coded to /etc/SUSEConnect.
YaST-Registration requires that it can be something different.
This commit allows that by separating out the setting of defaults
from the Load() method.


skazi0 (approved):
lgtm

PR #74, opened 2021-09-28
Add readme for YaST integration

djoreilly:


djoreilly:
More readable here https://github.com/djoreilly/connect-ng/tree/yast-integration/yast#readme

internal/connect/config_test.go
```diff
@@ -59,3 +59,24 @@ func TestSaveLoad(t *testing.T) {
...
+	}
+	expected := "regcode-42"
+	if c.Token != expected {
+		t.Errorf("got Token: %s, expected: %s", c.Token, expected)
+	}
+	expected = "jbloggs@acme.com"
+	if c.Email != expected {
+		t.Errorf("got Email: %s, expected: %s", c.Email, expected)
+	}
+	// check other fields were not touched
```
skazi0, line 78: maybe it would be good to also check overriding some existing field via json?

djoreilly: done

PR #75, opened 2021-10-05
Don't send unneeded Product fields to API

skazi0:
Product structure was split into two parts. BasicProduct with fields
needed to identify the product and Product which embeds BasicProduct
and adds more fields on top.
For API calls which only need the basic fields, BasicProduct part can be
used even if Product was passed.

### DIFF
--- a/internal/connect/api.go
+++ b/internal/connect/api.go
@@ -73,9 +73,7 @@ func showProduct(productQuery Product) (Product, error) {
 }
 
 func upgradeProduct(product Product) (Service, error) {
-	// NOTE: this can add some extra attributes to json payload which
-	//       seem to be safely ignored by the API.
-	payload, err := json.Marshal(product)
+	payload, err := json.Marshal(product.BasicProduct)
 	remoteService := Service{}
 	if err != nil {
 		return remoteService, err
@@ -96,16 +94,12 @@ func downgradeProduct(product Product) (Service, error) {
 
 func activateProduct(product Product, email string) (Service, error) {
 	var payload = struct {
-		Indentifier string `json:"identifier"`
-		Version     string `json:"version"`
-		Arch        string `json:"arch"`
+		BasicProduct
 		ReleaseType string `json:"release_type"`
 		Token       string `json:"token"`
 		Email       string `json:"email"`
 	}{
-		product.Name,
-		product.Version,
-		product.Arch,
+		product.BasicProduct,
 		product.ReleaseType,
 		CFG.Token,
 		email,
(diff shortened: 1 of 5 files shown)


internal/connect/product.go
```diff
@@ -5,13 +5,19 @@ import (
...
 	Arch    string `xml:"arch,attr" json:"arch"`
+}
+
+// Product represents an installed product or product information from API
+type Product struct {
+	BasicProduct
+	Release string `xml:"release,attr" json:"-"`
 	Summary string `xml:"summary,attr" json:"-"`
-	IsBase  bool   `xml:"isbase,attr" json:"-"`
+	IsBase  bool   `xml:"isbase,attr" json:"base"`
```
djoreilly, line 20: was changing the json tag intentional?

djoreilly (commented):
So this is just as optimization, as the api server just ignores those fields. Are there cases when having the extra fields make the payload significantly larger?

internal/connect/product.go
skazi0: it was but not for this PR :o)
removed.

skazi0:
This is more to match our payloads to API spec and original client than to save bandwidth.
The problem is just for bool fields as the rest is handled by "omitempty" tag.
That's why I created a separate PR for this to have some decision on weather we want this or not.

internal/connect/api.go
```diff
@@ -75,7 +75,7 @@ func showProduct(productQuery Product) (Product, error) {
 func upgradeProduct(product Product) (Service, error) {
 	// NOTE: this can add some extra attributes to json payload which
```
djoreilly, line 76: can this comment be removed now?

internal/connect/api.go
```diff
@@ -130,7 +126,7 @@ func activateProduct(product Product, email string) (Service, error) {
 func deactivateProduct(product Product) (Service, error) {
 	// NOTE: this can add some extra attributes to json payload which
```
djoreilly, line 127: can this comment be removed now?

internal/connect/product.go
```diff
@@ -5,11 +5,17 @@ import (
 	"strings"
 )
 
-// Product represents an installed product or product information from API
-type Product struct {
+// BasicProduct represents basic information on product, mostly for API requests
```
djoreilly, line 8: maybe this comment could also mention that BasicProduct is used instead of Product to avoid sending unneeded bool json or something

internal/connect/api.go
skazi0: done

internal/connect/api.go
skazi0: done

djoreilly:
We discussed this last week and decided the change is not necessary. Closing.

PR #88, opened 2021-11-03
Show subscription information if available

### LINKED ISSUE #84: Show subscription information if available
Hello,

I added changes the ruby `SUSEConnect` . I think it makes sense to inform you about these changes to keep feature parity.

I changed `SUSEConnect` to show subscription information if available not matter if the product is marked as "`free`". This means the output in `--status` and `--status-text` needs to be updated.

**Why?** Proxies do show all products as `free` therefore `SUSEConnect` never shows the subscription information but this is required by `supportconfig` to help our support engineers.

**Note:** There are cases when Proxies/SCC do return `null` for subscription information like `regcode`. Treat them as not activated with a subscription aka. not showing the subscription information.

pull request in question: https://github.com/SUSE/connect/pull/464

djoreilly:
Closes #84

### DIFF
--- a/internal/connect/status.go
+++ b/internal/connect/status.go
@@ -82,7 +82,7 @@ func buildStatuses(products []Product, activations map[string]Activation) []Stat
 		}
 		if activation, ok := activations[product.ToTriplet()]; ok {
 			status.Status = registered
-			if !activation.isFree() {
+			if activation.RegCode != "" {
 				status.Name = activation.Name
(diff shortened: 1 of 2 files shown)


djoreilly:
Needs testing on an RMT server that returns subscription information.

djoreilly:
> Needs testing on an RMT server that returns subscription information.

done

skazi0 (approved):
lgtm

PR #90, opened 2021-11-04
Allow de-register/de-activate a single product

### LINKED ISSUE #89: Allow deregister a single product when BYOS proxy instance

jesusbv:
- When running BYOS proxy instance,
  allow de-registration of a single product

This Fixes #89


djoreilly (approved):
lgtm

PR #91, opened 2021-11-08
Extract ReleaseType from zypper output

skazi0:
This is added to match original behavior for custom/OEM release types.

Original code change: https://github.com/SUSE/connect/pull/24/files#diff-947d0f660915c13c25028887223e8a202b48d34bb7f47c9780c013ce3b819562R83
Current version in Ruby: https://github.com/SUSE/connect/blob/master/lib/suse/connect/zypper/product.rb#L31

### DIFF
--- a/internal/connect/zypper.go
+++ b/internal/connect/zypper.go
@@ -2,11 +2,15 @@ package connect
 
 import (
 	"encoding/xml"
+	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 )
 
 const (
 	zypperPath = "/usr/bin/zypper"
+	oemPath    = "/var/lib/suseRegister/OEM"
 )
 
 const (
@@ -58,6 +62,26 @@ func installedProducts() ([]Product, error) {
 	return parseProductsXML(output)
 }
 
+// get first line of OEM file if present
+func oemReleaseType(productLine string) (string, error) {
+	if productLine == "" {
+		return "", fmt.Errorf("empty productline")
+	}
+	oemFile := filepath.Join(CFG.FsRoot, oemPath, productLine)
+	if !fileExists(oemFile) {
+		return "", fmt.Errorf("OEM file not found: %v", oemFile)
+	}
+	data, err := os.ReadFile(oemFile)
+	if err != nil {
+		return "", err
+	}
+	lines := strings.Split(string(data), "\n")
+	if len(lines) == 0 {
+		return "", fmt.Errorf("empty OEM file: %v", oemFile)
+	}
+	return strings.TrimSpace(lines[0]), nil
+}
+
 // parseProductsXML returns products parsed from zypper XML
 func parseProductsXML(xmlDoc []byte) ([]Product, error) {
 	var products struct {
@@ -65,6 +89,12 @@ func parseProductsXML(xmlDoc []byte) ([]Product, error) {
 	}
 	if err := xml.Unmarshal(xmlDoc, &products); err != nil {
 		return []Product{}, err
+	}
+	// override ProductType with OEM value if defined
+	for i, p := range products.Products {
+		if oemValue, err := oemReleaseType(p.ProductLine); err == nil {
+			products.Products[i].ReleaseType = oemValue
+		}
 	}
 	return products.Products, nil
 }
--- a/internal/connect/zypper_test.go
+++ b/internal/connect/zypper_test.go
@@ -1,6 +1,8 @@
 package connect
 
 import (
+	"os"
+	"path/filepath"
 	"testing"
 )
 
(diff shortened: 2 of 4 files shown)


internal/connect/zypper.go
```diff
@@ -58,6 +62,25 @@ func installedProducts() ([]Product, error) {
 	return parseProductsXML(output)
 }
 
+// get first file of OEM file if present
+func oemReleaseType(productLine string) (string, error) {
+	// NOTE: this returns oemPath for empty ProductLine,
+	//       original returned oemPath/
```
djoreilly, line 68: I don't think this comment is needed. I spent time trying to understand what it means, and then why it might be important.

internal/connect/zypper_test.go
```diff
@@ -17,6 +19,40 @@ func TestParseProductsXML(t *testing.T) {
 	}
 }
 
+func TestParseProductsXML_ReleaseType(t *testing.T) {
+	CFG.FsRoot = t.TempDir()
+	// write test OEM file
+	testRT := "SLES-OEM-TEST"
+	oemDir := filepath.Join(CFG.FsRoot, "/var/lib/suseRegister/OEM")
```
djoreilly, line 26: could oemPath be used here?

internal/connect/zypper.go
```diff
@@ -58,6 +62,25 @@ func installedProducts() ([]Product, error) {
 	return parseProductsXML(output)
 }
 
+// get first file of OEM file if present
```
djoreilly, line 65: I think this should be "first line"

djoreilly (changes requested):
just some comments inline

internal/connect/zypper.go
skazi0: true

internal/connect/zypper.go
djoreilly: I'm struggling to understand this comment. What does "this" refer to? I thought it was oemReleaseType(), but it should return "" if `productLine` is empty.  Is it the `filepath.Join` call?  Why does it matter that there is no trailing forward-slash?

djoreilly (changes requested):

internal/connect/zypper.go
skazi0: sorry, my comment from yesterday didn't get posted. hope it clarifies things a bit.

djoreilly: > when productline is empty
would something like this at the start make possible problems go away?
`if productLine == "" { return "", fmt.Error("productLine is empty") }`

> Go code doesn't add trailing slash so it should try to read the file if it existed instead of a directory. 
in that case `os.ReadFile()` will fail with `read /var/lib/suseRegister/OEM: is a directory`

djoreilly (changes requested):

skazi0:
I meant if `/var/lib/suseRegister/OEM` is a regular file for some reason and `productline` is empty. This is where the difference will show... but it's very small corner case I guess.
Anyway, additional pre-check sounds good.

internal/connect/zypper.go
```diff
@@ -66,6 +90,12 @@ func parseProductsXML(xmlDoc []byte) ([]Product, error) {
 	if err := xml.Unmarshal(xmlDoc, &products); err != nil {
 		return []Product{}, err
 	}
+	// override ProductType with OEM value if defined
+	for i, p := range products.Products {
+		if oemValue, err := oemReleaseType(p.ProductLine); err == nil {
+			products.Products[i].ReleaseType = oemValue
+		}
+	}
```
djoreilly, line 98: sorry, just realizing this now, but maybe it would be better if this was done at the end of `installedProducts()` instead of here. There are some existing tests for `parseProductsXML()` so they may now be invoking `oemReleaseType()` on the test system which is not ideal. Also this does not fit with the name of the function as it is nothing to do with parsing XML.

djoreilly (changes requested):

internal/connect/zypper.go
skazi0: In original this is done in the constructor of `SUSE::Connect::Zypper::Product` so I don't think that running this in tests is a problem.
There are some special cases placed in JSON unmarshaler maybe we should do the same for XML. I know that JSON version doesn't reach out to external resources but still this might be the most logical option.

djoreilly: it's not causing a problem now, but ideally tests would not be interacting with the system. I don't think adding this to a custom xml unmarshaler is more logical.

djoreilly (approved):
lgtm

internal/connect/zypper.go
skazi0: we do the same when testing calls which use credentials file. I think we're good as long as we use temp directories and not leave anything behind (which could be picked up by other tests).

PR #92, opened 2021-11-10
Print failed command on zypper error like the Ruby

djoreilly:
Reproduce by creating a lock file:
    echo 1 > /run/zypp.pid
    ./suseconnect -s


skazi0 (approved):
lgtm

PR #93, opened 2021-11-16
Add zypper search-packages plugin

skazi0:
port of https://github.com/jsrain/zypper-search-packages

### DIFF
--- a/internal/connect/stringset.go
+++ b/internal/connect/stringset.go
@@ -38,3 +38,12 @@ func (ss StringSet) Contains(s string) bool {
 func (ss StringSet) Len() int {
 	return len(ss.m)
 }
+
+// Strings returns all strings from the set as a slice
+func (ss StringSet) Strings() []string {
+	strs := make([]string, 0, len(ss.m))
+	for s := range ss.m {
+		strs = append(strs, s)
+	}
+	return strs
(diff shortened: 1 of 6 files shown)


skazi0:
NOTE: the "TEMP" commit is extracted to https://github.com/SUSE/connect-ng/pull/95. It should be removed from here when that PR is merged.

djoreilly (commented):
When testing the options from the man page, it seems `-d` and `-g` don't work the same as the ruby version.

skazi0:
@djoreilly `-d` might be affected by this: https://github.com/jsrain/zypper-search-packages/pull/6
I added `-g` which was missing.

djoreilly:
> @djoreilly `-d` might be affected by this: [jsrain/zypper-search-packages#6](https://github.com/jsrain/zypper-search-packages/pull/6) I added `-g` which was missing.

okay, does the `-d` need to be changed to `-s` in suseconnect/searchPackagesUsage.txt ?

skazi0:
> okay, does the `-d` need to be changed to `-s` in suseconnect/searchPackagesUsage.txt ?
good catch! updated

djoreilly (approved):
LGTM. Thanks

PR #94, opened 2021-11-16
Use DefaultTransport defaults + client timeout

djoreilly:
If starting with empty Transport, many settings are not properly
initialized (e.g. there are no timeouts set). Adding changes on top of
cloned DefaultTransport solves the problem.
In addition, default Client had no timeout set so non-responsive API
would hold the client forever.

### DIFF
--- a/internal/connect/connection.go
+++ b/internal/connect/connection.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"time"
 )
 
 type authType int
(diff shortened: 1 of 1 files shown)


djoreilly:
Looking into the test failure.

djoreilly:
`expected that command "SUSEConnect --url https://scc.suse.com -r ...." has exit status of "0", but has "15". (RSpec::Expectations::ExpectationNotMetError)`
From `suseconnect/suseconnect.go` the exitcode 15 must have come from a zypper error.

skazi0:
Didn't we have this "code 15" before?

djoreilly:
> Didn't we have this "code 15" before?

We had tests fail before and were unable to see a reason from the test console log. I can't remember if they had code 15 or not.

skazi0 (approved):

PR #95, opened 2021-11-18
Optional base product in package search

skazi0:
Original code set product default to Zypper.base_product. As Go doesn't
support default parameter values, explicit passing of empty Product will
now fall back to base product from zypper.

This is required by clients to avoid exporting `baseProduct()` function.


djoreilly (approved):
lgtm

PR #96, opened 2021-11-23
Fix OBS CI, it now requires target

JanZerebecki:


djoreilly (approved):
lgtm - but I'm not familiar with this.

JanZerebecki:
Works now as tested at https://github.com/SUSE/connect-ng/pull/97 . The existing PRs need a new commit or rebase for it to trigger.

PR #102, opened 2021-12-07
Remove dependency on systemd

skazi0:
This is only needed for detecting hypervisor on arm64. We don't want or
need to install systemd on all systems where SUSEConnect is used.

See https://github.com/SUSE/connect/pull/467 for discussion.

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -51,12 +51,12 @@ Requires:       dmidecode
 %ifarch s390x
 Requires:       s390-tools
 %endif
-Requires:       systemd
 Requires:       zypper
 # lscpu is only used on those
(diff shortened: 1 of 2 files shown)


suseconnect-ng.spec
```diff
@@ -51,7 +51,6 @@ Requires:       dmidecode
 %ifarch s390x
 Requires:       s390-tools
 %endif
-Requires:       systemd
```
JanZerebecki, line 54: Use Recommends instead of deleting it.

JanZerebecki (changes requested):

JanZerebecki (approved):

djoreilly (approved):

PR #103, opened 2021-12-08
Disable HTTP/2

skazi0:
Old implementation only uses HTTP/1.1. With HTTP/2 enabled requests to
SCC API sometimes fail with "http2: response body closed" error.

### DIFF
--- a/internal/connect/connection.go
+++ b/internal/connect/connection.go
@@ -86,6 +86,7 @@ func callHTTP(verb, path string, body []byte, query map[string]string, auth auth
 		tr := http.DefaultTransport.(*http.Transport).Clone()
 		tr.TLSClientConfig = &tls.Config{InsecureSkipVerify: CFG.Insecure}
 		tr.Proxy = proxyWithAuth
+		tr.ForceAttemptHTTP2 = false
 		httpclient = &http.Client{Transport: tr, Timeout: 60 * time.Second}
 	}
 	req, err := http.NewRequest(verb, CFG.BaseURL, bytes.NewReader(body))


JanZerebecki:
You said this was enabled in https://github.com/SUSE/connect-ng/pull/94 and it sometimes fails on POST to announce. Can you give the full URL it fails on? I want to run some tests.

skazi0:
This is the last part of debug trace I get when it fails:
```
POST /connect/systems/products HTTP/1.1
Host: scc.suse.com
User-Agent: SUSEConnect-ng/0.0.4
Content-Length: 121
Accept: application/json,application/vnd.scc.suse.com.v4+json
Accept-Language: pl_PL.utf8
Authorization: Basic <redacted>
Content-Type: application/json
Accept-Encoding: gzip

{"identifier":"SLES","version":"15.4","arch":"x86_64","release_type":"","token":"<redacted>","email":""}

SUSEConnect error: http2: response body closed

JanZerebecki (approved):

JanZerebecki:
Your failing request trace shows it is HTTP1.1? Which would mean this won*t help.

JanZerebecki:
Resolving it says it is an alias for an Amazon ELB, which means it is likely unwrapping https and reencoding requests. Maybe this was related the the recent AWS outage?

skazi0:
FTR, the root cause is a regression in go1.16.10 (backported from go1.17): https://github.com/golang/go/issues/49558
Upstream fix: https://github.com/golang/go/commit/5a93142cec6c4da528e12939398a615869361d5a

djoreilly:
> Your failing request trace shows it is HTTP1.1? Which would mean this won*t help.

But I think with the default transport an upgrade is attempted
https://github.com/golang/go/blob/a19e72cb89fd33e5bf1474887e267806f65b7a40/src/net/http/transport.go#L49
https://github.com/golang/go/blob/a19e72cb89fd33e5bf1474887e267806f65b7a40/src/net/http/transport.go#L278-L283

djoreilly:
So SCC server only supports HTTP 1.1 today?  If so, then there is no point trying to upgrade, so this patch probably makes things more reliable.

JanZerebecki:
No, SCC uses Amazon ELB which supports and prefers HTTP 2 over 1.1.

The build released in SLE Micro 5.1 is not affected as it was built with go1.16-1.16.7-1.23.1.
The last build in Factory was done with go1.16-1.16.7-1.3.
The build submitted for Factory pending acceptance is not affected as it was built with go1.17-1.17.4-1.1.
The build submitted for SLE SP4 pending acceptance is affected as it was built with go1.16-1.16.10-1.32.1.
Factory, SP3 and SP4 all have go1.17 available.
To fix it without waiting for a fixed 1.16 we could require go >= 1.17.

JanZerebecki:
Err scratch that there is no fixed go 1.17 yet either: https://github.com/golang/go/commit/ab175939e28df62fee09c33c7d9e008bff6b684c

JanZerebecki:
There is already a pending go MU, to update to go1.16.11: https://build.suse.de/package/show/SUSE:Maintenance:22069/go1.16.SUSE_SLE-15_Update
Looking at the dates go seems to make a minor release at the beginning of each month. Not sure if we should just wait or patch go or make a workaround.
Does the workaround in this PR work?
How irritating is this problem?

skazi0:
The workaround from https://github.com/SUSE/connect-ng/pull/104 was merged already. This makes the problem go away in non-debug mode which should be just enough for CI and staging stuff.
If go1.16.11 doesn't go out before the release, we can re-consider disabling HTTP/2 as proposed in this PR.
I don't want to do it if it's not needed because HTTP/2 seems to be preferred these days and if we disable it, it is likely to stay disabled for longer (i.e. until it is strictly required and somebody will find out that it was explicitly disabled).

djoreilly:
This PR is no longer needed. Closing.

PR #105, opened 2021-12-13
Use testing.Helper() in test helpers

djoreilly:
Now there is no need to use runtime.Caller() to find the caller.


skazi0 (approved):
Nice

PR #106, opened 2021-12-14
Main.go to logging.go

djoreilly:


internal/connect/connection.go
```diff
@@ -11,6 +11,10 @@ import (
 	"time"
 )
 
+const (
+	appName = "SUSEConnect-ng"
```
skazi0, line 15: isn't `config.go` a better place for this?

djoreilly: I don't think so. It's not used in `config.go`. It is used in `connection.go` only.  So it's easier for readers of `connection.go` to find the const declared at the top of the file. I also moved `APIVersion` into `connection.go` for the same reason.

PR #107, opened 2021-12-14
Simplify parsing JSON error response

djoreilly:
Use anonymous struct instead of empty interfaces and type assertions.


skazi0 (approved):
looks better

PR #108, opened 2021-12-17
Fix migration problems found in public cloud

skazi0:
Re-implementation of https://github.com/SUSE/zypper-migration/pull/9 and https://github.com/SUSE/zypper-migration/pull/10

**NOTE**: this doesn't include "products restore" part as it is not clear if file-based restore is a correct approach.

### DIFF
--- a/suseconnect/migration.go
+++ b/suseconnect/migration.go
@@ -548,12 +548,23 @@ func cleanupProductRepos(p connect.Product, force bool) error {
 	return nil
 }
 
+// checks if given service is provided by SUSE
+func isSUSEService(service connect.Service) bool {
+	return strings.Contains(service.URL, connect.CFG.BaseURL) ||
+		strings.Contains(service.URL, "plugin:/susecloud") ||
+		strings.Contains(service.URL, "plugin:susecloud") ||
+		strings.Contains(service.URL, "susecloud.net")
+}
+
 // updates system records in SCC/SMT
 // adds/removes services to match target state
 // disables obsolete repos
 // returns base product version string
 func migrateSystem(migration connect.MigrationPath, forceDisableRepos bool) (string, error) {
 	var baseProductVersion string
+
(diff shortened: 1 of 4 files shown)


suseconnect/migration.go
```diff
@@ -590,6 +601,19 @@ func migrateSystem(migration connect.MigrationPath, forceDisableRepos bool) (str
 		if interrupted {
 			return baseProductVersion, fmt.Errorf("%s: %v", msg, ErrInterrupted)
 		}
+		// mark OLD service as migrated to skip it from cleanup step below
+		migratedServices.Add(service.ObsoletedName)
+	}
+	// remove SUSE services which don't have migration available (bsc#1161891)
+	for _, s := range systemServices {
+		if isSUSEService(s) && !migratedServices.Contains(s.Name) {
+			msg := "Removing service " + s.Name
```
djoreilly, line 610: Can this message also give the reason, like  `"Removing service " + s.Name + " because it has no migration available"`. Then it will be different to the one above.

suseconnect/migration.go
```diff
@@ -590,6 +601,19 @@ func migrateSystem(migration connect.MigrationPath, forceDisableRepos bool) (str
...
+			VerboseOut.Println(msg)
+			err := connect.MigrationRemoveService(s.Name)
+			if err != nil {
+				return baseProductVersion, err
```
djoreilly, line 614: should this be
`return baseProductVersion, fmt.Errorf("%s: %v", msg, err)`

djoreilly (changes requested):
just a couple of comments about the logging

suseconnect/migration.go
skazi0: done

suseconnect/migration.go
skazi0: good catch!
done

djoreilly (approved):
LGTM

PR #110, opened 2021-12-22
Allow non-root users to see usage text

### LINKED ISSUE #109: Cannot display usage as a non-root user
```
./out/suseconnect -?
Root privileges are required to register products and change software repositories.
```

That's probably not the intended behavior.

skazi0:
Original version didn't allow that but it can be useful.

Closes #109


djoreilly (approved):
lgtm

PR #111, opened 2022-01-07
Switch from rubygem-ffi to fiddle

skazi0:
Fiddle is an FFI implementation which is included in standard Ruby.
Switching from FFI gem to Fiddle removes one dependency without loosing
any functionality.

### DIFF
--- a/yast/lib/suse/connect/yast.rb
+++ b/yast/lib/suse/connect/yast.rb
@@ -1,38 +1,39 @@
 require 'json'
-require 'ffi'
+require 'fiddle/import'
 require 'suse/toolkit/shim_utils'
 
 module Stdio
-  extend FFI::Library
-  ffi_lib FFI::Platform::LIBC
-  attach_function :free, [ :pointer ], :void
+  extend Fiddle::Importer
+  dlload 'libc.so.6'
+  typealias 'pointer', 'void*'
+  extern 'void free(pointer)'
 end
 
 module GoConnect
-  extend FFI::Library
-  ffi_lib 'suseconnect'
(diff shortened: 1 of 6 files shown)


yast/lib/suse/connect/yast.rb
```diff
@@ -1,38 +1,39 @@
...
-  attach_function :offline_system_migrations, [:string, :string, :string], :pointer
-  attach_function :upgrade_product, [:string, :string], :pointer
-  attach_function :synchronize, [:string, :string], :pointer
-  attach_function :system_activations, [:string], :pointer
-  attach_function :search_package, [:string, :string, :string], :pointer
+  extend Fiddle::Importer
+  dlload 'libsuseconnect.so'
+  typealias 'string', 'char*'
+
+  #callback type: void log_line(int, string)
```
lslezak, line 17: If it is not needed then I'd remove it completely, avoid dead code...

lslezak (approved):
Avoiding extra dependency is good! :+1: 
Thanks!

yast/lib/suse/connect/yast.rb
skazi0: This is not old code, just a plain comment. Fiddle doesn't support callback types so everything is passed as void pointer. Without this, future us would need to dig through the code to find out what is the expected function signature.

djoreilly (approved):
lgtm

yast/lib/suse/connect/yast.rb
lslezak: Ah, I overlooked that, I thought it is part of the previous ffi code...

PR #117, opened 2022-01-27
Exit with code 64 on connection refused like Ruby

djoreilly:
This was overlooked in the rewrite.
./suseconnect --url http://localhost:1234 -s; echo $?


skazi0 (approved):
lgtm

PR #118, opened 2022-02-04
Workaround system cert reloading after import (bsc#1195220)

skazi0:
This is a workaround for Golang's missing feature of reloading system certs.
The built in system certs pool is initialized only once on startup and there is no option for reloading it when new certs are installed.
The problem is known upstream and tracked at https://github.com/golang/go/issues/41888.
The cert_pool.go is almost 1:1 copy of original library code but exposes needed functionality. It can be removed when upstream fixes the problem.


djoreilly (approved):
LGTM

PR #119, opened 2022-02-04
Pass "insecure" to zypper addservice

skazi0:
In insecure mode zypper calls can still fail because of untrusted certs.
The cert checks can be disabled by adding ssl_verify=no argument to URLs.
Service adding was the first case where this was found. Maybe more will be needed.

Steps to reproduce:
1. set `insecure: true` in `/etc/SUSEConnect`.
2. `suseconnect --url https://migration-smt2.qa.suse.de/`

### DIFF
--- a/internal/connect/zypper.go
+++ b/internal/connect/zypper.go
@@ -3,6 +3,7 @@ package connect
 import (
 	"encoding/xml"
 	"fmt"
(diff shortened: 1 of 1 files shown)


djoreilly (approved):
LGTM

PR #121, opened 2022-02-10
Delegate free() calls back to Go (bsc#1195729)

skazi0:
Direct import from libc causes problems when custom malloc
implementation like jemalloc is used. Passing the pointer back to Go
seems to be the safest way to handle memory management as both
allocation and freeing is done on the same side and there is no explicit
imports.


djoreilly (approved):
LGTM

PR #123, opened 2022-03-02
Remove self from LD_PRELOAD (bsc#1196326)

skazi0:
The libsuseconnect library will remove itself from LD_PRELOAD variable
after loading. This will prevent child processes started from
libsuseconnect and/or from the host program from needlessly preloading
libsuseconnect again.

### DIFF
--- a/libsuseconnect/libsuseconnect.go
+++ b/libsuseconnect/libsuseconnect.go
@@ -12,7 +12,9 @@ import (
 	"errors"
 	"net"
 	"net/url"
+	"os"
 	"strconv"
+	"strings"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


djoreilly (approved):
LGTM

kobliha:
Thank you!

ancorgs:
Just for the record, this is a (not so nice) workaround for the unwanted side-effect of https://github.com/yast/yast-yast2/pull/1236, which is admittedly another workaround.

This should do the job for SLE-15-SP4, but ideally we should find a better solution that would save us from messing around with `LD_PRELOAD` (ie. that would allow us to revert this commit and the corresponding one in yast-yast2).

PR #129, opened 2022-04-12
go1.18 compatibility: export GOFLAGS="-buildvcs=false" 

### LINKED ISSUE #128: go1.18+ default -buildvcs=auto soft dependency on git or export GOFLAGS="-buildvcs=false"
New default behavior ([link](https://go.dev/doc/go1.18#go-version)) `-buildvcs=auto` in `go1.18` detects VCS data in sources e.g. `.git/`. If  active, `go tool` invocations including `go build` and `go test` will call vcs tools (typically git) to encode VCS info into the built binary or tool output.

Mitigation:
* In each RPM section where `go tool` will be called (`%build`, `%check`), add `export GOFLAGS="-buildvcs=false"`.

Follow up:
In `go1.18.1`,  `go test` will revert to `-buildvcs=off` behaviour per: 
* https://github.com/golang/go/issues/51767
* https://go-review.googlesource.com/c/go/+/393878/
* `%check` addition `export GOFLAGS="-buildvcs=false"` can be removed after `go1.18.1`

Alternatives:
  * Add BuildRequires dependency on git
  * go1.18+ 'go version -m ./binaryname' also provides build info

jfkw:
PR to RPM spec file for go1.18 compatibility: export GOFLAGS="-buildvcs=false" in sections %build and %check.

Closes #128.

After go1.18.1 a follow-up PR to #128 will remove the export in %check (go test).

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -30,6 +30,7 @@ Summary:        Utility to register a system with the SUSE Customer Center
 Group:          System/Management
 Source:         connect-ng-%{version}.tar.xz
 Source1:        %name-rpmlintrc
+BuildRequires:  git
 BuildRequires:  golang-packaging
 BuildRequires:  go >= 1.16
 BuildRequires:  zypper


djoreilly (dismissed):
LGTM

suseconnect-ng.spec
```diff
@@ -108,6 +108,15 @@ find %_builddir/..
...
 find %_builddir/..
+# go1.18+ default behavior has runtime soft dependency on git, hg, etc.
+# When VCS e.g. .git/ detected in sources 'go build', 'go list'
+# and related subcommands attempt to stamp/read vcs info.
+# As the VCS info is potentially useful for software supply chain,
+# prefer not to override defaults globally via golang-packaging at this time.
+# For now, suppress in each build section before go tool use,
+# or alternatively add BuildRequires dependency on git.
+# Also see 'go version -m ./binaryname' as alternative build info.
+export GOFLAGS="-buildvcs=false"
```
skazi0, line 119: This fails on go 1.17 with `go: parsing $GOFLAGS: unknown flag -buildvcs`

skazi0 (changes requested):
some conditions are needed around this to make it backward compatible

suseconnect-ng.spec
jfkw: A rebased commit has been pushed with conditional checks using `rpm_vercmp` and `go env GOVERSION` as the argument. This works for both older and current versions of go.

djoreilly:
> Maybe instead we can rm -rf .git in %prep.

or add git to BuildRequires? I think either way would be cleaner

skazi0 (approved):
lgtm

PR #131, opened 2022-05-03
WIP: fix migration against SMT (bsc#1198625)

djoreilly:
We are using the Product struct to unmarshall in the
POST /connect/systems/products/migrations call.

SMT add an "id" field with type string. SCC and RMT don't
return this field in the same call. But other places (yast?)
needs this field, but as type int.

This fixes it in the custom unmarshaller, but then a custom
marshaller was needed also.

Maybe better to create a Migration struct with just the fields we
need, then the migrations code can create []Product from that.

### DIFF
--- a/internal/connect/product.go
+++ b/internal/connect/product.go
@@ -26,7 +26,7 @@ type Product struct {
 	Extensions []Product `json:"extensions,omitempty"`
 
 	// these are used by YaST
-	ID           int    `json:"id"`
+	ID           int    `json:"-"` // handled by custom unmarshaller/marshaller
 	Description  string `xml:"description" json:"description,omitempty"`
 	EULAURL      string `json:"eula_url,omitempty"`
 	FormerName   string `json:"former_identifier,omitempty"`
@@ -38,10 +38,11 @@ type Product struct {
 }
 
 // UnmarshalJSON custom unmarshaller for Product.
-// Only SMT/RMT send the "available" field in their JSON responses.
-// SCC does not, and the default Unmarshal() sets Available to the
-// boolean zero-value which is false. This sets it to true instead.
+// Special decoding is needed for the Available, IsBase and ID fields.
 func (p *Product) UnmarshalJSON(data []byte) error {
+	// Only SMT/RMT send the "available" field in their JSON responses.
+	// SCC does not, and the default Unmarshal() sets Available to the
+	// boolean zero-value which is false. This sets it to true instead.
 	type product Product
 	prod := product{
 		Available: true,
@@ -49,6 +50,7 @@ func (p *Product) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &prod); err != nil {
 		return err
 	}
+
 	// migration paths contain is-base information as "base" attribute
 	// while we default to "isbase" for YaST integration.
 	// the rest of the SCC API uses `product_type="base"` instead.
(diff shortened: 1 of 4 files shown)


internal/connect/product.go
```diff
@@ -115,3 +129,27 @@ func (p Product) distroTarget() string {
...
+	idInt := struct {
+		ID int `json:"id"`
+	}{}
+	if err := json.Unmarshal(data, &idInt); err == nil {
+		return idInt.ID
+	}
+
+	// Decoding id as an int failed. Try to decode as a string.
+	idStr := struct {
+		ID string `json:"id"`
```
skazi0, line 146: can you do `ID int json:"id,string"` here and drop manual conversion below?

djoreilly: yep - that works - nice!

skazi0: another idea: make custom unmarshaller only for the ID field like here: https://go.dev/play/p/rityF2vTsFp?v=goprev.go?download=true

djoreilly: Thanks, I'll give that a try in another branch and see what it looks like. It's shorter, but adds a new type. Not sure which way is best.

djoreilly: > another idea: make custom unmarshaller only for the ID field like here: https://go.dev/play/p/rityF2vTsFp?v=goprev.go?download=true

https://github.com/djoreilly/connect-ng/commit/f742a5ab55046869af00918b633b3009c6c6b0ff

rsimai:
You probably already considered this while I can't find any comments: is SMT sending this field as string fine at all, when yast expects it to be int?

I guess it's fine if suseconnect is tolerant and corrects this, just wondering if there should be a bug or at least a note to the SMT/YaST people.

djoreilly:
> You probably already considered this while I can't find any comments: is SMT sending this field as string fine at all, when yast expects it to be int?
The id field from the POST...migrations call to SMT is not used - I don't know why SMT sends it - SCC and RMT don't send it in the same call. With the Ruby suseconnect it didn't matter, but with Go we are giving the fields types, and in other calls (even from SMT) the type is int for Product.ID.

> I guess it's fine if suseconnect is tolerant and corrects this, just wondering if there should be a bug or at least a note to the SMT/YaST people.
I guess we could fix SMT so it stops sending this field, and make people update their SMT if they hit this issue. Then we wouldn't have to carry this complication forever.

rsimai:
I frankly don't know the SMT maintenance and development situation. But if https://github.com/SUSE/smt is the repo and with no current submissions there it's probably best resolved (or worked around) in suseconnect.

djoreilly:
> I frankly don't know the SMT maintenance and development situation. But if https://github.com/SUSE/smt is the repo and with no current submissions there it's probably best resolved (or worked around) in suseconnect.

Agreed, The api calls are handled here on smt:
https://github.com/SUSE/smt/blob/dfdf264e248a487487a8135490e1f7f9687da34a/www/perl-lib/SMT/Rest/SCCAPIv4.pm#L42

I can't tell if this can be fixed on smt without breaking something else (yast?) that depends on it
https://github.com/SUSE/smt/blob/dfdf264e248a487487a8135490e1f7f9687da34a/www/perl-lib/SMT/Rest/SCCAPIv4.pm#L658

skazi0:
@djoreilly please rebase to include latest CI fixes

skazi0 (approved):
lgtm

PR #132, opened 2022-05-04
Some cleanups

skazi0:
* Remove redundant code + add comment
This was moved to setupHTTPClient().
Add missing comment for exported function.
* Add missing import
This is needed for _check_error().
*  Fix "VCS stamping" problem
Go1.18+ embeds some additional info if it finds VCS directories in sources (e.g. .git). If git metadata is broken or incomplete, the build can fail. Renaming .git to .git.bak keeps the metadata in tarball but "hides" it from the go commands.
Also remove "BuildRequires: git" which was supposed to solve above problem.

### DIFF
--- a/internal/connect/connection.go
+++ b/internal/connect/connection.go
@@ -97,13 +97,6 @@ func setupHTTPClient() {
 
 func callHTTP(verb, path string, body []byte, query map[string]string, auth authType) ([]byte, error) {
 	setupHTTPClient()
-	if httpclient == nil {
-		// use defaults from DefaultTransport
-		tr := http.DefaultTransport.(*http.Transport).Clone()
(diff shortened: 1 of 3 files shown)


_service
```diff
@@ -11,7 +11,6 @@
     <param name="versionrewrite-pattern">v(\d+\.\d+\.\d+)</param>
     <param name="versionrewrite-replacement">\1</param>
     <param name="changesgenerate">enable</param>
-    <param name="package-meta">yes</param>
```
JanZerebecki, line 14: The idea is to have a copy of the history, so I'd prefer to rename the .git during build. So instead add to the end of %prep something like:
mv .git .git-deactivated

JanZerebecki (changes requested):

_service
skazi0: done. this worked too.

JanZerebecki (approved):

PR #135, opened 2022-05-27
Add option to run local scc tests

skazi0:
This will run cross-tests from ruby-connect using suseconnect-ng
replacement parts similar to what is done in the CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 out
 internal/connect/version.txt
+connect-ruby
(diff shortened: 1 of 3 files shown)


integration/run.sh
```diff
@@ -5,6 +5,8 @@ then
   exit 1
 fi
 
+git config --global --add safe.directory $(pwd)
```
skazi0, line 8: this seems to be needed also for some gating runs: https://ci.suse.de/blue/organizations/jenkins/scc-connect-ng-prs/detail/main/57/pipeline/

JanZerebecki (approved):

PR #137, opened 2022-06-23
Print nested zypper errors (bsc#1200803)

skazi0:
If zypper fails with no output and no meaningful exit code, sometimes the nested error has some important details.

Example: '/usr/bin/zypper: no such file or directory' was returned as `Error: zypper returned -1 with ''` now it will be `Error: zypper returned -1 with '' (fork/exec /usr/bin/zypper: no such file or directory)`

### DIFF
--- a/internal/connect/errors.go
+++ b/internal/connect/errors.go
@@ -36,11 +36,12 @@ type ZypperError struct {
 	Commmand []string
 	ExitCode int
 	Output   []byte
+	Err      error
 }
 
 func (ze ZypperError) Error() string {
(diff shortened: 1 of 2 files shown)


internal/connect/errors.go
```diff
@@ -36,11 +36,12 @@ type ZypperError struct {
 	Commmand []string
 	ExitCode int
 	Output   []byte
+	Err      error
```
mvidner, line 39: This makes ZypperError the same inside as ExecuteError. I'm fine with it as long as it helps debug https://bugzilla.suse.com/show_bug.cgi?id=1200803 , but you may want to merge the types or add a comment why they stay separate.
Also, I don't know Golang.

PR #138, opened 2022-07-19
Added support for the System-Token header

### LINKED ISSUE #125: Feature request: handle the `System-Token` header
This header is meant to be saved on the credentials file (`/etc/zypp/credentials.d/SCCcredentials`), and each response from scc.suse.com will give you back a new value that is to be saved.

Take a look at https://github.com/SUSE/connect/pull/478 for more info.

mssola:
## Description

In order to detect system duplicates, SCC implemented a `System-Token` mechanism in which systems are to report a token that is generated for every request from systems.

The Ruby implementation of SUSEConnect stores this into the credentials file for every request. If the response did not contain any token (e.g. first time that a system contacts SCC or upgrading from and older SUSEConnect) then an empty token is stored into the credentials file. All in all, this commit brings this Go implementation on par with the Ruby one when it comes to detecting system duplicates.

## How to test this

1. Build with the latest changes.
2. Run any request (e.g. the recently added `--keepalive` command)
3. Check that the system credentials have changed by adding a new `system_token` value.
4. Check that the system from Glue's database has updated its `system_token` column.

Retry steps 2, 3 and 4 multiple times and check that the `system_token` field keeps being in sync.

## Related issues

Fixes #125

### DIFF
--- a/internal/connect/api.go
+++ b/internal/connect/api.go
@@ -14,13 +14,13 @@ func announceSystem(body []byte) (string, string, error) {
 		return "", "", err
 	}
 	var creds struct {
-		Login     string `json:"login"`
-		Passoword string `json:"password"`
+		Login    string `json:"login"`
+		Password string `json:"password"`
 	}
 	if err = json.Unmarshal(resp, &creds); err != nil {
 		return "", "", JSONError{err}
 	}
-	return creds.Login, creds.Passoword, nil
+	return creds.Login, creds.Password, nil
 }
 
 func upToDate() bool {
--- a/internal/connect/api_test.go
+++ b/internal/connect/api_test.go
@@ -8,8 +8,11 @@ import (
 )
 
 func TestAnnounceSystem(t *testing.T) {
+	createTestCredentials("", "", t)
+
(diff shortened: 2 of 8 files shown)


mssola:
[[URAS] Add `System-Token` handling in SUSEConnect-ng](https://trello.com/c/9AT8OLtB/2458-uras-add-system-token-handling-in-suseconnect-ng)

djoreilly (dismissed):
LGTM

internal/connect/api.go
```diff
@@ -8,19 +8,19 @@ import (
 // announceSystem announces a system to SCC
 // https://scc.suse.com/connect/v4/documentation#/subscriptions/post_subscriptions_systems
 // The body parameter is produced by makeSysInfoBody()
-func announceSystem(body []byte) (string, string, error) {
+func announceSystem(body []byte) (string, string, string, error) {
```
skazi0, line 11: Is there any path which would return non-empty token from `announceSystem()` and/or `AnnounceSystem()`? If not, then I don't think we need an additional return value. Caller of these functions should take care of this.

mssola: Good point! Done 👍🏻

ngetahun (approved):
LGTM!

Issue #122, opened 2022-02-17
Feature request: systemd timer for `SUSEConnect --keepalive`


mssola:
Related to #120, we need to ensure that clients periodically report their status. That is why we have added a systemd timer that is to be installed, enabled and ensured to be active whenever someone installs/upgrades the RPM package.

You can take a look at these changes on the Ruby implementation here: https://github.com/SUSE/connect/pull/474. This is still ongoing, so feel free to also review it and give us 2¢ 😉

PR #140, opened 2022-07-27
Add systemd timer for SUSEConnect --keepalive

skazi0:
port of https://github.com/SUSE/connect/pull/474

added parts of https://github.com/SUSE/connect/pull/484 and https://github.com/SUSE/connect/pull/486 for user notice about keepalive feature

Closes https://github.com/SUSE/connect-ng/issues/122

### DIFF
--- a/suseconnect-keepalive.service
+++ b/suseconnect-keepalive.service
@@ -5,6 +5,7 @@ Wants=suseconnect-keepalive.timer
 [Service]
 Type=oneshot
 ExecStart=/usr/bin/SUSEConnect --keepalive
+EnvironmentFile=-/etc/sysconfig/proxy
 
 # Filter out this special exit code, since it will be raised by `SUSEConnect
 # --keepalive` whenever a system is not registered, which is a pretty typical
--- a/suseconnect-keepalive.service
+++ b/suseconnect-keepalive.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Run SUSEConnect --keepalive
+Wants=suseconnect-keepalive.timer
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/SUSEConnect --keepalive
+
+# Filter out this special exit code, since it will be raised by `SUSEConnect
+# --keepalive` whenever a system is not registered, which is a pretty typical
+# scenario.
+SuccessExitStatus=71
+
+[Install]
+WantedBy=multi-user.target
--- a/suseconnect-keepalive.service
+++ b/suseconnect-keepalive.service
@@ -11,6 +11,3 @@ EnvironmentFile=-/etc/sysconfig/proxy
 # --keepalive` whenever a system is not registered, which is a pretty typical
 # scenario.
 SuccessExitStatus=71
-
-[Install]
-WantedBy=multi-user.target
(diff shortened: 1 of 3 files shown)


suseconnect-keepalive.service
```diff
@@ -0,0 +1,15 @@
...
+Type=oneshot
+ExecStart=/usr/bin/SUSEConnect --keepalive
+
+# Filter out this special exit code, since it will be raised by `SUSEConnect
+# --keepalive` whenever a system is not registered, which is a pretty typical
+# scenario.
+SuccessExitStatus=71
+
+[Install]
+WantedBy=multi-user.target
```
JanZerebecki, line 15: Is this what is wanted, that it should also be started every boot?

skazi0: this is copied from the original version: https://github.com/SUSE/connect/blob/master/package/suseconnect-keepalive.service#L15

JanZerebecki: I know and it is not a blocker, but that doesn't answer the question. Is it also a bug in the original version?

skazi0: Then lets wait for an answer from the authors of original.

suseconnect-ng.spec
```diff
@@ -131,10 +131,62 @@ install -D -m 644 %_builddir/go/src/%import_path/man/SUSEConnect.8 %buildroot/%_
...
 find %_builddir/..
 # we currently do not ship the source for any go module
 rm -rf %buildroot/usr/share/go
 
+%pre
+%service_add_pre suseconnect-keepalive.service suseconnect-keepalive.timer
+
+# in pre blocks the old version is still installed. This way we can detect
+# if --keepalive was already present before
+kainfo=0
```
skazi0, line 148: this part is different than the original because it needs to handle various upgrade scenarios (e.g. SUSEConnect -> suseconnect-ng).

suseconnect-keepalive.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Run SUSEConnect --keepalive
+Wants=suseconnect-keepalive.timer
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/SUSEConnect --keepalive
+
```
mssola, line 8: Notice that there is a bug open on our side in which we are not considering environment variables. As it has been suggested, adding the following `EnvironmentFile=-/etc/sysconfig/proxy` would do the trick.

mssola: See bugzilla report here: https://bugzilla.suse.com/show_bug.cgi?id=1200994

skazi0: added

suseconnect-keepalive.service
digitaltom: That looks like an unwanted side effect. The plan was only to run the service by the timer.

JanZerebecki: Then we can remove this line and the one above.

JanZerebecki (approved):
I certainly wouldn't want to run this (and even do not want zypp/AnonymousUniqueId), but code looks fine.

mssola (approved):

PR #143, opened 2022-08-15
WIP Test building with boringcrypto

djoreilly:
go1.19 lets you set an env var to build with boringcrypto
https://github.com/golang/go/blob/dev.boringcrypto/README.boringcrypto.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ internal/connect/version.txt:
 	@echo
 
 build: out internal/connect/version.txt
-	go build -v -o out/ github.com/SUSE/connect-ng/suseconnect
+	GOEXPERIMENT=boringcrypto go build -v -o out/ github.com/SUSE/connect-ng/suseconnect
 
 test: internal/connect/version.txt
 	go test -v ./internal/connect ./suseconnect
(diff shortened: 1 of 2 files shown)


djoreilly:
Just a test to see if this builds and passes the CI tests.

JanZerebecki:
Darragh noticed that https://build.opensuse.org/package/view_file/devel:languages:go/go1.19/go1.19.spec?expand=1 deletes all `*.syso` files before go is built. The missing `goboringcrypto_linux_amd64.syso` seems to cause this failure, interestingly other arches still succeed. To fix this we should build it after deleting the version that is included in the source. Calling src/crypto/internal/boring/build.sh should do that. I'm currently trying that.

---

I checked if there is a way to switch on and off the fips mode of boringssl during runtime: There is not, as https://github.com/golang/go/blob/master/src/crypto/internal/boring/build.sh#L34 sets the mode during its build.

So we need to answer another question: Do we need to provide two binaries, an additional weakcrypto variant for fips mode?

JanZerebecki:
So my stab at readding boring in our package shows that I would need to adapt their build script to work inside the restricted permissions: https://build.opensuse.org/package/show/home:jzerebecki:branches:devel:languages:go/go1.19

Meanwhile we decided to wait for a patched version of go that can link to the system openssl instead of a static boringssl.

PR #144, opened 2022-08-23
Strip the binaries & use standard buildconditionals for hwinfo tests

skazi0:
This is based on https://build.opensuse.org/request/show/994730


JanZerebecki:
While it was not always safe to strip go binaries ( https://en.opensuse.org/index.php?title=openSUSE:Packaging_Go&oldid=168773#%go_nostrip ) that seems to have changed (link found by Darragh): https://honnef.co/posts/2016/10/go-and-strip/

JanZerebecki (approved):

djoreilly (approved):

PR #146, opened 2022-09-05
Respect the PROXY_ENABLED environment variable

mssola:
## Description

This is a special environment variable set in SUSE systems and that is managed by tools like YAST. The usage of this environment variable is not standard and hence the net/http/httputil library does not support it.

Therefore, this commit adds an extra check for this environment variable when setting up the HTTP client.

See SUSE/connect@b2758adacf2092cbbef8b8286c201a7d1ba5e7e1

## How to test it

1. Build with this patch applied.
2. Check that the environment variable `PROXY_ENABLED` is respected. That is, even if `HTTP_PROXY` has been set, setting `PROXY_ENABLED` to `false` will make SUSEConnect to ignore the setup.
3. Note that not setting it at all (e.g. leave a blank value or not setting it at all) will tell SUSEConnect to go on as usual.

### DIFF
--- a/internal/connect/connection.go
+++ b/internal/connect/connection.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 )
@@ -81,7 +82,23 @@ func successCode(code int) bool {
 	return code >= 200 && code < 300
 }
 
+// Returns true if proxy setup is enabled at the system level. This is specific
+// to SUSE.
+func proxyEnabled() bool {
+	value := strings.ToLower(strings.TrimSpace(os.Getenv("PROXY_ENABLED")))
+
+	// NOTE: if the value is not set, we return true so Go figures this out.
+	return value == "" || value == "y" || value == "yes" || value == "t" || value == "true"
+}
+
 func proxyWithAuth(req *http.Request) (*url.URL, error) {
+	// Check for the special "PROXY_ENABLED" environment variable which might be
+	// set in a SUSE system. If it is set to a falsey value, then we skip proxy
+	// detection regardless of other environment variables.
+	if !proxyEnabled() {
+		return nil, nil
+	}
+
 	proxyURL, err := http.ProxyFromEnvironment(req)
 	// nil proxyUrl indicates no proxy configured
 	if proxyURL == nil || err != nil {


internal/connect/connection.go
```diff
@@ -81,7 +82,22 @@ func successCode(code int) bool {
 	return code >= 200 && code < 300
 }
 
+// Returns true if proxy setup is enabled at the system level. This is specific
+// to SUSE.
+func proxyEnabled() bool {
+	value := strings.ToLower(strings.TrimSpace(os.Getenv("PROXY_ENABLED")))
+
+	return value == "" || value == "y" || value == "yes" || value == "t" || value == "true"
```
skazi0, line 90: take a look at https://pkg.go.dev/strconv@go1.17#ParseBool

skazi0 (changes requested):

mssola:
[[URAS] [connect-ng] Make sure that the `PROXY_ENABLED` environment variable is respected](https://trello.com/c/ZTT2ferN/2531-uras-connect-ng-make-sure-that-the-proxyenabled-environment-variable-is-respected)

internal/connect/connection.go
mssola: Good point!

mssola: Done!

skazi0 (approved):
lgtm

internal/connect/connection.go
```diff
@@ -81,7 +83,33 @@ func successCode(code int) bool {
...
+	value := strings.TrimSpace(os.Getenv("PROXY_ENABLED"))
+
+	// If the environment variable is not set, ignore all this and let Go figure
+	// the rest of it.
+	if value == "" {
+		return true
+	}
+
+	// If the parsing fails, let's assume that the value is bad and ignore it.
+	v, err := strconv.ParseBool(value)
```
djoreilly, line 98: https://pkg.go.dev/strconv#ParseBool is different to https://github.com/SUSE/connect/commit/b2758adacf2092cbbef8b8286c201a7d1ba5e7e1#diff-925e25ec51fea79a5b120af303ad78fd1f34220e30134129edaa4bd73c77ff8dR86 

What exactly does yast use for true/false?

mssola:
@skazi0 @djoreilly note that I have basically reverted to the original code from this PR. That is, we don't use `ParseBool` and instead just do some basic comparisons. I've done this because `ParseBool` is not being too convenient on this case. Notice that the `/etc/sysconfig/proxy` files use a `yesno` notation, so this falls out of the `ParseBool` scope. I have also tried to use it eitherway as a final check, but the code was becoming more cumbersome than really needed. Thus, I have kept it simple and resorted to a much simpler implementation.

skazi0 (approved):
lgtm

internal/connect/connection.go
mssola: Fixed

internal/connect/connection.go
```diff
@@ -81,7 +82,23 @@ func successCode(code int) bool {
...
+}
+
 func proxyWithAuth(req *http.Request) (*url.URL, error) {
+	// Check for the special "PROXY_ENABLED" environment variable which might be
+	// set in a SUSE system. If it is set to a falsey value, then we skip proxy
+	// detection regardless of other environment variables.
+	if !proxyEnabled() {
+		return nil, nil
+	}
+
```
djoreilly, line 101: IMO something like this would be easier to understand 
```
       if strings.ToLower(strings.TrimSpace(os.Getenv("PROXY_ENABLED"))) == "no" {
               return nil, nil
       }
```

PR #151, opened 2022-09-20
Fix keepalive feature notice during installation

### LINKED ISSUE #149: calling  SUSEConnect in %pre
There is some slightly awkward code added in the %pre script:

https://github.com/SUSE/connect-ng/commit/f18a7120910ec5f74a09c6d52c756d9337bce385#diff-5d0b38a3678dc564a97924fc379f1d16d80d594c3c895100def8f0c9f0266f84R149-R153

This should be either moved to %post, or it needs to check whether SUSEConnect is actually installed (so this is an upgrade rather than a new install). for upgrade you can check whether `[ $1 -gt 1]`.

skazi0:
Previous condition worked fine but printed "command not found" for new installations. Added pre-test prevents this.

Closes #149

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -145,7 +145,7 @@ rm -rf %buildroot/usr/share/go
 # in pre blocks the old version is still installed. This way we can detect
 # if --keepalive was already present before
 kainfo=0
-helptext=$(SUSEConnect --help)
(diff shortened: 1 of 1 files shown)


djoreilly (approved):
LGTM

mssola (approved):

PR #152, opened 2022-09-20
suseconnect: added the `--list-ext` flag

mssola:
This is an alias for `-l` and `--list-extensions` which is supported in the Ruby SUSEConnect but not on this project.

Fixes bsc#1203574


mssola:
After some discussion we have decided to close this PR (see the conclusions on the bug report). Sorry for the noise.

PR #154, opened 2022-10-06
Drop .git from tar

skazi0:
The .git directory in tarball caused one of the package updates to be declined by reviewers. Removing as requested.


djoreilly (changes requested):
What is the reason for this change? Please add it to the commit message.

djoreilly (approved):
LGTM

PR #155, opened 2022-10-14
Use openssl go for SLE builds

skazi0:
The go1.18-openssl package is currently only available for SLE. Other targets should still use plain go versions.

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -31,7 +31,13 @@ Group:          System/Management
 Source:         connect-ng-%{version}.tar.xz
 Source1:        %name-rpmlintrc
 BuildRequires:  golang-packaging
+# use FIPS compliant go version for SLE targets and Leap 15.5+ targets
+%if ( 0%{?is_opensuse} == 0 && 0%{?sle_version} ) || ( 0%{?is_opensuse} == 1 && 0%{?sle_version} >= 150500 )
+# temporary until BuildRequires: go-openssl >= 1.16 works
+BuildRequires:  go1.18-openssl
+%else
 BuildRequires:  go >= 1.16
+%endif
 BuildRequires:  zypper
 BuildRequires:  ruby-devel
 %if %{with hwinfo}


djoreilly (dismissed):
LGTM

JanZerebecki:
I still need to investigate Leap, we will likely need to tweak the version selector to exclude Leap before 15.3. On SLE it looks good:
```
> osc buildinfo 'home:jzerebecki:CI:target:SUSE:connect-ng:PR-155/suseconnect-ng' SLE_15_SP3 x86_64 |grep go1.18-openssl
  <bdep name="go1.18-openssl" version="1.18.6.1" release="150000.1.3.1" arch="x86_64" hdrmd5="ea09fb2ba1edf3f818a40b56fb3aea69" project="SUSE:SLE-15:Update" repository="pool"/>
```

suseconnect-ng.spec
```diff
@@ -31,7 +31,13 @@ Group:          System/Management
 Source:         connect-ng-%{version}.tar.xz
 Source1:        %name-rpmlintrc
 BuildRequires:  golang-packaging
+# use FIPS compliant go version for SLE targets
+%if 0%{?sle_version}
```
JanZerebecki, line 35: We need to exclude Leap 15.3 and 15.4 here. 15.5 seems to build.

JanZerebecki (changes requested):

suseconnect-ng.spec
djoreilly: so it seems `sle_version` is also set on Leap. Should this conditional also check that `is_opensuse` is not set?
https://en.opensuse.org/openSUSE:Packaging_for_Leap#RPM_Distro_Version_Macros

JanZerebecki: the condition should be (I don't know the correct syntax from memory): sle_version is set or ( is_opensuse = 1 and sle_version >= 150500 )

djoreilly:
WIP - need to test that the syntax is correct.

JanZerebecki (dismissed):
The repo SUSE:SLE-15-SP1:Update is miss configured in OBS, ignore that failure. Looks good.

djoreilly:
repush - I put Leap 15.4+ in commit message by mistake

JanZerebecki (approved):

JanZerebecki:
Note: This will currently fail on SLE12 as there go-openssl was not yet released, once it is released it should work. However we can't test with SLE12 either way on OBS as the newer go versions are not synced from IBS.

digitaltom (approved):

digitaltom:
Will this be accepted in SLES like this? Let's merge and push to 'systemsmanagement:SCC/suseconnect-ng'.

PR #156, opened 2022-10-25
Bump to 1.0.0

mssola:
As part of the new release schema of SUSEConnect Ruby to only contain minor releases (thus, no major nor patch-level releases), it's safer to simply increase the version of SUSEConnect-ng to 1.0.0 and obsolete anything behind this major release.

Moreover, I have also removed the dots at the end of the summary of two packages because rpmlint was throwing an error on it.

I have tested this change by also adding a new `v1.0.0` tag, building it locally and then trying it out on a SLES15 SP3 with SUSEConnect 0.3.36 (the latest version from SUSEConnect ruby). The package was properly upgraded with no problems.

**Important**: whenever this PR is merged, I will push the v1.0.0 tag for this repo and start the release process of SUSEConnect-ng myself.

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -46,8 +46,8 @@ BuildRequires:  s390-tools
 BuildRequires:  systemd
 %endif
 
-Obsoletes:      SUSEConnect < 0.3.99
-Provides:       SUSEConnect = 0.3.99
+Obsoletes:      SUSEConnect < 1.0.0
+Provides:       SUSEConnect = 1.0.0
 Obsoletes:      zypper-migration-plugin < 0.99
 Provides:       zypper-migration-plugin = 0.99
 Obsoletes:      zypper-search-packages-plugin < 0.99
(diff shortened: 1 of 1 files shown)


mssola:
[[URAS] Release SUSEConnect-ng to 15SP4 as 1.0](https://trello.com/c/F18vobvC/2584-uras-release-suseconnect-ng-to-15sp4-as-10)

suseconnect-ng.spec
```diff
@@ -46,8 +46,8 @@ BuildRequires:  s390-tools
 BuildRequires:  systemd
 %endif
 
-Obsoletes:      SUSEConnect < 0.3.99
-Provides:       SUSEConnect = 0.3.99
+Obsoletes:      SUSEConnect < 1.0.0
+Provides:       SUSEConnect = 1.0.0
```
skazi0, line 50: if you plan to tag suseconnect-ng as version 1.0.0 why not use `%{version}` here instead of explicit values?

mssola: Because the version of suseconnect-ng might keep growing, but suseconnect ruby will never surpass `1.0.0`. I prefer being explicit on this.

skazi0: sounds good

skazi0 (approved):
lgtm

PR #157, opened 2022-11-21
Update Dockerfile.yast

djoreilly:
The test-yast make target was failing with:
STEP 2/8: RUN rm -r /usr/lib64/ruby/gems/*/gems/suse-connect-* rm: cannot remove '/usr/lib64/ruby/gems/*/gems/suse-connect-*': No such file or directory

The suseconnect-ng, libsuseconnect and suseconnect-ruby-bindings packages are now in the latest yast-ruby container image. This patch removes them, and the locally checked out code will be tested instead.


digitaltom (approved):
Works fine!

PR #159, opened 2022-12-01
Add: ExcludeArch: %ix86 s390 ppc64 to the .spec file, so we skip builds for unsupported architectures.

digitaltom:


suseconnect-ng.spec
```diff
@@ -36,6 +36,7 @@ BuildRequires:  zypper
 BuildRequires:  ruby-devel
 %if %{with hwinfo}
 %global test_hwinfo_args -test-hwinfo
+
 # packages required only for hwinfo tests
 %ifarch %ix86 ia64 x86_64 %arm aarch64
```
DimStar77, line 41: `%ifarch %ix86` when further down you Exclude %ix86 seems like some cleanup could be done

digitaltom: Updated to skip %ix86.

ngetahun (approved):
LGTM!

suseconnect-ng.spec
```diff
@@ -34,10 +34,13 @@ BuildRequires:  golang-packaging
 BuildRequires:  go >= 1.16
 BuildRequires:  zypper
 BuildRequires:  ruby-devel
+
 %if %{with hwinfo}
 %global test_hwinfo_args -test-hwinfo
+
+ExcludeArch: %ix86 s390 ppc64
```
DimStar77, line 41: Just noticed this in the Factory submission review: this is only excluded when with hwinfo is enabled.

Was that actually intended?

digitaltom: Oh, good catch, then we should move this out of the %if block.

digitaltom: Moved the 'ExcludeArch' out of the 'if' block: https://github.com/SUSE/connect-ng/pull/163

PR #162, opened 2022-12-06
Don't write system_token to service credentials files

### LINKED ISSUE #158: Don't write system_token to service credentials files
Zypper is only reading username + password from the service credentials files under `/etc/zypp/credentials.d`. 
The system_token line triggers a warning in zypper and a bug in `registercloudguest` which makes registration fail in public cloud. 

We need to make sure that suseconnect-ng doesn't write the system_token into the service credentials files, only in SCCcredentials. 

Bug: https://bugzilla.suse.com/show_bug.cgi?id=1205089
Fix for SUSEConnect: https://github.com/SUSE/connect/pull/503

djoreilly:
Zypper only reads the username and password from service credential files under /etc/zypp/credentials.d. The system_token line causes a problem with registercloudguest (bsc#1205089). Fixes #158

### DIFF
--- a/internal/connect/credentials.go
+++ b/internal/connect/credentials.go
@@ -111,7 +111,10 @@ func (c Credentials) write() error {
 		}
 	}
 	buf := bytes.Buffer{}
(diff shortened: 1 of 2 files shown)


digitaltom (approved):
Looks good to me, works as expected.

PR #164, opened 2023-02-21
Extract internal structs to improve readability & reusability

josegomezr:
Extracted some structs from internal functions outside to be readable. This will allow to spot similar looking data structures and merge them.


djoreilly:
I'm kinda neutral on this. I like the way anonymous structs allow you to avoid adding new types that are only going to be used once.

https://blog.boot.dev/golang/anonymous-structs-golang/

josegomezr:
That's a nice reading, thanks!

My endgoal is to use `SystemInformation` and `Credentials` (login, pw, token) more throughout the code.

To pave the way for a `--system-information` flag that displays what information the tool is collecting about the running system without contacting SCC at all.

But not making a gigantic PR on the process.

P.S.: I'll have a look at the broken pipeline image, ruby2 seems broken in the build process.

felixsch:
After speaking to Jose I close this here since we probably have not time to action on this in the foreseeable future and will integrate changes in our refactoring where possible!

PR #166, opened 2023-03-03
Fix Dockerfile bundler install

djoreilly:
ERROR:  Error installing bundler:
The last version of bundler (>= 0) to support your Ruby & RubyGems was 2.3.26. Try installing it with `gem install bundler -v 2.3.26` bundler requires Ruby version >= 2.6.0. The current ruby version is 2.5.0.


felixsch:
this is obsoleted by the CI changes

PR #167, opened 2023-03-03
Make keepalive on SUMA systems exit without error

djoreilly:
The keepalive option is called from a systemd timer. It is expected to fail on SUMA systems, so there is no need to to spam the logs. bsc#1207876

### DIFF
--- a/suseconnect/suseconnect.go
+++ b/suseconnect/suseconnect.go
@@ -147,6 +147,9 @@ func connectMain() {
 		exitOnError(err)
 		fmt.Println(output)
 	} else if keepAlive {
+		if isSumaManaged() {
+			os.Exit(0)
+		}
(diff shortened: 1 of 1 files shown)


digitaltom:
Thanks for the fix! Could you please add a reference to the changelog with the bug?

djoreilly:
> Thanks for the fix! Could you please add a reference to the changelog with the bug?

I moved `bsc#1207876` to the first line of the commit message. It should end up in the changes file.

https://build.opensuse.org/package/view_file/systemsmanagement:SCC/suseconnect-ng/suseconnect-ng.changes?expand=1

JanZerebecki (approved):
Note this still allows a user to use all the other commands, which is probably OK. Looks good, thx.

PR #168, opened 2023-03-09
Added MemTotal detection for HwInfo

mssola:


mssola:
Tests are still to be done

internal/connect/hwinfo.go
```diff
@@ -269,3 +273,44 @@ func readValues2map(b []byte) map[string]string {
...
+// be extracted for whatever reason.
+func systemMemory() int {
+	file, err := os.Open("/proc/meminfo")
+	if err != nil {
+		Debug.Print("'/proc/meminfo' could not be read!")
+		return 0
+	}
+	defer file.Close()
+
+	bufio.NewScanner(file)
```
djoreilly, line 306: I think L306 was left in by accident

internal/connect/hwinfo.go
```diff
@@ -269,3 +273,44 @@ func readValues2map(b []byte) map[string]string {
...
+// Returns an integer with the amount of megabytes of total memory (i.e.
+// `MemTotal` in /proc/meminfo). It will return 0 if this information could not
+// be extracted for whatever reason.
+func systemMemory() int {
+	file, err := os.Open("/proc/meminfo")
+	if err != nil {
+		Debug.Print("'/proc/meminfo' could not be read!")
+		return 0
+	}
+	defer file.Close()
```
djoreilly, line 304: You might consider passing `file` to a function that takes an `io.Reader` and that function will do everything else. Then there can be unit tests too. See https://github.com/SUSE/connect-ng/blob/b225bc392fef154bc2e5a69007f6695901e596fe/internal/connect/config.go#L84 and `config_test.go`

djoreilly (changes requested):

digitaltom:
Functionality works fine for me. 
I also tested the case where either `/proc/meminfo` or `MemTotal:` is not present: 
In that case suseconnect-ng sends `"memtotal":0`. Should we send that attribute at all in this case?

mssola:
@digitaltom 
> Should we send that attribute at all in this case?

I believe this is no longer sent. Either way, in the backend we will have to check for bogus values such as <= 0 and similar 👍🏻

internal/connect/hwinfo.go
```diff
@@ -41,6 +45,13 @@ func getHwinfo() (hwinfo, error) {
 	}
 	hw.Hostname = hostname()
 	hw.CloudProvider = cloudProvider()
+	hw.MemTotal = systemMemory()
+
+	// Include memory information if possible.
+	mem := systemMemory()
+	if mem > 0 {
+		hw.MemTotal = systemMemory()
+	}
```
djoreilly, line 55: I don't understand L51-54.  L48 has already set `hw.MemTotal`

mssola: Sorry about that, fixed!

felixsch (commented):
lgtm :+1: 


**How I reviewed the pull request:**

- Checked the changes and if I would suggest changes (which I did)
- Compiled and run the suseconnect and checked if the memory is transmitted to SCC as thought (yep it comes as `memtotal`? Do we want `mem_total` to be aligned with naming in the API?)
- Checked what happens if `/proc` is not readable: Connect fails gracefully because of `lscpu`.

**As always if you think I missed something I should test, please let me know!** :rocket:

felixsch (approved):
lgtm :+1:

internal/connect/hwinfo.go
```diff
@@ -31,6 +32,7 @@ type hwinfo struct {
 	Arch          string `json:"arch"`
 	UUID          string `json:"uuid"`
 	CloudProvider string `json:"cloud_provider"`
+	MemTotal      int    `json:"memtotal,omitempty"`
```
djoreilly, line 35: `mem_total` reads easier and would be consistent with the previous line.

internal/connect/hwinfo.go
```diff
@@ -269,3 +276,40 @@ func readValues2map(b []byte) map[string]string {
...
+		if len(fields) < 2 || fields[0] != "MemTotal:" {
+			continue
+		}
+
+		val, err := strconv.Atoi(fields[1])
+		if err != nil {
+			Debug.Printf("could not obtain memory information for this system: %v", err)
+			return 0
+		}
+		return val / 1024
```
djoreilly, line 296: Not sure why we are doing this. Maybe the scc api already expects MiB.

mssola: Yes, this was agreed with @digitaltom to just for for MiB, since we will usually care about this in the factor of TiB, so sending mere KiB might be an overkill.

internal/connect/hwinfo.go
mssola: Sure thing!

mssola: Done

felixsch (approved):
lgtm :+1:

Still looks good! :star:

djoreilly (approved):
lgtm

PR #174, opened 2023-05-30
Add --system-info(-text)? flags to connect

josegomezr:
These two flags are analogs of --status(-text)? flags, they'll output what connect detects for hardware info and what will end up in SCC when registering & doing keepalive.

It allow customers to inspect what will be going through the wire about the system without enabling all debug output.

As a nugget the text output of `--status-text` is now more terse by using more features from gotpl.


djoreilly (changes requested):
The commit message should also mention the addition of container detection.

felixsch:
After speaking to Jose I close this here since we probably have not time to action on this in the foreseeable future and will integrate changes in our refactoring where possible!

PR #175, opened 2023-06-14
Handle SLE12 system timers correctly

djoreilly:
The version of systemd on SLE12 does not support RandomizedDelaySec. So set a random time at installation instead.

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -193,6 +193,13 @@ if [ "$1" -eq 1 ]; then
 fi
 
 %post
+# Randomize schedule time for SLES12. SLES12 systemd does not support RandomizedDelaySec.
+%if (0%{?sle_version} > 0 && 0%{?sle_version} < 150000)
+    TIMER_HOUR=$(( RANDOM % 24 ))
+    TIMER_MINUTE=$(( RANDOM % 60 ))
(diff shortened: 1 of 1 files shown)


djoreilly:
This was taken from https://build.opensuse.org/package/rdiff/systemsmanagement:SCC/SUSEConnect?linkrev=base&rev=101

felixsch (approved):
lgtm :+1:

Thanks for fixing this.

- Checked that the code changes match the one from `SUSEConnect.spec`
- Built the package against SLE_15_SP3 and see it working
- Tried to build against SLE_12_SP5 but build is still broken due to missing dependecies

**As always, if you think I missed something I should test, please let me know!** :rocket:

djoreilly:
> * Tried to build against SLE_12_SP5 but build is still broken due to missing dependecies

I guess that is the go1.18-openssl dependency? That is not on OBS, but it is on IBS and it builds on SLE_12_SP5_Update here 
https://build.suse.de/package/show/home:doreilly:branches:Devel:SCC:suseconnect/suseconnect-ng so if you would like test  SLE_12_SP5 you can use https://download.suse.de/ibs/Devel:/SCC:/suseconnect/SLE_12_SP5_Update/Devel:SCC:suseconnect.repo

felixsch (approved):
lgtm :+1: 

- Run suseconnect from @djoreilly provided repository (thanks) and checked that the timer is setup correctly

Can you please merge the PR, I will merge it otherwise.

PR #182, opened 2023-08-28
Added EULA display for addons

mssola:
Display the license text before registering an addon product whenever that is available. SUSEConnect will then prompt with a message to accept or not said EULA.

This commit also adds a new configuration option (and flag) with which you can instruct SUSEConnect to auto-accept the license without any prompting.

Fixes bsc#1170267
Supersedes SUSE/connect-ng#114

## How to test

Before this patch:

1. Register a system.
2. Try to run a command such as: `# ./out/suseconnect -p sle-module-NVIDIA-compute/15/x86_64`
3. It should *fail*.

After this is merged, you should be able to re-run the same command and you will be given a license text to accept. If you accept, then you will proceed to register the addon as expected, otherwise it will exit with an error message.

Moreover, if you add the `--auto--auto-agree-with-licenses` flag, then there will be no user interaction needed and the license will be accepted automatically. The same happens if you toggle the configuration on `/etc/SUSEConnect` (see example file)

### DIFF
--- a/SUSEConnect.example
+++ b/SUSEConnect.example
@@ -1,0 +1,17 @@
+---
+## SUSEConnect configuration file example
+
+## URL of the registration server. (default: https://scc.suse.com)
+# url: https://scc.suse.com
+
+## Language code to use for error messages (default: $LANG)
+# language:
+
+## Do not verify SSL certificates when using https (default: false)
+# insecure: false
+
+## Do not refresh zypper service when registering (default: false)
+# no_zypper_refs: false
+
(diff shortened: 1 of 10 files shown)


mssola:
As for the failing CI, check out #183.

ngetahun (approved):
LGTM!

mssola:
[License Dialog for suseconnect-ng for 3rd Party Licenses](https://trello.com/c/tkD31bxN/2991-license-dialog-for-suseconnect-ng-for-3rd-party-licenses)

PR #186, opened 2023-09-15
packaging: updated the documentation

mssola:
Clarified the usage of the `revision` parameter from the `_service` file. This should clear up common pitfalls.


felixsch (approved):
lgtm :+1:

Issue #188, opened 2023-09-26
Tests crash when building a Fedora package


Conan-Kudo:
I'm trying to package up suseconnect-ng for Fedora, but the tests are failing with some strange crashes:

```
Executing(%check): /bin/sh -e /var/tmp/rpm-tmp.5E5B6M
+ umask 022
+ cd /builddir/build/BUILD
+ cd connect-ng-1.3.0
+ LDFLAGS=' -X github.com/SUSE/connect-ng/version=1.3.0'
+ GO_TEST_FLAGS='-buildmode=pie -compiler gc'
+ GO_TEST_EXT_LD_FLAGS='-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -Wl,--build-id=sha1 -specs=/usr/lib/rpm/redhat/redhat-package-notes  '
+ go-rpm-integration check -i github.com/SUSE/connect-ng -b /builddir/build/BUILD/connect-ng-1.3.0/_build/bin -s /builddir/build/BUILD/connect-ng-1.3.0/_build -V 1.3.0-0.fc40.1 -p /builddir/build/BUILDROOT/suseconnect-ng-1.3.0-0.fc40.1.x86_64 -g /usr/share/gocode -r '.*example.*'
Testing    in: /builddir/build/BUILD/connect-ng-1.3.0/_build/src
         PATH: /builddir/build/BUILD/connect-ng-1.3.0/_build/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin
       GOPATH: /builddir/build/BUILD/connect-ng-1.3.0/_build:/usr/share/gocode
  GO111MODULE: off
      command: go test -buildmode=pie -compiler gc -ldflags " -X github.com/SUSE/connect-ng/version=1.3.0 -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -Wl,--build-id=sha1 -specs=/usr/lib/rpm/redhat/redhat-package-notes  '"
      testing: github.com/SUSE/connect-ng
github.com/SUSE/connect-ng/internal/connect
--- FAIL: TestAnnounceSystem (0.02s)
[...]

PR #189, opened 2023-09-27
Fix invalid User-Agent field

djoreilly:
When the string in version.txt did not have a "~" GetShortenedVersion() was returning the string but adding a newline.

Fixes https://github.com/SUSE/connect-ng/issues/188


Conan-Kudo (approved):
This fixes it for me, thanks!

mssola (approved):

PR #190, opened 2023-10-10
Disallow the usage on transactional servers

mssola:
## Description

If SUSEConnect is being used from a transactional server, it should be disallowed since the recommended way on this situation is to use `transactional-update`. Otherwise, we will also disallow using SUSEConnect if the given root is mounted read-only.

## How to test

There are three sides for testing this PR.

### Still works

Check that using SUSEConnect on a regular system continues to work. Bonus points if this system is using BTRFS as the file system.

### Transactional server

1. Install the built binary into a recent MicroOS image (or a system with BTRFS and `transactional-update` installed).
2. Run the program and check that an error is given instructing the user to use `transactional-update` instead.
3. `transactional-update` works.

### Read only mount point

1. Mount a device in read only mode. One thing I did was to create a dummy loopback device and then mount that (check [this tutorial](https://linuxhandbook.com/create-virtual-block-device/) on how to do it).
2. Call SUSEConnect with `--root` pointing to this read only mount point.
3. You should get an error about trying to use a read only root.

### DIFF
--- a/internal/connect/system.go
+++ b/internal/connect/system.go
@@ -2,10 +2,21 @@ package connect
 
 import (
 	"bytes"
+	"errors"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"strings"
+	"syscall"
+)
+
+const (
+	// From <linux/mount.h>, <bits/statvfs.h>, etc.
+	ST_RDONLY = 0x1
+
+	// From <linux/magic.h>
+	BTRFS_SUPER_MAGIC = 0x9123683E
 )
 
 var systemEcho bool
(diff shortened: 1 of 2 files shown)


mssola:
[[6][ASR] Disallow users to run SUSEConnect-ng on a transactional update detected system](https://trello.com/c/kuijH4th/3037-6asr-disallow-users-to-run-suseconnect-ng-on-a-transactional-update-detected-system)

internal/connect/system.go
```diff
@@ -124,3 +135,37 @@ func UpdateCertificates() error {
...
+	}
+
+	// Just like zypper, we will assume that a BTRFS file system with the
+	// `transactional-update` binary installed is a transactional server.
+	_, err := os.Stat("/usr/sbin/transactional-update")
+	if statfs.Type == BTRFS_SUPER_MAGIC && err == nil {
+		// The user did not use the 'root' flag from the CLI: this is not
+		// `transactional-update` calling SUSEConnect but rather a user
+		// directly. This is not allowed.
+		if root == "" {
```
felixsch, line 160:
```suggestion
		if path == "/" {
```
I would change it to this, to make the behavior more clear (at least for me).

mssola: I chose `root` here because of the implicit meaning of it: `path` is what we compute, `root` is what the user provided. Hence, the if reads as "if the user did not provide anything".

As for the `--root /`, I'd say we follow the general logic: the user provided the `--root` flag (they know what they are doing, *hopefully*) and hence, if it's read only, it will be catched on the if down below. The general flow of the card was: if the user does not provide the root flag, then it better not be on a transactional system; otherwise, if they provide the flag, it's either `transactional-update` or someone doing something *special*, which we should only forbid if it's read only.

felixsch: I see your point. Fair enough!

Conan-Kudo:
This would fail if you're using the DNF stack with integrated transactional update support. What's the actual problem here that requires disallowing the tool from working?

felixsch:
> This would fail if you're using the DNF stack with integrated transactional update support. What's the actual problem here that requires disallowing the tool from working?

In wish SUSE specific scenario is it common to use the `dnf` stack? 

The problem at hand is, that `SUSEConnect` allows performing a registration and product activation which then is deleted due to the nature of a transactional system, leaving the system behind in an undefined state. This needs to be solved.

If the `dnf` stack is using `SUSEConnect` directly, it is possible to use the `--root` command (e.g. as `transactional-update` does) to circumvent the checks.

Conan-Kudo:
Wouldn't it make more sense to have the SUSEConnect tool be tukit aware and do the right thing anyway? tukit has both an executable wrapper and a library interface that could be used for this.

mssola:
I believe that SUSEConnect should strive to be as simple as possible. In this case, we don't want to deal with transactional updates because, for SUSE, this is handled through other ways. This PR is a way of saying: if you are using SUSEConnect in a SUSE OS using a transactional update strategy, then use the proper tool for it. That is, this "tukit aware" tool that you are mentioning already exists at SUSE, with this PR we are pointing to it from SUSEConnect.

Note that you can still use the `--root` flag as pointed out by @felixsch if you know what you are doing. With this PR this would still work in other scenarios like one managed through dnf.

Conan-Kudo:
> You seem to know about the `dnf` stack. Can you please point out where this would break the `dnf` integration? Because we are not aware that `SUSEConnect` is directly related to the project.

Right now, we don't have anything. But I'm hoping that we can use this to implement [a similar plugin to the one in DNF that uses librhsm](https://github.com/rpm-software-management/dnf5/commit/dccea32e5a12e8fe6a4afa96139959d4c76556e6).

asmorodskyi:
Can we improve this PR by not blocking `SUSEConnect --status-text` ( and any other calls which not changing anything so will work perfectly even on read-only FS ) ?

PR #191, opened 2023-10-24
Set docker and podman auth credentials if possible

felixsch:
## Write system credentials into `docker` and `podman` authentication file
Populate `$HOME/.docker/config.json` and `$XDG_RUNTIME_DIR/containers/auth.json` with system credentials provided after registration.

internal card: https://trello.com/c/k6GueN3v/3051-7asr-suseconnect-ng-preconfigures-docker-configuration

*How to test this feature:*

helpful snippets:
```
# bootstrap container with connect + branch added
host$ docker run --rm -ti --privileged registry.suse.com/suse/sle15:15.5
> zypper ar https://download.opensuse.org/repositories/home:/fschnizlein:/branches:/systemsmanagement:/SCC/SLE_15_SP5/home:fschnizlein:branches:systemsmanagement:SCC.repo
> zypper ref
> zypper in -y suseconnect-ng jq
host$ docker commit <container_id> ng-testing
# docker run --rm -ti --privileged ng-testing

# Paths
docker: $HOME/.docker/config.json
podman: $XDG_RUNTIME_DIR/containers/auth.json

# Check credentials:
cat <path> | jq -Mr '.auths["https://registry.suse.com"].auth' | base64 -d
```

Test matrix:
- Registering a system normally and `XDG_RUNTIME_DIR` is `unset`. Check if login is available in docker config
- Registering a system but before set `XDG_RUNTME_DIR` to `$HOME` And check that `containers/auth.json` is populated
- Check if configs (docker/podman) already exists that the config is added and all existing data is preserved
- Check if a login for `https://registry.suse.com` already exists, nothing is changed
- Check when de-registering that configs are reset
- Check when de-registering that configs are *NOT* reset if the login does not match the system credentials

The hard part:
- Run an environment where you actually populate `XDG_RUNTIME_DIR` with real world values (namely `/var/run/1000`) and make sure the above works

If you need help, feel free to reach out to me!

Thanks for looking into this!

### DIFF
--- a/internal/connect/client.go
+++ b/internal/connect/client.go
@@ -124,6 +124,12 @@ func Deregister() error {
 		}
 	}
 
+	// remove potential docker and podman configurations for our registry
+	creds, err := getCredentials()
+	if err == nil {
+		removeRegistryAuthentication(creds.Username, creds.Password)
+	}
+
 	if err := deregisterSystem(); err != nil {
 		return err
 	}
@@ -350,13 +356,6 @@ func DeactivateProduct(product Product) (Service, error) {
 
 // DeregisterSystem deletes current system in SMT/SCC
 func DeregisterSystem() error {
-	// In case fetching credentials fails do not break
-	// but continue to allow deregistration with remote API
-	creds, err := getCredentials()
-	if err == nil {
-		removeRegistryAuthentication(creds.Username, creds.Password)
-	}
-
 	return deregisterSystem()
 }
 
--- a/internal/connect/client.go
+++ b/internal/connect/client.go
@@ -227,7 +227,11 @@ func announceOrUpdate() error {
 	if err != nil {
 		return err
 	}
-	return writeSystemCredentials(login, password, "")
+
+	if err = writeSystemCredentials(login, password, ""); err == nil {
+		setupRegistryAuthentication(login, password)
+	}
+	return err
 }
 
 // IsRegistered returns true if there is a valid credentials file
@@ -346,6 +350,13 @@ func DeactivateProduct(product Product) (Service, error) {
 
 // DeregisterSystem deletes current system in SMT/SCC
 func DeregisterSystem() error {
+	// In case fetching credentials fails do not break
+	// but continue to allow deregistration with remote API
+	creds, err := getCredentials()
+	if err == nil {
+		removeRegistryAuthentication(creds.Username, creds.Password)
+	}
+
 	return deregisterSystem()
 }
 
--- a/internal/connect/registry_auth.go
+++ b/internal/connect/registry_auth.go
@@ -7,17 +7,15 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
 )
 
-// If the configuration file is not found for podman which resides
-// in ${XDG_RUNTIME_DIR}/containers/auth.json it will fall back to
-// docker configuration in ${HOME}/.docker/config.json
-// See: https://docs.podman.io/en/stable/markdown/podman-login.1.html
+// NOTE: Podman will read the same configuration file as the docker cli.
+//       We utilize this to not have to write multiple files. Since the default
+//       config is set to $XDG_RUNTIME_DIR/container/auth.json which should considered
+//       as volatile storage.
 
 const (
 	DEFAULT_DOCKER_CLIENT_CONFIG = ".docker/config.json"
-	DEFAULT_PODMAN_CONFIG        = "containers/auth.json"
 	DEFAULT_SUSE_REGISTRY        = "https://registry.suse.com"
 )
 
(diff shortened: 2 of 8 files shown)


internal/connect/client.go
```diff
@@ -346,6 +350,13 @@ func DeactivateProduct(product Product) (Service, error) {
 
 // DeregisterSystem deletes current system in SMT/SCC
 func DeregisterSystem() error {
+	// In case fetching credentials fails do not break
+	// but continue to allow deregistration with remote API
+	creds, err := getCredentials()
+	if err == nil {
```
ngetahun, line 356: This call should go in `Deregister` function, `deregisterSystem()` is also called early.

internal/connect/client.go
```diff
@@ -346,6 +350,13 @@ func DeactivateProduct(product Product) (Service, error) {
 
 // DeregisterSystem deletes current system in SMT/SCC
 func DeregisterSystem() error {
```
ngetahun, line 352: Preliminary test looks good. Testing it on `registry.suse.com/suse/sle15:15.4` and found the credentials written in `/root/.docker/config.json`. Deregistering doesn't remove the credentials, see below comment.

ngetahun (changes requested):

internal/connect/registry_auth.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return json.Unmarshal(data, cfg)
+}
+
+func (cfg *RegistryAuthConfig) SaveTo(path string) error {
+	data, err := json.MarshalIndent(cfg, "", "  ")
+	if err != nil {
+		return err
+	}
+	return writeFile(path, data, 0600)
```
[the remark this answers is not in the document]
felixsch: Both `podman` and `docker` set `0600` as permissions. I do not think there should be an issue then:
docker: https://github.com/docker/cli/blob/master/cli/config/configfile/file_unix.go#L14
podman: https://github.com/containers/image/blob/main/pkg/docker/config/config.go#L618

But I have no strong opinion, should I set it to `0640`?

internal/connect/client.go
felixsch: thanks for finding the spot, deregistering was broken as you found out. I will move it there

internal/connect/registry_auth.go
```diff
@@ -0,0 +1,174 @@
...
+
+		if path, ok := pathFn(); ok {
+			// This also fails if the file does not yet exist
+			// so we continue to create it.
+			if err := config.LoadFrom(path); err != nil {
+				Debug.Printf("Could not load `%s`: %s", path, err)
+			}
+			config.Set(DEFAULT_SUSE_REGISTRY, login, password)
+
+			if err := mkDirAll(filepath.Dir(path), 0777); err != nil {
```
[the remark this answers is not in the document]
felixsch: No, I checked before using it here. I already had a comment about it here and removed it because std go lib specifies the behaviour:

See: https://pkg.go.dev/os#MkdirAll

internal/connect/client.go
felixsch: Jep, this and the ownership of the files is wrong. While `docker` config is fine, `podman` config should follow the ownership of `%XDG_RUNTIME_DIR` and not the calling user (aka. root). Will fix!

internal/connect/client.go
felixsch: moved!

internal/connect/client.go
felixsch: moved!

internal/connect/registry_auth.go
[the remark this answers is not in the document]
ngetahun: Actually, the file access mode will not help at all here. I tested with a rootless docker installation and seems like root will not have `XDG_RUNTIME_DIR` set. This is a problem, but something that we need to clearly communicate with the customer.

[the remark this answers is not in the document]
felixsch: Is this resolved after, we ditched all of the `XDG_RUNTIME_DIR` handling? Now its only `root` which will have a configured `auth.json` which is fine I believe. For ALP it is yet to be declared how most containers will run. Since they aim for system applications in containers, I assume most containers will need root access anyway..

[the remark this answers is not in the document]
ngetahun: My comment is that we need to note somewhere that we don't support rootless/podman setup.

ngetahun (approved):
I will approve this now. Only [this](https://github.com/SUSE/connect-ng/pull/191#discussion_r1374452267) discussion is left which is a non-blocker.

felixsch:
I will merge this now, even jenkins is red but the affected feature test is testing translation which is not available due to Hernans changes in the API, invalidating this one test.

PR #194, opened 2023-11-23
Implemented the --json flag when de/registering

mssola:
## Description

This commit adds the `json` flag just for the register and de-register commands. This flag will then disable all the usual output from both these commands and instead print to stdout the JSON schema as agreed on RFC 0038 regarding the SUSEConnect and Agama integration.

## How to test

Simply build this and call the register/de-register commands with `--json`. If you do so with a command that is not either of those, you will get:

```
# ./out/suseconnect --json -l
SUSEConnect error: cannot use the json option with the 'list-extensions' command
```

Then, if you register:

```
# ./out/suseconnect --json -r <regcode> | jq
```

This will take some time but in the end it will give you JSON output as expected from the RFC. As for de-registering, the RFC does not provide a model, but I have implemented the same schema as for registering. Hence, the following will provide a similar output as the command above:

```
# ./out/suseconnect --json --de-register | jq
```

If the operation fails, a json-formatted error is also given. For example, if you try to de-register a system which is not registered, this is given:

```
# ./out/suseconnect --json --de-register | jq
{
  "success": false,
  "products": null,
  "message": "System not registered"
}
```

### DIFF
--- a/internal/connect/client.go
+++ b/internal/connect/client.go
@@ -1,16 +1,43 @@
 package connect
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
 )
 
+type RegisterOut struct {
+	Success  bool             `json:"success"`
+	Products []ProductService `json:"products"`
+	Message  string           `json:"message"`
+}
+
+type ProductService struct {
+	Product ProductOut `json:"product"`
+	Service ServiceOut `json:"service"`
+}
+
+type ProductOut struct {
+	Name       string `json:"name"`
+	Identifier string `json:"identifier"`
+	Version    string `json:"version"`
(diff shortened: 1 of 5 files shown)


mssola:
[[6][ASR] Add --json cli option (only for registration, de-registration)](https://trello.com/c/UA5AHtWA/3109-6asr-add-json-cli-option-only-for-registration-de-registration)

mssola:
Update the deregister command to also add the products that have been deregistered. Hence, now register and deregister follow the same schema.

The only thing missing from the RFC is localization, which is a big topic that I believe it's out of the scope from this PR.

ngetahun (approved):
LGTM!

PR #195, opened 2023-12-11
Add JSON output for --list-extensions

felixsch:
card: https://trello.com/c/IkPSyJf1/3125-asr-add-json-output-for-list-extensions

**How to review this pull request:**

```
$ = host
> = container

$ docker run --rm -it registry.suse.com/suse/sle15:15.5
> zypper in go1.21-openssl jq make vim curl wget

# At this point I would grab the container id on the host and commit it
$ docker ps
$ commit <container_id> connect-devel
> exit

$ docker run --rm -it $(pwd):/connect connect-devel
> cd /connect
> make build
> .out/suseconnect -r <regcode>
> .out/suseconnect --list-extensions
> .out/suseconnect --json --list-extensions | jq
```

Thank you very much for your review :)

PR #196, opened 2023-12-19
Rename name to identifier and friendly_name to name

felixsch:
Addition to: https://github.com/SUSE/connect-ng/pull/195

**How to test this pull request:**

```
$ = host
> = container

$ docker run --rm -it registry.suse.com/suse/sle15:15.5
> zypper in go1.21-openssl jq make vim curl wget

# At this point I would grab the container id on the host and commit it
$ docker ps
$ commit <container_id> connect-devel
> exit

$ docker run --rm -it -v $(pwd):/connect connect-devel
> cd /connect
> make build
> .out/suseconnect -r <regcode>
> .out/suseconnect --list-extensions
> .out/suseconnect --json --list-extensions | jq
# expect: identifier and name both are present and now friendly_name is shown
```

Thank you very much for your review :)

### DIFF
--- a/internal/connect/extensions.go
+++ b/internal/connect/extensions.go
@@ -25,10 +25,10 @@ var (
 )
 
 type extension struct {
-	Name         string       `json:"name"`
+	Name         string       `json:"identifier"`
 	Version      string       `json:"version"`
 	Arch         string       `json:"arch"`
-	FriendlyName string       `json:"friendly_name"`
+	FriendlyName string       `json:"name"`
 	Activated    bool         `json:"activated"`
 	Available    bool         `json:"available"`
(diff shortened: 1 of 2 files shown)


mssola:
Would it make sense to extend this to the `Product` struct as well? There is a note from me on commit 6eec655d7f90fd3e1eb60965a399f9efe3253861 on the same issue but for `Product` 😄

digitaltom (approved):
Amazing instructions :green_heart:

felixsch:
@mssola I agree we should map the better in general but `Product` depends on `xml` output from zypper + SCC reponses. All of them use different attribute names.

I'm already happy if we manage to have a consistent output names in connect :)

PR #199, opened 2024-01-16
Don't error out in AcceptEULA() if base product wasn't found, just skip

digitaltom:
How to reproduce: 

```
$ cd <connect-ng>
$ docker run --rm -it -v $(pwd):/connect registry.suse.com/suse/sle15:15.3
> zypper rm -y container-suseconnect
> zypper in -y go1.21 make
> cd /connect
> make build
> zypper in -y http://updates.suse.de/SUSE/Updates/SLE-Module-Server-Applications/15-SP3/x86_64/update/noarch/migrate-sles-to-sles4sap-15.1.2-3.9.1.noarch.rpm
> cp out/suseconnect /usr/sbin/SUSEConnect
# Check the card for a regcode
> /usr/sbin/Migrate_SLES_to_SLES-for-SAP.sh
```

### DIFF
--- a/internal/connect/eula.go
+++ b/internal/connect/eula.go
@@ -196,7 +196,7 @@ func AcceptEULA() error {
 	// Fetch list of extensions and search for requested product
 	base, err := baseProduct()
 	if err != nil {
-		return err
+		return nil
 	}
 
 	// If we encounter a switch of base products (e.g. SLES -> SLES_SAP)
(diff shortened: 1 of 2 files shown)


felixsch (approved):
lgtm :+1: 

**How I tested this pull request:**

- Checked the source changes for parts I would suggest to change
- Run the migration from SLES to SLES_SAP using the script and the fixed connect to check if it works.

PR #200, opened 2024-01-18
Bring back EULA display

digitaltom:
Bring back eula display on product activation, but skip when the base product cannot get loaded, for example because the system is not yet registered, or it is switching the base product.

How to test: 

```
docker run --rm -it -v $(pwd):/connect connect-devel
cd connect
make build
# direct base product activation works
./out/suseconnect -p sles/15.5/x86_64 -r ...
./out/suseconnect -d
# base product switch works
./out/suseconnect -p SLES_SAP/15.5/x86_64 -r ...
./out/suseconnect -d
# show eula
./out/suseconnect -r ...
./out/suseconnect -p sle-module-desktop-applications/15.5/x86_64
./out/suseconnect -p sle-module-development-tools/15.5/x86_64
./out/suseconnect -p sle-module-NVIDIA-compute/15/x86_64

```

### DIFF
--- a/internal/connect/eula.go
+++ b/internal/connect/eula.go
@@ -199,22 +199,19 @@ func AcceptEULA() error {
 		return nil
 	}
 
-	// If we encounter a switch of base products (e.g. SLES -> SLES_SAP)
-	// we can not fetch the product information and thus the EULA since there
+	// Skip if we can not fetch the product information and thus the EULA since there
 	// might be no registration in place right now.
 	// See bsc#1218649 and bsc#1217961
-	if base.ToTriplet() != CFG.Product.ToTriplet() {
-		return nil
-	}
-
 	prod, err := showProduct(base)
 	if err != nil {
-		return err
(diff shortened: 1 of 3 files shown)


internal/connect/eula.go
```diff
@@ -199,22 +199,19 @@ func AcceptEULA() error {
...
 	// See bsc#1218649 and bsc#1217961
-	if base.ToTriplet() != CFG.Product.ToTriplet() {
-		return nil
-	}
-
 	prod, err := showProduct(base)
 	if err != nil {
-		return err
+		Debug.Print("Cannot load base product details for EULA.")
+		return nil
```
ngetahun, line 208: Why not return `err` here?

digitaltom: If we return `err` here, the process will fail when the base product cannot get loaded, for example because the system is not yet registered, or it is switching the base product

felixsch (dismissed):
lgtm :+1:

PR #206, opened 2024-02-20
Expect go1.21 for our builds

felixsch:
Changes for the rpm builds in the dawn of a new CI implementation:
  * Update to go1.21-openssl
  * Remove go-packaging (important!)
  * Cleanup spec file
  * Remove unit testing in the spec file, since this is done on repository level
  * Do not name releases after the git SHA. We do not use this and it makes things more complicated than they need to be

part of: https://trello.com/c/98dlYPsd/3205-pre-tedd-feature-tests-in-suseconnect-ng-without-the-old-repository-suseconnect and https://trello.com/c/GTlSegDS/5006-suseconnect-ng-update-go-openssl-dependency

**How to test these changes:**

- Check the spec file changes
- Think if this might break in any instance
- Run a functional test with help of:

```
$ docker run -it --rm registry.suse.com/suse/sle15:15.5
> zypper ar "https://download.suse.de/ibs/home:/fschnizlein:/branches:/Devel:/SCC:/suseconnect/SLE_12_SP5_Update/home:fschnizlein:branches:Devel:SCC:suseconnect.repo"
> zypper ref
> zypper in suseconnect-ng

# From here it is up to you to check everything working
# examples to test:
> SUSEConnect -r <regcode>
# expect: works!

> zypper in yast2-registration
> yast2 registration
# expect: works!
``

Thank you very much for reviewing this PR :heart: 

**As always, if you have trouble with testing this PR, do not hesitate to reach out to me**

### DIFF
--- a/suseconnect-ng.spec
+++ b/suseconnect-ng.spec
@@ -16,48 +16,30 @@
 #
 
 
-%global provider_prefix github.com/SUSE/connect-ng
-%global import_path     %{provider_prefix}
-
-# set this to enable hwinfo test in %check
-%bcond_with hwinfo
+%global project github.com/SUSE/connect-ng
 
 Name:           suseconnect-ng
-# the version will get set by the 'set_version' service
-Version:        1.1.0~git2.f42b4b2
+Version:        1.7.0
 Release:        0
 URL:            https://github.com/SUSE/connect-ng
 License:        LGPL-2.1-or-later
 Summary:        Utility to register a system with the SUSE Customer Center
 Group:          System/Management
-Source:         connect-ng-%{version}.tar.xz
-Source1:        %name-rpmlintrc
+Source:         suseconnect-ng-%{version}.tar.xz
+Source1:        %{name}-rpmlintrc
 
 # Build against latest golang in Tumbleweed and
 # go1.18-openssl on all other distributions
-%if 0%{?suse_version} >= 1600
-BuildRequires: golang(API)
+%if 0%{?suse_version} > 1600
+BuildRequires:  golang(API)
 %else
-BuildRequires: go1.18-openssl
+BuildRequires:  go1.21-openssl
 %endif
 
-BuildRequires:  golang-packaging
 BuildRequires:  ruby-devel
 BuildRequires:  zypper
 
 ExcludeArch:    %ix86 s390 ppc64
-%if %{with hwinfo}
-%global test_hwinfo_args -test-hwinfo
-
-# packages required only for hwinfo tests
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: suseconnect-ng.spec [hunk 1]
```diff
@@ -16,60 +16,44 @@
...
 License:        LGPL-2.1-or-later
 Summary:        Utility to register a system with the SUSE Customer Center
 Group:          System/Management
-Source:         connect-ng-%{version}.tar.xz
-Source1:        %name-rpmlintrc
+Source:         suseconnect-ng-%{version}.tar.xz
+Source1:        %{name}-rpmlintrc
 
 # Build against latest golang in Tumbleweed and
 # go1.18-openssl on all other distributions
```
digitaltom, line 32: Maybe there's a special macro for Tumbleweed/Factory that's not related to the version number? Then we would not need to change this on each release.

digitaltom:
```suggestion
# go1.21-openssl on all other distributions
```

suseconnect-ng.spec
```diff
@@ -16,60 +16,44 @@
...
-%global import_path     %{provider_prefix}
-
-# set this to enable hwinfo test in %check
-%bcond_with hwinfo
+%global project github.com/SUSE/connect-ng
 
 Name:           suseconnect-ng
-# the version will get set by the 'set_version' service
-Version:        1.1.0~git2.f42b4b2
+Version:        1.7.0
```
josegomezr, line 22: The version is still applied by `set_version` above, right?

suseconnect-ng.spec
felixsch: problem is the `-openssl` version. I ask around if there is also a option to automatically default to the latest one

suseconnect-ng.spec
felixsch: yep. Do you want back the comment? I find it unnecessary since the version supplied here is actually already overwritten by `set_version` at this stage.

josegomezr: I don't mind either of the two, was just curious :+1:

suseconnect-ng.spec
felixsch: I checked. The `golang(API)` is just a label for the `Provides:` which is added by the `go-x.xx` packages. Both the non and the openssl version does provide the same flag. This means it will not necessarily select the openssl version then. So no special handling for this. we need it hardcoded for now!

likhitha77:
SUSEConnect -r and yast2 registration works as expected!

likhitha77 (approved):

PR #207, opened 2024-02-27
Automatic builds in OBS with new spec file

felixsch:
build against new target using the original sources rather then a specific CI project in obs

### DIFF
--- a/.obs/workflows.yml
+++ b/.obs/workflows.yml
@@ -1,6 +1,6 @@
 workflow:
   steps:
     - branch_package:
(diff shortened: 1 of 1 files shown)


digitaltom (approved):
in the OBS project systemsmanagement:SCC:CI, there is a package 'connect', while this would probably create one named 'suseconnect-ng'.

mssola:
Is this PR still on? Looks good from my side if this project is able to continuously build this package as intended.

felixsch:
No this is obsolete from my perspective. In future we should add building the package in OBS again but we need a dedicated task for it. I close this here now!

PR #210, opened 2024-03-06
add system uptime capability

guangyee:
Add the ability to upload the system uptime logs, produced by the suse-uptime-tracker daemon, to SCC/RMT as part of daily heartbeat report. We need to ability to track system uptime to hourly granularity so MSPs can utility that data accordingly (i.e. billing).

This feature is optionally enabled by installing the suse-uptime-tracker package and enabling the suse-uptime-tracker.timer service.

### DIFF
--- a/SUSEConnect.example
+++ b/SUSEConnect.example
@@ -15,3 +15,6 @@
 
 ## Automatically agree to extension and module license confirmation prompts (default: false)
 # auto_agree_with_licenses: false
+
+## Enable system uptime tracking. The system uptime log will be sent to SCC/RMT as part of heartbeat.
+# enable_system_uptime_tracking: false
--- a/SUSEConnect.example
+++ b/SUSEConnect.example
@@ -16,5 +16,5 @@
 ## Automatically agree to extension and module license confirmation prompts (default: false)
 # auto_agree_with_licenses: false
 
-## Enable system uptime tracking. The system uptime log will be sent to SCC/RMT as part of heartbeat.
+## Enable system uptime tracking. The system uptime log will be sent to SCC/RMT as part of keepalive.
 # enable_system_uptime_tracking: false
--- a/build/packaging/suseconnect-ng.changes
+++ b/build/packaging/suseconnect-ng.changes
@@ -7,6 +7,9 @@ Thu Apr 25 15:39:00 UTC 2024 - Felix Schnizlein <fschnizlein@suse.com>
   * Add --gpg-auto-import-keys flag before action in zypper command (bsc#1219004)
   * Include /etc/products.d in directories whose content are backed
     up and restored if a zypper-migration rollback happens. (bsc#1219004)
+  * Add the ability to upload the system uptime logs, produced by the
+    suse-uptime-tracker daemon, to SCC/RMT as part of keepalive report.
+    (jsc#PED-7982) (jsc#PED-8018)
 
 -------------------------------------------------------------------
 Tue May  7 14:12:47 UTC 2024 - Thomas Schmidt <tschmidt@suse.com>
--- a/internal/connect/api.go
+++ b/internal/connect/api.go
@@ -189,8 +189,13 @@ func updateSystem(body []byte) error {
 // it will be interpreted as if the system uptime log feature is not
 // enabled. Hence an empty array will be returned.
 func readUptimeLogFile(uptimeLogFilePath string) ([]string, error) {
-	// NOTE: if uptime log file does not exist, we assume the uptime
-	// tracking feature is not enabled
+	// NOTE: the uptime log file is produced by the suse-uptime-tracker
+	// (https://github.com/SUSE/uptime-tracker) service. If the service
+	// is installed and enabled, barring any unforeseen errors, the
+	// uptime log file is expected to there and updated on the regular
+	// basis. If the service is not installed or otherwise disabled, the
+	// uptime log file may not exist. In that case we assume the uptime
+	// tracking feature is disabled.
 	_, err := os.Stat(uptimeLogFilePath)
 	if errors.Is(err, os.ErrNotExist) {
 		return nil, nil
@@ -200,15 +205,14 @@ func readUptimeLogFile(uptimeLogFilePath string) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer uptimeLogFile.Close()
 	fileScanner := bufio.NewScanner(uptimeLogFile)
-	fileScanner.Split(bufio.ScanLines)
 	var logEntries []string
 
 	for fileScanner.Scan() {
 		logEntries = append(logEntries, fileScanner.Text())
 	}
-	err = uptimeLogFile.Close()
-	if err != nil {
+	if err = fileScanner.Err(); err != nil {
 		return nil, err
 	}
 	return logEntries, nil
@@ -239,9 +243,9 @@ func makeSysInfoBody(distroTarget, namespace string, instanceData []byte, includ
 	if includeUptimeLog {
 		uptimeLog, err := readUptimeLogFile(UptimeLogFilePath)
 		if err != nil {
-			util.Info.Print("Unable to system uptime log")
-		}
-		if uptimeLog != nil {
+			util.Debug.Printf("Unable to read uptime log: %v", err)
+			util.Info.Print("Unable to read system uptime log")
+		} else {
 			payload.OnlineAt = uptimeLog
 		}
 	}
--- a/internal/connect/api.go
+++ b/internal/connect/api.go
@@ -1,10 +1,18 @@
 package connect
 
 import (
+	"bufio"
 	"encoding/json"
+	"errors"
 	"net/http"
+	"os"
 
+	"github.com/SUSE/connect-ng/internal/util"
 	"github.com/SUSE/connect-ng/internal/zypper"
+)
+
+const (
+	UptimeLogFilePath = "/etc/zypp/suse-uptime.log"
 )
 
 // announceSystem announces a system to SCC
@@ -176,14 +184,45 @@ func updateSystem(body []byte) error {
 	return err
 }
 
+// readUptimeLogFile reads the system uptime log from a given file and
+// returns them as a string array. If the given file does not exist,
+// it will be interpreted as if the system uptime log feature is not
+// enabled. Hence an empty array will be returned.
+func readUptimeLogFile(uptimeLogFilePath string) ([]string, error) {
+	// NOTE: if uptime log file does not exist, we assume the uptime
+	// tracking feature is not enabled
+	_, err := os.Stat(uptimeLogFilePath)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil, nil
+	}
+
+	uptimeLogFile, err := os.Open(uptimeLogFilePath)
+	if err != nil {
+		return nil, err
+	}
+	fileScanner := bufio.NewScanner(uptimeLogFile)
+	fileScanner.Split(bufio.ScanLines)
+	var logEntries []string
+
+	for fileScanner.Scan() {
+		logEntries = append(logEntries, fileScanner.Text())
+	}
+	err = uptimeLogFile.Close()
+	if err != nil {
+		return nil, err
+	}
+	return logEntries, nil
+}
+
 // makeSysInfoBody returns the JSON payload needed for the announce/update system calls
-func makeSysInfoBody(distroTarget, namespace string, instanceData []byte) ([]byte, error) {
+func makeSysInfoBody(distroTarget, namespace string, instanceData []byte, includeUptimeLog bool) ([]byte, error) {
 	var payload struct {
-		Hostname     string `json:"hostname"`
-		DistroTarget string `json:"distro_target"`
-		InstanceData string `json:"instance_data,omitempty"`
-		Namespace    string `json:"namespace,omitempty"`
-		Hwinfo       hwinfo `json:"hwinfo"`
+		Hostname     string   `json:"hostname"`
+		DistroTarget string   `json:"distro_target"`
+		InstanceData string   `json:"instance_data,omitempty"`
+		Namespace    string   `json:"namespace,omitempty"`
+		Hwinfo       hwinfo   `json:"hwinfo"`
+		OnlineAt     []string `json:"online_at,omitempty"`
 	}
 	if distroTarget != "" {
 		payload.DistroTarget = distroTarget
@@ -196,6 +235,16 @@ func makeSysInfoBody(distroTarget, namespace string, instanceData []byte) ([]byt
 	}
 	payload.InstanceData = string(instanceData)
 	payload.Namespace = namespace
+
+	if includeUptimeLog {
+		uptimeLog, err := readUptimeLogFile(UptimeLogFilePath)
+		if err != nil {
+			util.Info.Print("Unable to system uptime log")
+		}
+		if uptimeLog != nil {
+			payload.OnlineAt = uptimeLog
+		}
+	}
 
 	hw, err := getHwinfo()
 	if err != nil {
--- a/internal/connect/api_test.go
+++ b/internal/connect/api_test.go
@@ -307,19 +307,18 @@ func createTestUptimeLogFileWithContent(content string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+	defer tempFile.Close()
 	tempFilePath := tempFile.Name()
 	if _, err := tempFile.WriteString(content); err != nil {
 		os.Remove(tempFilePath)
 		return "", err
 	}
-	defer tempFile.Close()
 
 	return tempFilePath, nil
 }
 
 func TestUptimeLogFileDoesNotExist(t *testing.T) {
-	uptimeLogFileContent := ``
-	tempFilePath, err := createTestUptimeLogFileWithContent(uptimeLogFileContent)
+	tempFilePath, err := createTestUptimeLogFileWithContent("")
 	if err != nil {
 		t.Fatalf("Failed to create temp uptime log file for testing")
 	}
@@ -340,7 +339,7 @@ func TestReadUptimeLogFile(t *testing.T) {
 	defer os.Remove(tempFilePath)
 	uptimeLog, err := readUptimeLogFile(tempFilePath)
 	if err != nil {
-		t.Fatalf("Failed to read uptime log file")
+		t.Fatalf("Failed to read uptime log file: %s", err)
 	}
 	if uptimeLog == nil {
 		t.Fatal("Failed to open uptime log file")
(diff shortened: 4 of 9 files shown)


digitaltom:
@guangyee don't we need a config option in suseconnect to enable the uptime collection? What's the planned way for customers to turn the collection on/off?

guangyee:
> @guangyee don't we need a config option in suseconnect to enable the uptime collection? What's the planned way for customers to turn the collection on/off?

Per my understanding, in the short term, by installing and enabling the `uptime-tracker` service constitutes an opt-in to uptime collection. We'll document it as such. However, in the long run, we will consider the configuration aspect via REST API.

internal/connect/api.go
```diff
@@ -1,8 +1,15 @@
...
 import (
+	"bufio"
 	"encoding/json"
+	"errors"
 	"net/http"
+	"os"
+)
+
+const (
+	UptimeLogFilePath = "/etc/zypp/suse-uptime.log"
```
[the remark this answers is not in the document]
guangyee: This uptime log location is not meant to be configurable.

internal/connect/api.go
```diff
@@ -174,14 +181,45 @@ func updateSystem(body []byte) error {
 	return err
 }
 
+// readUptimeLogFile reads the system uptime log from a given file and
+// returns them as a string array. If the given file does not exist,
+// it will be interpreted as if the system uptime log feature is not
+// enabled. Hence an empty array will be returned.
+func readUptimeLogFile(uptimeLogFilePath string) ([]string, error) {
+	// NOTE: if uptime log file does not exist, we assume the uptime
```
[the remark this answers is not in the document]
guangyee: Per my understanding, in the short term, by installing and enabling the uptime-tracker service constitutes an opt-in to uptime collection. We'll document it as such. However, in the long run, we will consider the configuration aspect via REST API.

internal/connect/api.go
```diff
@@ -174,14 +181,45 @@ func updateSystem(body []byte) error {
...
+	fileScanner.Split(bufio.ScanLines)
+	var logEntries []string
+
+	for fileScanner.Scan() {
+		logEntries = append(logEntries, fileScanner.Text())
+	}
+	err = uptimeLogFile.Close()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
guangyee: Thanks for the suggestion. I'd considered using defer in the first incarnation. But opted not to in order to capture the error returned on file closure.

internal/connect/api.go
```diff
@@ -174,14 +181,45 @@ func updateSystem(body []byte) error {
...
+	return logEntries, nil
+}
+
 // makeSysInfoBody returns the JSON payload needed for the announce/update system calls
 func makeSysInfoBody(distroTarget, namespace string, instanceData []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
guangyee: Good point! I agree we should do this only for sending keep alive. I'll update the code.

internal/connect/api.go
```diff
@@ -195,6 +233,14 @@ func makeSysInfoBody(distroTarget, namespace string, instanceData []byte) ([]byt
 	payload.InstanceData = string(instanceData)
 	payload.Namespace = namespace
 
+	uptimeLog, err := readUptimeLogFile(UptimeLogFilePath)
+	if err != nil {
```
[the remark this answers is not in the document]
guangyee: Yes, we can take the pessimistic approach. Should we also introduce logging facility here to at least log the errors that we are ignoring to aid troubleshooting?

internal/connect/api_test.go
```diff
@@ -301,3 +303,54 @@ func TestProductMigrationsSMT(t *testing.T) {
 		t.Fatalf("Got ID: %d, expected: %d", gotID, expectedID)
 	}
 }
+
+func createTestUptimeLogFileWithContent(content string) (string, error) {
+	tempFile, err := ioutil.TempFile("", "testUptimeLog")
```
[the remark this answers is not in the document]
guangyee: Thanks for pointing that out. Will update the code.

internal/connect/api_test.go
```diff
@@ -301,3 +303,54 @@ func TestProductMigrationsSMT(t *testing.T) {
...
+	}
+	tempFilePath := tempFile.Name()
+	if _, err := tempFile.WriteString(content); err != nil {
+		os.Remove(tempFilePath)
+		return "", err
+	}
+	if err := tempFile.Close(); err != nil {
+		os.Remove(tempFilePath)
+		return "", err
+	}
```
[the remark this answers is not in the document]
guangyee: We do want to at least capture any errors if we can't close the temp file. Otherwise, we may end up with a corrupted file.

digitaltom:
> Per my understanding, in the short term, by installing and enabling the `uptime-tracker` service constitutes an opt-in to uptime collection. We'll document it as such. However, in the long run, we will consider the configuration aspect via REST API.

I don't think it's a good idea to assume customer agreement on collecting system uptime data by presence of a specific package. The package could already be installed on the image, or get pulled in unintentionally. 
Having it as an option in the suseconnect config also creates the base for receiving that setting from the server, and setting it intentionally in provisioning or manually. 
The target audience for metering are MSP systems. It's technically possible to enable it for 'normal' customers, but it shouldn't be the default and should be an informed decision to enable it for the customer.

guangyee:
> > Per my understanding, in the short term, by installing and enabling the `uptime-tracker` service constitutes an opt-in to uptime collection. We'll document it as such. However, in the long run, we will consider the configuration aspect via REST API.
> 
> I don't think it's a good idea to assume customer agreement on collecting system uptime data by presence of a specific package. The package could already be installed on the image, or get pulled in unintentionally. Having it as an option in the suseconnect config also creates the base for receiving that setting from the server, and setting it intentionally in provisioning or manually. The target audience for metering are MSP systems. It's technically possible to enable it for 'normal' customers, but it shouldn't be the default and should be an informed decision to enable it for the customer.

Sorry I wasn't clear earlier. It's not just the presence of the `uptime-tracker` package. Customer will also need to explicitly enable the `uptime-tracker.timer` service, which is similar to the keepalive heartbeat timer, in order to enable system uptime tracking. If that's not good enough, then I can also add a flag in SUSEConnect to enable the sending of system uptime log itself. But the drawback would be the increase in complexity in enabling the feature. Please let me know if that the direction we want to go.

digitaltom:
> Please let me know if that the direction we want to go. 

I think that needs  to get decided in the project planning regarding what is the desired UX enabling that feature. 
I would tend to having an option in the suseconnect config because that is very explicit.

jeremy-moffitt:
> > Please let me know if that the direction we want to go.
> 
> I think that needs to get decided in the project planning regarding what is the desired UX enabling that feature. I would tend to having an option in the suseconnect config because that is very explicit.

my understanding is the feedback is to have configurations for both connect-ng and the uptime tracker... I'm not sure why we want to require the customer to enable it twice, in addition to having the required packages and services configured. If we put a config option in suseconnect-ng, then for the tracker service itself, the presence of the package and the service running should be sufficient... That would look something like:
1. suseconnect has an explicit config
2. the uptime tracker is considered enabled if the service is running and does not additionally have a config

guangyee:
@digitaltom @jeremy-moffitt I will add a flag `enable_system_uptime_tracking` to control the sending of the system uptime tracking log.

internal/connect/api.go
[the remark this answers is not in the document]
guangyee: Will use a `enable_system_uptime_tracking` configuration to enable sending system uptime tracking log.

internal/connect/api.go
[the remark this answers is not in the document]
guangyee: Since we are going with an option to explicitly enable the feature. I think we should error out if we can't collect system uptime logs. At the very least, we should find a way to log the errors to help troubleshooting.

internal/connect/api.go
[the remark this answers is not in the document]
ngetahun: ok, I see

internal/connect/api.go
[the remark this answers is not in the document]
ngetahun: Logging is better as we are adding this information to an api request payload. The UX will suffer greatly if customers can't perform operations.

internal/connect/api_test.go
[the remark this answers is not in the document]
ngetahun: This is in a test scenario so that doesn't really matter.

internal/connect/api.go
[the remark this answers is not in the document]
guangyee: Will log it at `Info` level.

ngetahun (approved):

ngetahun (changes requested):
@guangyee `enable_system_uptime_tracking` is not shown as a cli option because it's missing from flag parsing logic [here](https://github.com/SUSE/connect-ng/blob/4ff33953224dbd20320e2edeca8fa5348cf58c81/suseconnect/suseconnect.go#L84). Please add it so I can manually test with the `uptime_tracker` package.

guangyee:
> @guangyee `enable_system_uptime_tracking` is not shown as a cli option because it's missing from flag parsing logic [here](https://github.com/SUSE/connect-ng/blob/4ff33953224dbd20320e2edeca8fa5348cf58c81/suseconnect/suseconnect.go#L84). Please add it so I can manually test with the `uptime_tracker` package.

Per my understanding, we will not be exposing this option in CLI as it is meant to control whether keepalive heartbeat will send the system uptime log. Sending system uptime log is only meaningful when working in conjunction with keepalive.

felixsch:
> Per my understanding, we will not be exposing this option in CLI as it is meant to control whether keepalive heartbeat will send the system uptime log. Sending system uptime log is only meaningful when working in conjunction with keepalive.

I agree on this. We spoke about this in one of the MSP syncs and all agreed there is no need for a CLI option. Maybe you can provide test instructions how to test this pull request?

ngetahun (commented):
@guangyee this PR is good to merge but need to do 2 things:
1. Rebase/merge from/unto `main`
2. add an entry into changelog for release 1.9.0 (`suseconnect-ng.changes`)

guangyee:
> @guangyee this PR is good to merge but need to do 2 things:
> 
> 1. Rebase/merge from/unto `main`
> 2. add an entry into changelog for release 1.9.0 (`suseconnect-ng.changes`)

Done. Thanks for the suggestion.

guangyee:
> > Per my understanding, we will not be exposing this option in CLI as it is meant to control whether keepalive heartbeat will send the system uptime log. Sending system uptime log is only meaningful when working in conjunction with keepalive.
> 
> I agree on this. We spoke about this in one of the MSP syncs and all agreed there is no need for a CLI option. Maybe you can provide test instructions how to test this pull request?

We have created some preliminary instructions for the MSPs on how to enable the feature. We can use that for our testing. https://confluence.suse.com/display/CloudSolutions/White+paper+documentation

build/packaging/suseconnect-ng.changes
```diff
@@ -1,3 +1,9 @@
+-------------------------------------------------------------------
+Wed Apr 15 12:37:29 UTC 2024 - Guang Yee <gyee@suse.com>
+
+- Update to version 1.9.0
+  * Add the ability to upload the system uptime logs, produced by the suse-uptime-tracker daemon, to SCC/RMT as part of daily heartbeat report.
```
ngetahun, line 5: Please add tracking cards here, either from bugzilla (bsc) or jira (ped). Look at past entries for suggestions

ngetahun (changes requested):

build/packaging/suseconnect-ng.changes
jeremy-moffitt: We have an epic on metering at PED-7982 , and the ECO at PED-8018... still working on the 3rd and final approval for the ECO if that matters.

guangyee: Thanks. I've added the two PEDs.

chajain:
Performed the end to end testing along with SCC latest PR https://github.com/SUSE/happy-customer/pull/7339 and RMT PR https://github.com/SUSE/rmt/pull/1091 and verified foloowing scenarios
- Downloaded the latest suse-uptime-tracker rpm package from OBS and started the suse-uptime-tracker.timer service
- When enable_system_uptime_tracking flag set to true, the payload contains the online_at params and system_uptimes table is populated on SCC when the keepalive hearbeat is sent

Also Validated backward and forward compatibility tests 
- older version of connect-ng still works with latest SCC/RMT PR
- upgraded connect-ng(this PR) talking to an older version of SCC/RMT which does not recognize system_uptime and it does not crash on the extra payload

internal/connect/api.go
```diff
@@ -176,14 +184,45 @@ func updateSystem(body []byte) error {
 	return err
 }
 
+// readUptimeLogFile reads the system uptime log from a given file and
+// returns them as a string array. If the given file does not exist,
+// it will be interpreted as if the system uptime log feature is not
+// enabled. Hence an empty array will be returned.
+func readUptimeLogFile(uptimeLogFilePath string) ([]string, error) {
```
mssola, line 191: A couple of notes. First of all, idiomatically the `uptimeLogFile.Close()` call should be done right after the Open with a `defer` statement. Hence:

```go
uptimeLogFile, err := os.Open(uptimeLogFilePath)
if err != nil {
	return nil, err
}
defer uptimeLogFile.Close()
```

This is important because you are supposed to check for `fileScanner.Err()` right after the `fileScanner.Scan()` loop and return early if there was any error when scanning. That is:

```go
for fileScanner.Scan() {
	logEntries = append(logEntries, fileScanner.Text())
}
if err = fileScanner.Err() {
	return nil, err
}
```

Also, I believe that the call `fileScanner.Split(bufio.ScanLines)` is not needed, no? Isn't splitting by lines the default behavior of Scanner?

internal/connect/api.go
```diff
@@ -197,6 +236,16 @@ func makeSysInfoBody(distroTarget, namespace string, instanceData []byte) ([]byt
 	payload.InstanceData = string(instanceData)
 	payload.Namespace = namespace
 
+	if includeUptimeLog {
+		uptimeLog, err := readUptimeLogFile(UptimeLogFilePath)
```
mssola, line 240: From the function implementation, it's pretty much guaranteed that if err == nil, then uptimeLog != nil, right? Then this can be further simplified to something like:

```go
if uptimeLog, err := readUptimeLogFile(UptimeLogFilePath); err != nil {
	util.Info.Print("Unable to system uptime log")
} else {
	payload.OnlineAt = uptimeLog
}
```

mssola: It might also be a good idea to include the error message as a debug line. That is, we make an effort inside of `readUptimeLogFile` to get the error and the caller does not use it at all 😄

internal/connect/api_test.go
```diff
@@ -300,3 +301,51 @@ func TestProductMigrationsSMT(t *testing.T) {
...
+	tempFile, err := os.CreateTemp("", "testUptimeLog")
+	if err != nil {
+		return "", err
+	}
+	tempFilePath := tempFile.Name()
+	if _, err := tempFile.WriteString(content); err != nil {
+		os.Remove(tempFilePath)
+		return "", err
+	}
+	defer tempFile.Close()
```
mssola, line 315: Maybe move it right after `CreateTemp`?

internal/connect/api_test.go
```diff
@@ -300,3 +301,51 @@ func TestProductMigrationsSMT(t *testing.T) {
...
+
+	return tempFilePath, nil
+}
+
+func TestUptimeLogFileDoesNotExist(t *testing.T) {
+	uptimeLogFileContent := ``
+	tempFilePath, err := createTestUptimeLogFileWithContent(uptimeLogFileContent)
```
mssola, line 322: Since `uptimeLogFileContent` is an empty string, couldn't you just call `createTestUptimeLogFileWithContent` with an empty string as an argument? Also, declaring an empty string with backquotes is really unusual in Go.

SUSEConnect.example
```diff
@@ -15,3 +15,6 @@
 
 ## Automatically agree to extension and module license confirmation prompts (default: false)
 # auto_agree_with_licenses: false
+
+## Enable system uptime tracking. The system uptime log will be sent to SCC/RMT as part of heartbeat.
```
mssola, line 19: Instead of `heartbeat`, call it `keepalive`, since this is how we call it in the documentation as well as in the flag itself.

build/packaging/suseconnect-ng.changes
```diff
@@ -7,6 +7,9 @@ Thu Apr 25 15:39:00 UTC 2024 - Felix Schnizlein <fschnizlein@suse.com>
   * Add --gpg-auto-import-keys flag before action in zypper command (bsc#1219004)
   * Include /etc/products.d in directories whose content are backed
     up and restored if a zypper-migration rollback happens. (bsc#1219004)
+  * Add the ability to upload the system uptime logs, produced by the
+    suse-uptime-tracker daemon, to SCC/RMT as part of daily heartbeat report.
```
mssola, line 11: Again: `heartbeat` -> `keepalive`. Moreover, I'd remove the `daily` because that's not guaranteed since customers can always disable the daily timer and that's ok.

internal/connect/api.go
```diff
@@ -176,14 +184,45 @@ func updateSystem(body []byte) error {
 	return err
 }
 
+// readUptimeLogFile reads the system uptime log from a given file and
+// returns them as a string array. If the given file does not exist,
+// it will be interpreted as if the system uptime log feature is not
+// enabled. Hence an empty array will be returned.
+func readUptimeLogFile(uptimeLogFilePath string) ([]string, error) {
+	// NOTE: if uptime log file does not exist, we assume the uptime
+	// tracking feature is not enabled
```
mssola, line 193: I'd maybe add a note for reference as to where this file comes from, and that the existence of this file is guaranteed when using "whatever external tool" in "that environment" 😉

mssola (changes requested):

SUSEConnect.example
guangyee: Done

build/packaging/suseconnect-ng.changes
guangyee: Done

internal/connect/api.go
guangyee: Good point! Yes, you're right, splitting is done by default.

internal/connect/api.go
guangyee: Done

internal/connect/api_test.go
guangyee: Good call. Done.

internal/connect/api_test.go
guangyee: Good point. I changed it to passing an empty directly instead.

mssola (approved):
LGTM! 🚀

PR #211, opened 2024-03-11
Allow rollback on readonly FS

josegomezr:
From @laenion explanation:

> on a transactional system the rollback operation should never
> require write permissions because the snapshot will always
> contain the correct information already.

Addresses bsc#1220679

### DIFF
--- a/suseconnect/suseconnect.go
+++ b/suseconnect/suseconnect.go
@@ -185,8 +185,11 @@ func connectMain() {
 	// Reading the configuration/flags is done, now let's check if the
 	// filesystem can handle operations from SUSEConnect for specific actions
 	// which require filesystem to be read write (aka writing outside of /etc)
-	// /etc is writable at any time, system token roation works just fine
-	if deRegister || rollback || cleanup {
+	// /etc is writable at any time, system token roation works just fine.
+	//
+	// Rollback *must* be allowed because is used as a synchonization mechanism
+	// in the transactional-update toolkit.
+	if deRegister || cleanup {
 		if err := connect.ReadOnlyFilesystem(connect.CFG.FsRoot); err != nil {
 			exitOnError(err)
 		}


suseconnect/suseconnect.go
```diff
@@ -186,7 +186,7 @@ func connectMain() {
 	// filesystem can handle operations from SUSEConnect for specific actions
 	// which require filesystem to be read write (aka writing outside of /etc)
 	// /etc is writable at any time, system token roation works just fine
-	if deRegister || rollback || cleanup {
+	if deRegister || cleanup {
```
digitaltom, line 189: It would be great to align the rules here with the comment above and https://confluence.suse.com/display/scc/SLE+Micro. 
It's not clear why de-register and cleanup need r/w access outside of /etc.

laenion: I haven't checked the code, but I was also wondering whether these rules are necessary at all... At least they didn't seem critical because the commands are not used in MicroOS' default scripts, so I didn't mention them.

josegomezr:
@laenion and I did a testing session of this change and have some findings (i'll be adding those to the confluence page too).

`--rollback`  is used _extensively_ by the `transactional-update` wrapper and by `rollback-helper` on boot.

Scenario: Migrate from 5.4 to 5.5

On the [transactional-update wrapper][tx-update@zypper-migration] `zypper migration` is called to perform the full migration process. By the end of `zypper migration` the system is in the next version. But in a transactional system, until there's a restart, there's no "next version". The system is still in the old state and the new state will be applied on the next reboot. (Or can be interacted with `transactional-update --continue <snapshot number>` ).

To circumvent this fact, [the script calls `suseconnect --rollback`][tx-update@connect-rollback] so the system is back on the original version. And marks the system for "registration on boot".

Following the reboot, just on boot time, (because of the previous "mark"), [`rollback-helper`'s rollback-reset-registration will try up to five times calling `suseconnect --rollback`][rollback-helper@connect-rollback]. And that will "reset" the registration accordingly to the target version 5.5.

[tx-update@zypper-migration]: https://github.com/openSUSE/transactional-update/blob/master/sbin/transactional-update.in#L1490
[tx-update@connect-rollback]: https://github.com/openSUSE/transactional-update/blob/master/sbin/transactional-update.in#L1524-L1533
[rollback-helper@connect-rollback]: https://github.com/openSUSE/rollback-helper/blob/master/sbin/rollback-reset-registration

digitaltom (approved):
Approving to unblock. Please don't forget to update the comment before that line and probably the confluence page with the findings. Also a link to that confluence page from the code comment would be helpful.

PR #218, opened 2024-04-02
Refactor zypper into its own package

felixsch:
Move `zypper` implementation to its own package to start separating implementations by its usage.

This does not implement any new features but should also not break anything!

part of: https://trello.com/c/dWLjq1Bk/3280-tedd-connect-source-is-structured-according-to-golang-standards

**Prepare and test hints:**

```
# Build the rpm (after building it is available in `artifacts/`)
$ docker run --rm -it -w /usr/src/connect-ng -v $(pwd):/usr/src/connect-ng registry.suse.com/bci/golang:1.21-openssl bash -c 'build/ci/build-rpm'

# Run a test env
$ docker run --rm -it -v $(pwd):/connect-ng registry.suse.com/bci/golang:1.21-openssl
> zypper in /connect-ng/artifacts/*.rpm

# Run a microos
# Note: This is a hint how to run Microos do not expect copy-paste work for your system!
$ curl -k "https://download.suse.de/install/SLE-Micro-6.0-Beta2/SLE-Micro.x86_64-6.0-Default-qcow-Beta2.qcow2" -O
$ sudo cp SLE-Micro.x86_64-6.0-Default-qcow-Beta2.qcow2 /var/lib/libvirt/images
$ virt-install --import --name slem-6.0-beta2 \
--memory 4048 --vcpus 2 --cpu host \
--disk SLE-Micro.x86_64-6.0-Default-qcow-Beta2.qcow2,format=qcow2,bus=virtio \
--network bridge=virbr0,model=virtio \
--os-variant=slem5.3 \
--graphics spice \
--noautoconsole

# In the vm
> passwd root
> echo 'PermitRootLogin yes' > /etc/ssh/sshd_config.d/allow-root.conf
> systemctl restart sshd
> ip addr
# HINT: Make a snapshot now, so you can rollback anytime!
```

**How to review this pull request:**

  1. Make sure it is building the binaries, libraries and the ruby bindings
  2. Test the implementation by running the following scenarios
      a) Register and de-register a system using the new build connect
      b) Activate a module/extension.. disable it again
      c) Run --status-text
      d) Run zypper migration and try to migrate (15SP5 to 15SP6?)
      e) Run zypper search-packages and make sure its working!
      f) Run Microos and register, deregister and finally rollback (validate the system in SCC)
  3. You are the best! Grab a chocolate from Felix desk ;)

As always, thank you very much for taking the time to review and please do not hesitates if you have trouble with the test setup or anything else! :rocket:

Thank you!


mssola:
One thing to note from this PR is that we are not able to run tests with non-root users. That is, if I run them with a non-root user, I get lots of:

```
open /etc/zypp/credentials.d/SCCcredentials: permission denied
```

I believe this is bad, but I don't recall right now if this was always the case. We should either make it explicit on the documentation test instructions because this is not self-evident, or try to mock this path to a local one or something like that.

go.mod
```diff
@@ -1,3 +1,3 @@
 module github.com/SUSE/connect-ng
 
-go 1.16
+go 1.21
```
mssola, line 3: This is actually a big change. Has this been handled in packaging?

internal/connect/client.go
```diff
@@ -5,6 +5,11 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+
+	"github.com/SUSE/connect-ng/internal/credentials"
+	cred "github.com/SUSE/connect-ng/internal/credentials"
```
mssola, line 10: You are importing this package twice? This shouldn't be the case.

internal/credentials/credentials.go
```diff
@@ -53,15 +55,15 @@ func curlrcCredentialsFile() string {
...
 }
 
-// getCredentials reads the system credentials from the SCCcredentials file
-func getCredentials() (Credentials, error) {
-	return ReadCredentials(systemCredentialsFile())
-}
+// // GetCredentials reads the system credentials from the SCCcredentials file
+// func GetCredentials(fsRoot string) (Credentials, error) {
+// 	return ReadCredentials(SystemCredentialsPath(fsRoot))
+// }
```
mssola, line 61: This should be removed, then.

internal/credentials/credentials.go
```diff
@@ -98,13 +100,14 @@ func parseCredentials(r io.Reader) (Credentials, error) {
...
-	Debug.Print("Writing credentials: ", c)
+	util.Debug.Print("Writing credentials: ", c)
 	path := c.Filename
-	if !filepath.IsAbs(path) {
-		path = filepath.Join(CFG.FsRoot, defaulCredentialsDir, path)
-	}
+	// FIXME : Make sure this idn't breaking anything
+	// if !filepath.IsAbs(path) {
+	// 	path = filepath.Join(fsRoot, defaulCredentialsDir, path)
+	// }
```
mssola, line 108: Can we test it in the scope of this PR? 😄

internal/credentials/credentials.go
```diff
@@ -129,34 +132,46 @@ func CreateCredentials(login, password, systemToken, path string) error {
...
+// func removeSystemCredentials(fsRoot string) error {
+// 	path := systemCredentialsFile(fsRoot)
+// 	return util.RemoveFile(path)
+// }
+
+// func removeServiceCredentials(serviceName string) error {
+// 	util.Debug.Print("Removing service credentials for: ", serviceName)
+// 	path := serviceCredentialsFile(serviceName)
+// 	return util.RemoveFile(path)
+// }
```
mssola, line 171: This looks pretty much WIP, no?

mssola (changes requested):
This is a really big PR to review! I might have missed things, but I have pointed out what I've seen so far. Thanks a lot for your hard work!

go.mod
felixsch: yes this is part of: https://bugzilla.suse.com/show_bug.cgi?id=1219185
I has been approved and we also changed it already in the spec file. This is more or less a remnant.

felixsch:
logging is everything but nice to use I agree. We had a discussion about it. But changing it now here is even more work and even more heavy to review. We decided to go step by step. Just moving out the zypper part resulted in this very big pull request! We should aim to improve this. I will add it as refactoring in the next TEDD epic?!

internal/credentials/credentials.go
felixsch: its gone!

internal/credentials/credentials.go
felixsch: This is a remainder for us to test this properly, since we want to move to absolute paths everywhere. We spent quiet some time to manually test this and looking through code for potential regressions where a relative path might sneak in but we weren't able to find one.

Sadly there is no unit test for these type of errors, or?

internal/credentials/credentials.go
felixsch: It does and its gone!

internal/credentials/credentials_test.go
```diff
@@ -28,78 +30,87 @@ func TestParseCredentials(t *testing.T) {
...
 			c.Username, c.Password)
 	}
-	if err := removeSystemCredentials(); err != nil {
+	if err := util.RemoveFile(sysCredsPath); err != nil {
 		t.Errorf("Unexpected error: %s", err)
 	}
-	path := systemCredentialsFile()
-	if fileExists(path) {
-		t.Error("File was not deleted: ", path)
+	// TODO : Move to util package
```
mssola, line 53: The comment is no longer needed.

internal/zypper/zypper.go
```diff
@@ -37,28 +42,58 @@ const (
...
+	// these are used by YaST
+	Description string `xml:"description"`
+}
+
+type ZypperService struct {
+	URL  string `xml:"url,attr"`
+	Name string `xml:"name,attr"`
+}
+
+// FIXME: see how we cna do this better
```
mssola, line 64: "cna" -> "can"

mssola (commented):
Just a couple of very minor things.

internal/credentials/credentials_test.go
likhitha77: It looks like we're testing if the `util.RemoveFile` method is doing it’s job correctly. But it makes more sense to have a test for the RemoveFile method itself within the util package instead of repeatedly checking if it works in this file. So just leaving a note to make sure we handle it in the next stages of refactoring

internal/zypper/zypper.go
likhitha77: Fixed it, thanks!

internal/credentials/credentials_test.go
likhitha77: Added more context in the commit message

mssola (approved):

PR #220, opened 2024-04-11
auto import keys flag comes before zypper action bsc#1219004

rtamalin:
When including the --gpg-auto-import-keys flag in the zypper command line, it must be placed before the action, e.g. ref, rather than after as it is a global flag.

Found while testing DMS migrations after switching to suseconnect-ng based zypper migration workflow.

### DIFF
--- a/internal/connect/zypper.go
+++ b/internal/connect/zypper.go
@@ -280,7 +280,7 @@ func RefreshRepos(version string, force, quiet, verbose, nonInteractive bool) er
 		args = append(args, "-f")
 	}
 	if CFG.AutoImportRepoKeys {
-		args = append(args, "--gpg-auto-import-keys")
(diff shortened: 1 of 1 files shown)


felixsch (approved):
lgtm :+1: 

- I build the branch and checked if `zypper-migration --gpg-auto-import-keys` is working with this change

image


rtamalin:
FYI in my latest round of testing that pulled this fix into a v1.8.0 based build, the DMS migration step completed successfully.

rtamalin:
I'll work on adding the feature tests today

rtamalin:
Closing in favour of PR#222 which is based upon a branch in the main repo which allows the feature testing CI to work, whereas it fails for PRs based upon personal forks.

PR #221, opened 2024-04-22
collectors: Add cpu and socket detection

felixsch:
card: https://trello.com/c/BiyCjXXG/3316-tedd-reimplement-cpu-information-gathering

Read values from `lscpu` for any architecture and remove s390x specific implementation since `util-linux` is available on all platforms now (in contrary to the initial implementation 10 years ago).

This also uses a more parsable output of `lscpu`. Check `lscpu -p` for more information.

**How to test this change:**

```
$ docker run --rm -it -h some-test-name -v $(pwd):/test registry.suse.com/bci/golang:1.21-openssl
> cd /test
> git config --global --add safe.directory /test
> make build
# expect: Check the output for the payload sent to SCC and make sure it is matching your cpu and socket count
```

**As always, if you have question, or need help testing this change, do not hesistate to reach out!**:rocket:

Thank you very much! :green_heart:


internal/collectors/cpu.go
```diff
@@ -2,32 +2,59 @@ package collectors
...
+
+	cpu, err1 := strconv.Atoi(last[0])
+	socket, err2 := strconv.Atoi(last[1])
+
+	if err1 != nil || err2 != nil {
+		return 0, 0
+	}
+
+	// We take the last line of the lscpu -p=cpu,socket
+	// output which is indicates the highest count number
```
mssola, line 41:
```suggestion
	// output which indicates the highest count number
```

mssola (approved):
Good job!

PR #222, opened 2024-04-22
auto import keys flag comes before zypper action bsc#1219004

rtamalin:
When including the --gpg-auto-import-keys flag in the zypper command line, it must be placed before the action, e.g. ref, rather than after as it is a global flag.

Found while testing DMS migrations after switching to suseconnect-ng based zypper migration workflow.

### DIFF
--- a/build/packaging/suseconnect-ng.changes
+++ b/build/packaging/suseconnect-ng.changes
@@ -4,6 +4,7 @@ Thu Apr 25 15:39:00 UTC 2024 - Felix Schnizlein <fschnizlein@suse.com>
 - 1.9.0 (unreleased)
   * Build zypper-migration and zypper-packages-search as standalone
     binaries rather then one single binary
+  * Add --gpg-auto-import-keys flag before action in zypper command (bsc#1219004)
 
 -------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


rtamalin:
Note that this is an updated version of the fix proposed in PR#220, this time proposed from a branch in the main repo, rather than a personal form, so the feature tests CI should be able to run now.

felixsch (approved):
lgtm :+1: 

I already tested this in the old pull request!

felixsch:
It would be nice to have a Changelog entry added for this bug

rtamalin:
> It would be nice to have a Changelog entry added for this bug

Happy to do so if there is a recommended procedure for doing so.

Looking at [suseconnect-ng.changes](build/packaging/suseconnect-ng.changes) it appears to be maintained as part of creating a specific version release.

rtamalin:
> > It would be nice to have a Changelog entry added for this bug
> 
> Happy to do so if there is a recommended procedure for doing so.
> 
> Looking at [suseconnect-ng.changes](build/packaging/suseconnect-ng.changes) it appears to be maintained as part of creating a specific version release.

Looking at @guangyee's PR#210 changes I see a changelog entry being added for a v1.9.0 release; I'm happy to add a similar entry if that is the best way to go forward?

PR #223, opened 2024-04-24
systemd unit files have moved to build/packaging

rtamalin:


rtamalin:
Note that the build in the feature test is failing because it appears to be using the v1.8.0 tagged sources, given that the spec, changes and unit files are still in the top level directory: https://github.com/SUSE/connect-ng/actions/runs/8820508369/job/24214295245?pr=223#step:5:186

digitaltom (approved):

digitaltom:
> Note that the build in the feature test is failing because it appears to be using the v1.8.0 tagged sources, given that the spec, changes and unit files are still in the top level directory: https://github.com/SUSE/connect-ng/actions/runs/8820508369/job/24214295245?pr=223#step:5:186

@felixsch shouldn't the feature tests use the sources from the current branch?

felixsch (approved):
lgtm :+1:

rtamalin:
> @felixsch shouldn't the feature tests use the sources from the current branch?

We figured it out - the `make dist` was setting up paths in the tarball in the old location, now fixed to put the systemd unit files in the correct location.

PR #224, opened 2024-04-25
TEDD: Implement other mandatory collectors

ngetahun:
- Add google uuid dep
- Implement hostname, memory and uuid collectors 
- Use new collectors in hwinfo payload creation 
- Add test fixtures

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,13 +4,14 @@ DIST          = $(NAME)-$(VERSION)/
 
 all: test build build-so
 
-dist: clean internal/connect/version.txt
+dist: download_deps clean internal/connect/version.txt
 	@mkdir -p $(DIST)
 	@cp -r internal $(DIST)
 	@cp -r third_party $(DIST)
 	@cp -r cmd $(DIST)
 	@cp -r docs $(DIST)
 	@cp go.mod $(DIST)
+	@cp go.sum $(DIST)
 	@cp LICENSE LICENSE.LGPL README.md $(DIST)
 	@cp SUSEConnect.example $(DIST)
 	@cp -r build/packaging/* $(DIST)
@@ -43,6 +44,9 @@ connect-ruby:
 
 gofmt:
 	@if [ ! -z "$$(gofmt -l ./)" ]; then echo "Formatting errors..."; gofmt -d ./; exit 1; fi
+
+download_deps:
+	go mod download
 
 build-so: out internal/connect/version.txt
 	go build -v -buildmode=c-shared -o out/libsuseconnect.so github.com/SUSE/connect-ng/third_party/libsuseconnect
--- a/internal/collectors/memory.go
+++ b/internal/collectors/memory.go
@@ -45,8 +45,8 @@ func parseMeminfo(file io.Reader) int {
 		return val / 1024
 	}
 
-	if scanner.Err() != nil {
-		util.Debug.Printf("could not obtain memory information for this system: %v", scanner.Err())
+	if err := scanner.Err(); err != nil {
+		util.Debug.Printf("could not obtain memory information for this system: %v", err)
 		return 0
 	}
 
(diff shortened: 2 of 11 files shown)


ngetahun:
[[TEDD] Add hostname, architecture, total_memory and UUID collectors to collectors package](https://trello.com/c/pBEw7lYd/3329-tedd-add-hostname-architecture-totalmemory-and-uuid-collectors-to-collectors-package)

internal/collectors/memory.go
```diff
@@ -0,0 +1,50 @@
...
+			continue
+		}
+
+		val, err := strconv.Atoi(fields[1])
+		if err != nil {
+			util.Debug.Printf("could not obtain memory information for this system: %v", err)
+			return 0
+		}
+		return val / 1024
+	}
```
mssola, line 46: Note, you after using `Scan()` you should always check on `Err()`. Thus, after the scan loop, you should do something like:

```go
if err := scanner.Err() {
// whatever
}
```

internal/connect/hwinfo.go
```diff
@@ -59,21 +62,20 @@ func getHwinfo() (hwinfo, error) {
...
+	hw.MemTotal = result["mem_total"].(int)
 
 	var lscpuM map[string]string
 	if hw.Arch == archX86 || hw.Arch == archARM || hw.Arch == archPPC {
 		if lscpuM, err = lscpu(); err != nil {
 			return hwinfo{}, err
 		}
-		hw.UUID, _ = uuid() // ignore error to match original
 	}
+	hw.UUID = result["uuid"].(string) // ignore error to match original
```
mssola, line 78: We should not discard this error, imho.

felixsch: this is only until we can implement the real collection implementation.

Makefile
```diff
@@ -4,13 +4,14 @@ DIST          = $(NAME)-$(VERSION)/
 
 all: test build build-so
 
-dist: clean internal/connect/version.txt
+dist: download_deps clean internal/connect/version.txt
```
mssola, line 7: Is this really needed? Could you explain why?

internal/collectors/memory.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if scanner.Err() != nil {
```
mssola, line 48:
```go
if err := scanner.Err(); err != nil {
// use `err` again instead of calling `scanner.Err()` again
}
```

internal/connect/hwinfo.go
```diff
@@ -59,21 +62,20 @@ func getHwinfo() (hwinfo, error) {
 	hw.Cpus = result["cpus"].(int)
 	hw.Sockets = result["sockets"].(int)
 
-	hw.Hostname = hostname()
+	hw.Hostname = result["hostname"].(string)
```
mssola, line 65: As noted before, leave a `TODO` note to tackle the possible error on casting here. I understand that we keep compatibility with the previous implementation, but it's actually quite bad.

internal/connect/hwinfo.go
```diff
@@ -59,21 +62,20 @@ func getHwinfo() (hwinfo, error) {
...
 	hw.CloudProvider = cloudProvider()
 
 	// Include memory information if possible.
-	if mem := systemMemory(); mem > 0 {
-		hw.MemTotal = mem
-	}
+
+	hw.MemTotal = result["mem_total"].(int)
```
mssola, line 70: Same here

internal/connect/hwinfo.go
mssola, line 78: And here

mssola (changes requested):

Makefile
ngetahun: This was a redundancy but will be cleaned up later.

mssola: Why not cleaning it now?

internal/collectors/memory.go
felixsch: addressed!

felixsch:
Can we have test instructions here?

mssola (approved):

PR #225, opened 2024-04-26
ci: added staticcheck and vet as linters

mssola:
Following Go's standards, we should be using `go vet` and `staticcheck` as well.


mssola:
Marked it as WIP because of the new cpu collector, which is not used and `staticcheck` rightfully detects its functions as not being used.

mssola:
> when we are ready with the collector then?

Yes, because right now the `staticcheck` is rightly complaining about functions not being used, like the ones from the CPU collector. Whenever we are using the new implementation, we can continue with this PR.

mssola:
In the end I had to split off the linting job into its new github action because running inside of the SLES container where the tests are run was basically impossible.

mssola:
As discussed, we will work on this in another way.

PR #229, opened 2024-04-29
Backup and restore /etc/products.d (bsc#1219004)

rtamalin:
The Distribution Migration System (DMS) implements a non-interactive
fully automated offline migration of systems by booting into a ISO
image that runs through the migration. In some cases, if a migration
from one product stream to another product stream fails, the normal
rollback handling doesn't always restore the products state correctly.
    
To address this we should include the /etc/products.d contents in the
backup that is created and restored.
    
Add changelog entry for upcoming v1.9.0 release stream, including the
bug tracking id.
    
Add unit test for the zypper Backup and Restore routines.
    
Also fixed two bugs found while developing the unit test:
    
  * CreateTarball()'s check for potential tar arguments incorrectly
    checked under the local file system hierarchy, rather than under
    the specified root hierarchy, meaning it could skip backing up
    entries that might exist under the root if they didn't exist
    locally.
    
  * The /var/adm/backup/system-upgrade directory created under the
    specified root hierarchy, and missing intermediary directories,
    were being created with no access permissions, meaning that only
    a privileged superuser could access them. In reality migrations
    are run as root so it didn't impact customer usage, but it did
    cause problems for the unit tests running as a non-privileged
    user.
    
Fixes: #231, #232
    
Co-authored-by: Felix Schnizlein <felixsch@users.noreply.github.com>

### DIFF
--- a/build/packaging/suseconnect-ng.changes
+++ b/build/packaging/suseconnect-ng.changes
@@ -4,6 +4,8 @@ Thu Apr 25 15:39:00 UTC 2024 - Felix Schnizlein <fschnizlein@suse.com>
 - 1.9.0 (unreleased)
   * Build zypper-migration and zypper-packages-search as standalone
     binaries rather then one single binary
+  * Include /etc/products.d in directories whose content are backed
+    up and restored if a zypper-migration rollback happens. (bsc#1219004)
 
 -------------------------------------------------------------------
 Wed Mar 13 12:37:29 UTC 2024 - José Gómez <1josegomezr@gmail.com>
--- a/internal/zypper/backup.go
+++ b/internal/zypper/backup.go
@@ -21,19 +21,23 @@ func createTarball(tarballPath, root string, paths []string) error {
 	// So we have to check this before.
 	var existingPaths []string
 	for _, p := range paths {
-		if !util.FileExists(p) {
+		// remove leading "/" from paths to allow using them from different root
+		candidatePath := strings.TrimLeft(p, "/")
+
+		// need to check for existence of the path under the specified root
+		rootedPath := path.Join(root, candidatePath)
+		if !util.FileExists(rootedPath) {
 			continue
 		}
-		// remove leading "/" from paths to allow using them from different root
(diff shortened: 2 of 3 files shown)


build/packaging/suseconnect-ng.changes
```diff
@@ -4,6 +4,8 @@ Thu Apr 25 15:39:00 UTC 2024 - Felix Schnizlein <fschnizlein@suse.com>
 - 1.9.0 (unreleased)
   * Build zypper-migration and zypper-packages-search as standalone
     binaries rather then one single binary
+  * Include /etc/products.d in directories whose content are backed
+    up and restored if a zypper-migration rollback happens.
```
felixsch, line 8:
```suggestion
    up and restored if a zypper-migration rollback happens. (bsc#1219004)
```
So maintenance can auto detect the fix for the bug :)

felixsch:
Can you please provide a test case how to test this change? Should be easy I assume but we have our internal ruling to try to provide at least one test case per pull request :)

build/packaging/suseconnect-ng.changes
rtamalin: Oops yes, will do ;-)

rtamalin:
> Can you please provide a test case how to test this change? Should be easy I assume but we have our internal ruling to try to provide at least one test case per pull request :)

Do you mean a feature test or a unit test? There don't appear to be existing unit or feature tests for backup or restore, though I may not be grokking things in the Ruby testing stuff properly...

rjschwei (approved):

felixsch (approved):
lgtm :+1:

- Ran a migration from `15.3` to `15.5` and checked that both:
  * `/var/adm/backup/system-upgrade/repos.sh`
  * `/var/adm/backup/system-upgrade/repos.tar.gz` exist
- Checked that repos.sh includes `products.d` directory
- Checked that repos.tar.gz includes `products.d/` directory content
- Let the migration rollback and check that everything as been reset

PR #233, opened 2024-05-06
Fix yast2-registration issue with unhandled ssl/tls

ngetahun:


third_party/libsuseconnect/libsuseconnect.go
```diff
@@ -110,8 +113,14 @@ func credentials(path *C.char) *C.char {
 
 //export create_credentials_file
 func create_credentials_file(login, password, token, path *C.char) *C.char {
+	credPath := C.GoString(path)
+
+	if !filepath.IsAbs(credPath) {
+		credPath = filepath.Join(cred.DefaultCredentialsDir, credPath)
```
josegomezr, line 119: why not use the helpers in `cred` ?

```suggestion
		credPath = cred.SystemCredentialsPath(credPath)
```

josegomezr (approved):

third_party/libsuseconnect/libsuseconnect.go
josegomezr: I guess that's what's happening nowadays, `fsRoot` is assumed to `/`, or am I missing a piece of context how the `.so` is used? :sweat_smile:

felixsch: You can change fsRoot to anything in suseconnect with `--root` but the library has no way to alter `fsRoot`.

ngetahun: `SystemCredentialsPath` only computes the global credential file `SCCcredentials` from filesystem root. I think there might be cases where `credPath` points to a different creds file. @felixsch might corroborate this better

build/packaging/suseconnect-ng.changes
```diff
@@ -1,13 +1,21 @@
...
 Thu Apr 25 15:39:00 UTC 2024 - Felix Schnizlein <fschnizlein@suse.com>
 
-- 1.9.0 (unreleased)
+- 1.10.0 (unreleased)
   * Build zypper-migration and zypper-packages-search as standalone
     binaries rather then one single binary
   * Add --gpg-auto-import-keys flag before action in zypper command (bsc#1219004)
   * Include /etc/products.d in directories whose content are backed
     up and restored if a zypper-migration rollback happens. (bsc#1219004)
+  * Fix certificate import for Yast when using a registration proxy with
```
ngetahun, line 10: This entry seems duplicated

third_party/libsuseconnect/libsuseconnect.go
josegomezr: I guess it wouldn't hurt to hardcode the *.so to `/` so it's explicit.

felixsch (approved):
lgtm :+1: 


**How I tested this pull request:**

- Checked the code changes if they make sense
- Contributed the missing parts about the credential file generation
- Tested migration (`yast2 migration`) between 15SP5 and 15SP6
- Tested registering against RMT and imported TTL certificates

**As always, if you think I missed something I should test, please let me know**:rocket:

PR #235, opened 2024-05-20
cmd/suseconnect: only allow Linux as the platform

mssola:
Even if this binary can be built and used in other platforms, we should not allow it because there are certain inner workings that already assume that we are in a Linux system (e.g. where configuration files are to be stored).


cmd/suseconnect/suseconnect.go
```diff
@@ -43,6 +44,10 @@ func (p *singleStringFlag) Set(value string) error {
 }
 
 func main() {
+	if runtime.GOOS != "linux" {
```
felixsch, line 47: can you add an comment linking https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63 so next person looking at this has the information at hand?

mssola: Done. I've linked the Go's source code directly.

felixsch (approved):
lgtm :+1:

PR #236, opened 2024-05-22
internal/collectors: cloudprovider detection

felixsch:
card: https://trello.com/c/XsnNZbC3/3362-tedd-cloud-provider-detection-in-the-new-collectors-framework

* Move the existing implementation to the right internal subpackage and lowercase results generated.
* Cleanup existing hwinfo implementation and tests to prepare for complete switch over to collectors subpackage.

**How to review this pull request:**

*Non cloud:*
```
$ docker run --rm --privileged -ti -v (pwd):/connect registry.suse.com/bci/golang:1.21-openssl
> zypper in -y dmidecode systemd
> make build
# expect: `cloud_provider` is empty string
> ./out/suseconnect -d
```

*AWS cloud:*
```
TBD
```

As always, if you have any question regarding how to test/review this merge request, please do not hesitate to contact me! :rocket:

Thank you very much for your review! :green_heart:


internal/collectors/cloudprovider.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+var (
+	cloudEx = `Version: .*(amazon)|Manufacturer: (Amazon)|Manufacturer: (Google)|Manufacturer: (Microsoft) Corporation`
+	cloudRe = regexp.MustCompile(cloudEx)
+)
+
+// findCloudProvider returns the cloud provider from "dmidecode -t system" output
+func findCloudProvider(b []byte) (string, bool) {
```
mssola, line 17: Can't we just assume that if an empty string is returned, then no cloud provider was detected? Maybe we can add a debug line to be extra sure, but returning a "found" bool is kind of weird in golang I think.

mssola (commented):
Generally looks good to me 👍🏻

internal/collectors/cloudprovider.go
felixsch: Is it, I thought the:

```
if x, ok := action(); ok {

}
```
semantic is pretty common, no?

mssola: Ok, and there's also the part of not sending empty strings, and this would ease up the implementation. Nevermind, then 👍🏻

mssola (approved):
LGTM 🎉

PR #238, opened 2024-05-22
internal/connect: switch to collectors package

felixsch:
card: https://trello.com/c/otFkJAIS/3363-tedd-switch-to-collector-framework

Switch to collectors package instead of using outdated `hwinfo` implementation.

*Simply running connect:*
```
$ docker run --rm --privileged -ti -v (pwd):/connect registry.suse.com/bci/golang:1.21-openssl
> zypper in -y dmidecode systemd
> make build
# expect: hwinfo information to be collected and fine.
# expect: to not see hypervisor or virtualization attributes set
> ./out/suseconnect -d
```

*KVM test case:*
```
tbd
```

*s390x test case:*
```
tdb
```

*AWS test case:*
```
tdb
```

As always, if you have any question regarding how to test/review this merge request, please do not hesitate to contact me! :rocket:

Thank you very much for your review! :green_heart:


internal/collectors/architecture.go
```diff
@@ -0,0 +1,19 @@
+package collectors
+
+import "github.com/SUSE/connect-ng/internal/util"
+
+type Architecture struct{}
+
+func (Architecture) run(arch string) (Result, error) {
+	// For now just a playback of the already existing value but in future
```
mssola, line 8: This confused me at first. We should write that the given `arch` actually comes from calling `DetectArchitecture` beforehand. Hence, I would also write the origin of this value, and then adding the note as it is that in the future we might further expand this info.

internal/collectors/collectors.go
```diff
@@ -35,19 +36,31 @@ const (
...
-			return NoResult, err
+			util.Debug.Printf("Collecting system information failed: %s", err)
 		}
 		maps.Copy(obj, res)
 	}
 
 	return obj, nil
 }
+
+func FromResult[R any](result Result, key string, def R) R {
```
mssola, line 56: Oh, generics, fancy 😉 ! Could you document this?

internal/collectors/virtualization.go
```diff
@@ -22,14 +22,10 @@ func (Virtualization) run(arch Architecture) (Result, error) {
 
 	output, err := util.Execute([]string{systemdDetectVirtExecutable, "-v"}, []int{0, 1})
 
-	if err != nil {
+	if err != nil || string(output) == "none" {
```
mssola, line 25: Just realized about this `"none"` string, which is pretty wild. Could you document where this is coming from?

mssola (changes requested):

internal/collectors/architecture.go
felixsch: do you think it makes sense to actually call `DetectArchitecture` again?

mssola: Not really, because we really need to know the arch before collecting anything, right? So it's fine, but we should document it anyways just so a newcomer is not confused in this file.

internal/collectors/virtualization.go
felixsch: I added a comment addressing where the "none" stems from.

internal/collectors/collectors.go
felixsch: Added a comment about what it does

internal/collectors/architecture.go
felixsch: I updated the comments above mentioning that `DetectArchitecture` is called before already

mssola (approved):
lgtm 🎉

PR #240, opened 2024-05-23
Be more specific on the supported ARM architecture

mssola:
The GOARCH and the GOARM environment variables being provided were conflicting: if we are using `arm64`, then GOARM is ignored. If we really wanted `v7`, then we should use GOARCH=arm and GOARM={5, 6, 7}. See: https://go.dev/wiki/GoArm for further documentation on ARM.

To align the development environment to what we actually ship through OBS, I have been more explicit on using arm64 v8, which is the latest version from the ARM family of microprocessors. I also thought on allowing this to be configureable, just in case a developer wanted to mess with older ARM devices, but we actually introduced some 64-bit dependencies when we added magic numbers like `BTRFS_SUPER_MAGIC`.

Because all of this, I have added further comments just in case we decide to go the extra mile for this architecture in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,10 @@ feature-tests:
 test-yast: build-so
 	docker build -t go-connect-test-yast -f third_party/Dockerfile.yast . && docker run -t go-connect-test-yast
 
+# This "arm" means ARM64v8 little endian, the one being delivered currently on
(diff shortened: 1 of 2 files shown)


felixsch (approved):
lgtm :+1:

PR #247, opened 2024-06-18
collectors: fallback to machine ID on ARM64

mssola:
In some devices (e.g. non-ACPI compatible ones) the UUID will not be available through the default way. In these scenarios just fallback to using the value on `/etc/machine-id` just like we do for s390x.

## How to test

1. Pick up a non-ACPI compliant ARM64 machine.
2. Build SUSEConnect there.
3. Now if you do `SUSEConnect --debug --keepalive`, the `uuid` parameter will be set (before it was not).

### DIFF
--- a/internal/collectors/uuid.go
+++ b/internal/collectors/uuid.go
@@ -18,9 +18,20 @@ type UUID struct{}
 func (UUID) run(arch string) (Result, error) {
 	var uuid string
 	var err error
+
 	switch arch {
+	case ARCHITECTURE_ARM64:
+		uuid, err = uuidDefault()
+
+		// This can happen because non-ACPI compliant ARM64 devices will provide
+		// an empty output from `dmidecode` (i.e. just a comment explaining that
+		// it's not supported). In this cases, just fallback to using the
+		// machine-id value as we do for s390x.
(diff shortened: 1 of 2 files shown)


mssola:
[[CAT] Fix UUID on aarch64 without ACPI support](https://trello.com/c/NKSB46bj/3411-cat-fix-uuid-on-aarch64-without-acpi-support)

internal/collectors/uuid.go
```diff
@@ -18,9 +18,15 @@ type UUID struct{}
 func (UUID) run(arch string) (Result, error) {
 	var uuid string
 	var err error
+
 	switch arch {
+	case ARCHITECTURE_ARM64:
+		uuid, err = uuidDefault()
+		if err == nil && uuid == "" {
```
felixsch, line 25: Can you add a comment here why this is happening. Aka describing the non ACPI platforms do not provide a `uuid` via `dmidecode`

felixsch (approved):
lgtm :+1: 

- Checked the code for improvements I would suggest,
- Ran the changes on `aarch64` and checked that it falls back to `/etc/machine-id` in case `dmidecode` does not provide a value.

cheers,

PR #254, opened 2024-07-01
Fix UUID gathering on some ARM64 devices

mssola:
On non-ACPI compatible ARM64 machines having `dmidecode` installed does not make any sense (it's not supported and the tool will simply print a warning about it when called). Hence, only try to gather the UUID through `dmidecode` if installed, otherwise the potential error will not make much sense on these other machines where this tool is not installed.

### How to test

Build SUSEConnect on a non-ACPI compliant device (I used my Rasperry Pi 5 but I'm sure we can pick up a similar machine from inside SUSE).

**Before this patch**: no UUID was sent if `dmidecode` was not installed. Check that by running it with `--debug` before installing `dmidecode` and realize how no UUID is sent. Then install the binary and see that is fixed.

**After this patch**: it doesn't matter if `dmidecode` is installed or not, a UUID is given.

### DIFF
--- a/internal/collectors/uuid.go
+++ b/internal/collectors/uuid.go
@@ -21,13 +21,22 @@ func (UUID) run(arch string) (Result, error) {
 
 	switch arch {
 	case ARCHITECTURE_ARM64:
-		uuid, err = uuidDefault()
+		// ARM machines don't necessarily have `dmidecode` installed (i.e. on
+		// non-ACPI devices it does not even make sense to have it installed).
+		// Thus, we will first try to fill the uuid if such an executable even
+		// exists.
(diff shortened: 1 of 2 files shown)


mssola:
[[CAT] Fix UUID gathering when `dmidecode` is not installed on ARM64 devices](https://trello.com/c/zdnjAcis/3432-cat-fix-uuid-gathering-when-dmidecode-is-not-installed-on-arm64-devices)

ngetahun (approved):
LGTM!

mssola:
Feature tests seem temporarily broken (unrelated to this PR). I'll merge this PR and take a closer look at these tests later.

PR #255, opened 2024-07-01
Grab the home directory from /etc/passwd if needed

mssola:
Go's `os.UserHomeDir` has one big gotcha: it only fetches the value from the HOME environment variable and that's it: it does not try to fetch this value from anywhere else if that environment variable is not set.

In some conditions (e.g. a systemd service file) this environment variable might not be set and hence some important paths will not be properly generated. This commit ensures that the home directory is picked: if Go's `os.UserHomeDir` fails, then we will try to read this value from `/etc/passwd`.

Fixes bsc#1226128

### DIFF
--- a/internal/credentials/credentials.go
+++ b/internal/credentials/credentials.go
@@ -47,12 +47,7 @@ func ServiceCredentialsPath(service string, fsRoot string) string {
 }
 
 func CurlrcCredentialsPath() string {
-	home, err := os.UserHomeDir()
-	if err != nil {
-		// TODO: handle error? log?
-		return ""
-	}
-	return filepath.Join(home, CurlrcUserFile)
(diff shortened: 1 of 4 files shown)


mssola:
[L3: allow SUSEConnect to properly load the `.curlrc` file even if HOME is not set](https://trello.com/c/0wxEMbKF/3434-l3-allow-suseconnect-to-properly-load-the-curlrc-file-even-if-home-is-not-set)

ngetahun (approved):
LGTM!

mssola:
Feature tests seem temporarily broken. I'll merge this PR and take a closer look at these tests later.

albertvaka:
Maybe this could be contributed upstream to Go?

PR #253, opened 2024-06-26
Correct CPU information gathering on ppc64

mssola:
## How to test

1. Run `make build-ppc64le`.
2. Upload the resulting binary into a PowerPC machine (ask me for access to one).
3. Run SUSEConnect with `--debug` and see new parameters like `device_tree` or `lpar_mode` being sent inside of `arch_specs`.

PR #256, opened 2024-07-04
 Add another lookup path for devicetree

mssola:
Apparently devicetree on emulated ARM64 will save the `compatible` file into another path under `hypervisor`. This commit extends the support for this file so multiple paths can be provided.

~~Depends on #253.~~ (done!)


likhitha77 (approved):
Works for me!
I tested this in docker container for mac where the linux images are emulated. The devicetree is detected correctly!

Screenshot 2024-07-09 at 10 21 38 PM

PR #258, opened 2024-07-08
ci: be explicit on the commands for feature tests

mssola:
This way we can further debug which variables are actually set for the commands being run.


felixsch (approved):
lgtm :+1:

mssola:
🎉 🎉 🎉

PR #259, opened 2024-07-10
Update changelog for 1.11 release

mssola:


build/packaging/suseconnect-ng.changes
```diff
@@ -1,8 +1,14 @@
 -------------------------------------------------------------------
-Tue Aug 15 15:49:20 UTC 2024 - Felix Schnizlein <fschnizlein@suse.com>
-
-- Update version to 1.11 (unreleased)
-  * Add detection for container runtimes.
+Thu Jul 10 12:03:20 UTC 2024 - Miquel Sabaté <msabate@suse.com>
+
+- Update version to 1.11
+  - Added uname as collector (8858fc1d7693, 191d782054a9 and b99d42fda31e)
```
felixsch, line 5: do we need the sha sums for the individual commits?

mssola: Came for free with the git log command I used 😄 But if you want we can remove it, np

felixsch: I would remove them, they kinda confusing in the context of a spec file I would argue. But in the end your call!

felixsch (approved):
lgtm :+1:

build/packaging/suseconnect-ng.changes
mssola: Removed 😉

PR #260, opened 2024-07-11
Add a command to show the info being gathered

mssola:
## Why

This was a nice to have feature for our latest epic on information gathering. It's about being as transparent as possible to users as to what kind of information is being gathered by SUSEConnect. This is already available with the `--debug` flag on, but this is a more user-friendly approach.

## How

The PR simply exports the `FetchSystemInformation` function and uses it. The output is always in JSON because it is more transparent to also show this in the format it's delivered/used in the end.

## How to test

Simply running with `suseconnect --info` or `suseconnect -i`, and you should get a proper output. Pair it with `jq` and you'd get an extra dope output.


felixsch (approved):
lgtm :+1:

```shell
gondor:/connect/out # ./suseconnect --info | jq
{
  "arch": "x86_64",
  "container_runtime": "docker",
  "cpus": 8,
  "hostname": "gondor",
  "mem_total": 13857,
  "sockets": 1,
  "uname": "6.9.7-1-default #1 SMP PREEMPT_DYNAMIC Fri Jun 28 05:50:47 UTC 2024 (a5efffa)"
}
```

- Checked the code changes for improvements I would suggest
- Ran the code changes and checked the command exists and works as expected

**As always, if you think I missed something I should review, please let me know!**:rocket:

cheers,

PR #262, opened 2024-08-05
POC for integration telemetry into SUSEConnect

mssola:
**DO NOT MERGE THIS!**

## How to test this

1. Download both https://github.com/SUSE/telemetry and https://github.com/SUSE/telemetry-server (yes, you need both).
2. You can deploy the server in a variety of ways. I've just gone the bare-metal route:
  - `cd telemetry-server/server/telemetry-server`
  - `mkdir -p /tmp/telemetry/{client,server} /tmp/susetelemetry`
  - `go run . --config ../../testdata/config/localServer.yaml`
3. On another terminal:
  - `cd telemetry` (NOTE: this is the client repo)
  - `cp examples/config/telemetry.yaml /tmp/susetelemetry/config.yaml`

And now you can go on with SUSEConnect the usual way. I've done it through a container. Hence, from the source code of suseconnect-ng, go to this branch and:

```
$ docker run --network=host --rm --privileged -ti -v $(pwd):/connect -v /tmp/susetelemetry:/tmp/susetelemetry  registry.suse.com/bci/golang:1.21-openssl
```

From there go to `/connect`, build the code with `make build` and do something like `./out/suseconnect -r <regcode>`. You will see messages of telemetry data being sent.


rtamalin:
Note that if you have docker installed you can use `make docker-start` from the top-level directory of `telemetry-server` to start a local server, and `make docker-stop` to stop it.
Similarly if you have a viable `docker compose` framework locally, you can run `make compose-start` to start a telemetry-server + postgres based deployment, and correspondingly `make compose-stop` to stop it.
Work is on-going to provide equivalent kubernetes deployment (defaulting to local k3s) integration, if that is of interest.
See the [SUSE/telemetry-server/README.md](https://github.com/SUSE/telemetry-server/README.md) for more details.

internal/connect/telemetry.go
```diff
@@ -0,0 +1,32 @@
...
+	"log/slog"
+
+	"github.com/SUSE/telemetry"
+)
+
+func pushTelemetry(content []byte) {
+	telemetryType := telemetry.TelemetryType("SCC-CONNECT-MANDATORY")
+	class := telemetry.MANDATORY_TELEMETRY
+	tags := telemetry.Tags{"mandatory", "scc", "connect"}
+	flags := telemetry.GENERATE | telemetry.SUBMIT
```
rtamalin, line 13: Note that the telemetry.SUBMIT flag should only really be used by an application that wants the telemetry to be immediately submitted without relying on the background submission agent (systemd timer triggered) to bundle up and send the telemetry at a later date.

Additionally if you plan to generate multiple telemetry data items, and want to submit immediately, only the last Generate() call needs to include the SUBMIT flag.

internal/connect/telemetry.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"log/slog"
+
+	"github.com/SUSE/telemetry"
+)
+
+func pushTelemetry(content []byte) {
+	telemetryType := telemetry.TelemetryType("SCC-CONNECT-MANDATORY")
+	class := telemetry.MANDATORY_TELEMETRY
+	tags := telemetry.Tags{"mandatory", "scc", "connect"}
```
rtamalin, line 12: You probably don't need to include these tags, since they are redundant, given that this information is derivable from the telemetry type you have chosen.

It is ok for tags to be left empty by telemetry generating applications; they are intended more for specifying ancillary grouping information, potentially set up by the product registration, e.g. SCC connection style of direct vs RMT vs Cloud BYOS, or maybe to track a customer deployment being an MSP (or something like that) generating uptime tracking info, or anything else that may be useful in the long run...

internal/connect/telemetry.go
```diff
@@ -0,0 +1,43 @@
...
+	"log/slog"
+	"path/filepath"
+
+	"github.com/SUSE/connect-ng/internal/util"
+	"github.com/SUSE/telemetry"
+	"github.com/SUSE/telemetry/pkg/client"
+)
+
+func pushTelemetry(content []byte) {
+	// This configuration file is to be provided by an external package.
```
rtamalin, line 13: This can be better handled using the [Status()](https://github.com/SUSE/telemetry/blob/e727356476d90c2976cbd9ed41d272175a19e371/telemetry.go#L169) interface call, with a check like:

```
        if telemetry.Status() != telemetry.CLIENT_REGISTERED {
                return
        }
```

This will skip submitting telemetry until the telemetry client is able to register with the upstream telemetry server.

Or if you want the equivalent of this check it would be:

```
        if telemetry.Status() != telemetry.CLIENT_CONFIG_ACCESSIBLE {
                return
        }
```

mssola:
[[TSC] Implement a POC that integrates SUSE/telemetry into SUSEConnect.](https://trello.com/c/4fgeQtb8/3494-tsc-implement-a-poc-that-integrates-suse-telemetry-into-suseconnect)

PR #267, opened 2024-09-23
Make SUSEConnect -d error message on Public Cloud generic

rjschwei:
SUSEConnect -d is disabled on Public Cloud instances when registrecloudguest is present. This is to ensure that the correct de-registration action happens whether the system is connected to SCC or the update infrastructure. Initially this only mattered for PAYG. However with BYOS instances being able to connect to the update infrastructure we need the message to not refer to PAYG anymore.


mssola:
What about a more generic message just in case we do something different in the future? Something like `De-registration has been disabled on this context.` (or better phrased...)

rjschwei:
> What about a more generic message just in case we do something different in the future? Something like `De-registration has been disabled on this context.` (or better phrased...). Possibly with a link to documentation pointing at scenarios where de-register is disabled? (I don't know if that even exists)

Well, while nothing is impossible, replacing registercloudguest would be a rather large lift given its 10+ year history in our images. The idea behind using the command in the message is to avoid a really long explanation that would have to have a link to documentation which changes quite frequently compared to `registercloudguest`

mssola:
Fair enough. Could you rebase your commit given that there are conflicts now?

rjschwei:
> Fair enough. Could you rebase your commit given that there are conflicts now?

Done.

Not sure about the test failure. And will need help with that

mssola:
The error is basically that you need to setup a bunch of environment variables on your fork in order to run feature tests. See: https://github.com/SUSE/connect-ng?tab=readme-ov-file#testing.

This is quite tedious, but I happen to also have them locally. I just pulled your branch locally and feature tests happen to work. Hence, I will simply skip this test on github and merge it right away.

Thanks for your contribution!

PR #268, opened 2024-10-07
Feature addreg code ways

paragjain0910:
# SUSE Connect - Registration Code Input Handling

### Support for Two New Ways to Pass Registration Codes

1. **Standard Input (stdin)**: 
   - By using `-r -`, you can pass the registration code directly from stdin.
   - Example:
     ```bash
     echo "1234" | go run suseconnect.go -r -
     ```

2. **File Input**: 
   - By using `-r @/file/path`, you can specify a file path that contains the registration code.  The current directory is considered to be same as of the suseconnect.go file. Pass the path reference accordingly.
   - Example:
     ```bash
     go run suseconnect.go -r @/my/file
     ```

## `processToken` Function Documentation

The `processToken` function is a part of the `suseconnect` package located in the `cmd/suseconnect/suseconnect.go` file. This function is responsible for processing the token input. The token can be:

- A string (the registration code directly),
- A file path prefixed with `@`, or 
- `-` indicating that the registration code should be read from stdin.

Added the Test cases as well.
### Function Signature

```go
func processToken(token string) (string, error)

### DIFF
--- a/cmd/suseconnect/suseconnect.go
+++ b/cmd/suseconnect/suseconnect.go
@@ -437,7 +437,7 @@ func readTokenFromReader(reader io.Reader) (string, error) {
 	if err != nil && err != io.EOF {
 		return "", fmt.Errorf("failed to read token from reader: %w", err)
 	}
-	token := strings.TrimSuffix(tokenBytes, "\n")
+	token := strings.TrimSpace(tokenBytes)
 	if token == "" {
 		return "", fmt.Errorf("error: token cannot be empty after reading")
 	}
--- a/cmd/suseconnect/suseconnect.go
+++ b/cmd/suseconnect/suseconnect.go
@@ -149,15 +149,7 @@ func main() {
 		connect.CFG.Namespace = namespace
 		writeConfig = true
 	}
-	if token != "" {
-		connect.CFG.Token = token
-		processedToken, processTokenErr := processToken(token)
-		if processTokenErr != nil {
-			util.Debug.Printf("Error Processing token %+v", processTokenErr)
-			os.Exit(1)
-		}
-		connect.CFG.Token = processedToken
-	}
+	parseRegistrationToken(token)
 	if product.isSet {
 		if p, err := connect.SplitTriplet(product.value); err != nil {
 			fmt.Print("Please provide the product identifier in this format: ")
@@ -310,6 +302,18 @@ func main() {
 	}
 }
 
+func parseRegistrationToken(token string) {
+	if token != "" {
+		connect.CFG.Token = token
+		processedToken, processTokenErr := processToken(token)
+		if processTokenErr != nil {
+			fmt.Printf("Error Processing token %+v", processTokenErr)
+			os.Exit(1)
+		}
+		connect.CFG.Token = processedToken
+	}
+}
+
 func maybeBrokenSMTError() error {
 	if !connect.URLDefault() && !connect.UpToDate() {
 		return fmt.Errorf("Your Registration Proxy server doesn't support this function. " +
@@ -425,6 +429,7 @@ func readTokenFromReader(reader io.Reader) (string, error) {
 		return "", fmt.Errorf("failed to read token from reader: %w", err)
 	}
 	token := strings.TrimSuffix(tokenBytes, "\n")
+	token = strings.TrimSpace(token)
 	if token == "" {
 		return "", fmt.Errorf("error: token cannot be empty after reading")
 	}
--- a/cmd/suseconnect/suseconnect.go
+++ b/cmd/suseconnect/suseconnect.go
@@ -403,8 +403,6 @@ func isSumaManaged() bool {
 }
 
 func processToken(token string) (string, error) {
-	var reader io.Reader
-
 	if strings.HasPrefix(token, "@") {
 		tokenFilePath := strings.TrimPrefix(token, "@")
 		file, err := os.Open(tokenFilePath)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mssola** [comment]
> @mssola Do we plan to have the cli implementation reworked? If this is the case (what I hope) we should maybe wait if this functionality is not provided by the library itself (idk, https://github.com/urfave/cli)

That is true... That being said, when do we want to do such a thing? Is there any epic you think that could work out? If so, let's talk about it during the workshop next week and plan things.

**Comment by paragjain0910** [comment]
I was going through the documentation of urfave. Here the key points :

1. It seems to be a more declarative way to simply managing commands, subcommands and flags. 
2. Considering the performance and the size of the urfave binary, I feel there will be no significant performance impact.
3.  It can easily support existing cli features like such as aliases, subcommands, and flexible argument handling like -r @/filepath, -r -, and -r 1234)  
4. If there are plans to scale suseconnect-ng or add more complex CLI interactions in the future, refactoring with urfave/cli could save time and effort down the line.

Also I tried creating a sample go file using urfave for supporting the above registration code features. @felixsch you can also have a look once, maybe it can help in reflecting  upon the effort required to bring the change.
[main.txt](https://github.com/user-attachments/files/17303117/main.txt)


Given these points, I would recommend reworking the CLI implementation using urfave/cli.

**Comment by mssola** [comment]
@paragjain0910 thanks for the pointers. I have had some experience with https://github.com/urfave/cli in the past (e.g. back when I was involved with https://github.com/SUSE/zypper-docker, even if that was an old version). So far I believe that it would be the right choice.

That being said, this would be a big change and I believe that the workshop next week can clarify when and how to apply this change. Moreover, I'd even envision a SUSEConnect which can break backwards compatibility, and have a proper action-like system in which you would `SUSEConnect register <blabla>` instead of the mess we have right now.

So, food for thought, @felixsch 😄

File: cmd/suseconnect/suseconnect.go [hunk 1]
```diff
@@ -390,3 +401,37 @@ func fileExists(path string) bool {
...
+	var reader io.Reader
+
+	if strings.HasPrefix(token, "@") {
+		tokenFilePath := strings.TrimPrefix(token, "@")
+		file, err := os.Open(tokenFilePath)
+		if err != nil {
+			return "", fmt.Errorf("failed to open token file '%s': %w", tokenFilePath, err)
+		}
+		defer file.Close()
+		reader = file
```
mssola, line 415: Just

```go
return readTokenFromReader(file)
```

cmd/suseconnect/suseconnect.go
```diff
@@ -390,3 +401,37 @@ func fileExists(path string) bool {
...
+	} else if token == "-" {
+		reader = os.Stdin
```
mssola, line 417: And here:

```go
return readTokenFromReader(os.Stdin)
```

cmd/suseconnect/suseconnect.go
```diff
@@ -390,3 +401,37 @@ func fileExists(path string) bool {
...
+	} else {
+		return token, nil
+	}
+
+	return readTokenFromReader(reader)
```
mssola, line 422: This way this return and the `reader` declaration can be removed.

cmd/suseconnect/suseconnect.go
```diff
@@ -390,3 +401,37 @@ func fileExists(path string) bool {
...
+}
+
+func readTokenFromReader(reader io.Reader) (string, error) {
+	bufReader := bufio.NewReader(reader)
+	tokenBytes, err := bufReader.ReadString('\n')
+	if err != nil && err != io.EOF { // Handle potential errors, including EOF
```
mssola, line 428: Don't inline comments, put them before the line instead. Hence:

```go
// Handle potential errors, including EOF
if err != nil && err != io.EOF {
```

cmd/suseconnect/suseconnect.go
```diff
@@ -390,3 +401,37 @@ func fileExists(path string) bool {
...
+		return "", fmt.Errorf("failed to read token from reader: %w", err)
+	}
+	token := strings.TrimSuffix(tokenBytes, "\n")
```
mssola, line 431: Note that `strings.TrimSpace` below also trims out `\n` characters. More generally, it removes all characters defined by [unicode.IsSpace](https://pkg.go.dev/unicode#IsSpace). Hence, you can remove this line.

mssola (commented):
It mostly looks good to me, good job! Just fix my couple of suggestions.

cmd/suseconnect/suseconnect.go
paragjain0910: done removed the reader.

cmd/suseconnect/suseconnect.go
paragjain0910: added the above

cmd/suseconnect/suseconnect.go
paragjain0910: removed the reader

cmd/suseconnect/suseconnect.go
paragjain0910: removed the inline comment. not required as such to put.

cmd/suseconnect/suseconnect.go
paragjain0910: got it understood. removed the strings.trimspace.

mssola: No. The other way around, what I meant is that `TrimSuffix` is not needed, but `TrimSpace` is enough and it *also* removes `\n` characters because it's in `unicode.IsSpace`.

mssola (approved):

PR #270, opened 2024-10-08
Do not generate docker configuration on pubcloud

mssola:
Fixes bsc#1231185


mssola:
[SUSEConnect: do not create docker configuration when being used behind a proxy bsc#1231185](https://trello.com/c/BwGikLFC/3575-suseconnect-do-not-create-docker-configuration-when-being-used-behind-a-proxy-bsc1231185)

internal/connect/config.go
```diff
@@ -26,6 +26,15 @@ const (
 	defaultEnableSystemUptimeTracking = false
 )
 
+// Kinds of servers which are supported by SUSEConnect.
+type ServerType uint64
+
+const (
+	Unknown ServerType = iota
```
felixsch, line 33: Can we prefix this somehow? aka. `ApiProviderSCC` or `ApiProviderRMT` to avoid potential clashes?

mssola: Done!

felixsch (approved):
lgtm :+1:

- Checked the code changes to suggest improvements
- Built connect and ran against a RMT and SCC to checked that the credentials are set accordingly

**As always, if you think I missed something, I should look into, please let me know!**:rocket:

cheers,

mssola:
@felixsch check again. Now tests are passing. Note that I have added a couple of other paths in which to double check the "server type". I have tested it locally and it appears to work, but double check just to be sure 😉

felixsch:
> @felixsch check again. Now tests are passing. Note that I have added a couple of other paths in which to double check the "server type". I have tested it locally and it appears to work, but double check just to be sure 😉

Nice will do. Do you mind adding it to labels as well? Then I do not need to do another PR :)

felixsch (approved):
lgtm :+1: 

- Checked the code changes and see it working when:
  * Running against SCC and check if the file is placed and labels are set
  * Running against RMT and check that the file is not placed and labels are not sent
- Checked the changes and added improvements if I think they would make sense

**As always, if you think I missed something I should test please let me know!**:rocket:

cheers,

PR #271, opened 2024-10-25
Honor auto-import-gpg-keys flag on migration

mssola:
## Description

The parsing of CLI arguments and how these are passed through the internal SCC client, zypper and other tools is not ideal and needs to be re-worked quite urgently. This is why other silly bugs have appeared in which we are not passing the correct arguments to our backend code.

Until this rework is not done, let's simply apply this quick fix which simply passes the `--gpg-auto-import-keys` flag to the zypper backend when finding product packages.

Fixes bsc#1231328

## How to test

I have manually tested this code, but I still have to do an end to end test which considers the exact path as described in bsc#1231328.

### DIFF
--- a/cmd/zypper-migration/migration.go
+++ b/cmd/zypper-migration/migration.go
@@ -497,8 +497,8 @@ func compareEditions(left, right string) int {
 	return 0
 }
 
-func cleanupProductRepos(p connect.Product, force bool) error {
-	productPackages, err := zypper.FindProductPackages(p.Name)
+func cleanupProductRepos(p connect.Product, force, autoImportRepoKeys bool) error {
+	productPackages, err := zypper.FindProductPackages(p.Name, autoImportRepoKeys)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


mssola:
[Bug 1231328 - Upgrade from 15SP3 or earlier to 15SP5 or later with Package Hub registered fails when using zypper-migration](https://trello.com/c/bGPMyeDs/3588-bug-1231328-upgrade-from-15sp3-or-earlier-to-15sp5-or-later-with-package-hub-registered-fails-when-using-zypper-migration)

felixsch (approved):
lgtm :+1: 

- Checked the code for potential improvements I would  suggest
- Ran a migration with enabled PackageHub from LTSS 15.3 and see it working

I saw it works fine also without `--auto-import-gpg-key` it will just interactively ask to import the key.
What I'm wondering the Bug sounds like they want it enabled by default and not only when you add `--auto-import-gpg-key` or?

Anyway, works fine for me

cheer,s

mssola:
> What I'm wondering the Bug sounds like they want it enabled by default and not only when you add --auto-import-gpg-key or?

I'm not exactly sure about that. To me it sounds like they are just trying to perform the migration and it failed on both Yast and `zypper migration`. I belive that just giving them a clear answer to use this flag they'd be ok (or at least that's what I understood :D).

PR #272, opened 2024-11-08
Write diagnostics/error messages to stderr

wstephenson:
Like a good POSIX program

https://bugzilla.suse.com/show_bug.cgi?id=1233147 describes a complaint from Amazon that the error output on `--json --status` when the system registration is not found on SCC is not in json format. The underlying problem is that the error messages are all output on stdout.  The user should check the return code of the process before interpreting stdout. 

Nevertheless, it's good practice to put error messages and diagnostics on stdout. 

I'm unsure whether the error output on Register (lines 290ff) and deRegister (233ff) with `--json` *can* be put on stderr, because the code is outputting the result of the API operation as a json object, and consumers may rely on this behaviour.

### DIFF
--- a/cmd/suseconnect/suseconnect.go
+++ b/cmd/suseconnect/suseconnect.go
@@ -132,7 +132,7 @@ func main() {
 	connect.CFG.Load()
 	if baseURL != "" {
 		if err := validateURL(baseURL); err != nil {
-			fmt.Printf("URL \"%s\" not valid: %s\n", baseURL, err)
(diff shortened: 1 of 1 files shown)


felixsch:
I asked Robert and Ignaz if they see problems with this

rjschwei:
This can proceed from my POV, we have https://github.com/SUSE-Enceladus/cloud-regionsrv-client/pull/223 to accommodate this change.

PR #280, opened 2025-01-02
[3/4] public library: Implement product activation and extension traversal

felixsch:
tbd


pkg/registration/activate.go
```diff
@@ -21,16 +25,104 @@ type Metadata struct {
...
+	ObsoletedName string  `json:"obsoleted_service_name"`
+	Product       Product `json:"product"`
+}
+
 // Activate a product by pairing an authorized connection (which contains the
 // system at hand), plus the "triplet" being used to identify the desired
 // product.
-func Activate(conn connection.Connection, identifier, version, arch string) (*Metadata, *Product, error) {
-	return &Metadata{}, &Product{}, nil
+func Activate(conn connection.Connection, identifier, version, arch, regcode string) (*Metadata, *Product, error) {
```
felixsch, line 46: Added `regcode` to support products which require an additional registration code. Did you not added specific tests thoug. Fine?

pkg/registration/product.go
```diff
@@ -1,12 +1,19 @@
...
+)
+
 // Product as defined from SCC'S API.
 type Product struct {
-	Name    string `json:"identifier"`
-	Version string `json:"version"`
-	Arch    string `json:"arch"`
-	Summary string `json:"summary,omitempty"`
-	IsBase  bool   `json:"isbase"`
+	Name       string `json:"name"`
```
felixsch, line 11: I changed `Name` to `name`. I was wondering what did we think about the `LongName`, `FriendlyName` topic?
I added `identifier` to make things more clear.

@mssola do you remember?

pkg/registration/activate.go
mssola: Fine by me!

pkg/registration/product.go
mssola: Identifier should be better. iirc there was a mismatch on how "names" were named on SCC and how they were named here in SUSEConnect. This is a good chance to come up with a clean slate, so let's use names like `Identifier` which sound more to the point.

I'd still include `FriendlyName` as it is used by `zypper-migration`. As for `LongName`, I'm not sure, and I see that we do use it internally. Hence, I'd still keep it but remove it if it doesn't make sense whenever we use this public library from within SUSEConnect.

felixsch: I removed `LongName`. `FriendlyName`, `Name` and `Identifier` exists now as in the API

pkg/registration/product.go
```diff
@@ -22,7 +29,69 @@ type Product struct {
...
+	return nil
+}
+
+type productShowRequest struct {
+	Identifier string `json:"identifier"`
+	Version    string `json:"version"`
+	Arch       string `json:"arch"`
+}
+
+func FetchProductInfo(conn connection.Connection, identifier, version, arch string) (*Product, error) {
```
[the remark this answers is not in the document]
felixsch: done!

mssola (approved):

PR #283, opened 2025-01-27
Add fetching product activations from SCC

felixsch:
card: https://trello.com/c/hnFo5GtQ/3724-rr2-library-add-systems-activations-functionality

This adds fetching `activation` from `/connect/systems/activations` and enables library users to determine which products are already activated on their system, as well as have a minimal input into the subscription consumed.

**How to test:**
```bash
$ cd <connect>
$ git checkout library-support-fetching-activations
$ make build
$ env REGCODE=<you know it!> ./out/public-api-demo SLES 15.6 x86_64
# expect: There is step 6 which shows all activation of the SLES system
```

**As always, if you need help testing this merge request or anything else, do not hesitate to reach out to me!**:rocket:

Thank you!

### DIFF
--- a/pkg/registration/activation.go
+++ b/pkg/registration/activation.go
@@ -1,0 +1,80 @@
+package registration
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/SUSE/connect-ng/pkg/connection"
+)
+
+// Activation holds information about a product activation.
+// subscription information.
+type Activation struct {
+	Name             string    `json:"name"`
+	Status           string    `json:"status"`
+	RegistrationCode string    `json:"regcode"`
+	Type             string    `json:"type"`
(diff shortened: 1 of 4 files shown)


pkg/registration/activation.go
```diff
@@ -0,0 +1,80 @@
...
+	Metadata *Metadata
+	Product  *Product
+}
+
+type activationResponse struct {
+	Activation
+	MetadataAndProduct activateResponse `json:"service"`
+}
+
+var noActivations = []*Activation{}
```
mssola, line 29: Do we really need this? As in, down below, wouldn't  `return []*Activation{}` work as well?

felixsch: no not necessarily. I'm just lazy, should I replace it?

felixsch: done!

PR #284, opened 2025-01-30
Implement a specific error type for API responses

mssola:
## How to test

```
$ make build
$ REGCODE=invalid ./out/public-api-demo SLES 15.6 x86_64
```

Then you will receive a proper message:

```
public-api-demo: A connect client library demo
1) Setup connection and perform an request
<- fetch token
-> update token
API error: Unknown Registration Code. (code: 401)
```

### DIFF
--- a/pkg/connection/connection.go
+++ b/pkg/connection/connection.go
@@ -78,9 +78,9 @@ func (conn ApiConnection) Do(request *http.Request) ([]byte, error) {
 		return nil, err
 	}
 
-	if !successCode(response.StatusCode) {
-		msg := parseError(response.Body)
-		return nil, fmt.Errorf("API error: %v (code: %v)", msg, response.StatusCode)
+	// Check if there was an error from the given API response.
+	if apiError := ErrorFromResponse(response); apiError != nil {
+		return nil, apiError
 	}
 
 	data, readErr := io.ReadAll(response.Body)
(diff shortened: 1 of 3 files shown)


pkg/connection/connection.go
```diff
@@ -78,9 +78,9 @@ func (conn ApiConnection) Do(request *http.Request) ([]byte, error) {
 		return nil, err
 	}
 
-	if !successCode(response.StatusCode) {
-		msg := parseError(response.Body)
-		return nil, fmt.Errorf("API error: %v (code: %v)", msg, response.StatusCode)
+	// Check if there was an error from the given API response.
+	if api_error := ErrorFromResponse(response); api_error != nil {
+		return nil, api_error
```
felixsch, line 83:
```suggestion
	// Check if there was an error from the given API response.
	if apiError := ErrorFromResponse(response); apiError != nil {
		return nil, apiError
```

somebody is in the C syntax world I see :P

felixsch (approved):
lgtm :+1: 

- Checked the changes for things I want to suggest
- Ran the `public-demo-api` command and checked the error returned

**As always, if you think I'm missing something which I should test, please let me know!** :rocket:

cheers,

PS: I'm wondering maybe in the future we should add different types of errors to make it more easy to distinguish between different errors. Is this a thing in golang land?

PR #285, opened 2025-02-19
[RR2] library: Signed offline registration validation support

felixsch:
card: https://trello.com/c/EWTBCHJ4/3780-rr2-library-signed-offline-registration-validation-support

This merge request adds the offline registration certificate handling to the public library. This includes, parsing and validation using the provided RSA public key and the supplied RSA-PSS signature.

**How to review this merge request:**

```bash
$ make build
$ ./out/validate-offline-certificate testdata/pkg/registration/offline_certificate/valid.cert some-scc-regcode invalid-uuid
validate-offline-certificate: Validate a offline registration certificate and print useful information
valid: yes
regcode matches: yes
uuid matches: false
subscription: Rancher PRIME unlimited
expires: 2026-01-27 11:53:51.223 +0000 UTC

$ ./out/validate-offline-certificate testdata/pkg/registration/offline_certificate/valid.cert invalid-regcode 3a4d46b4-0b06-488f-8d20-a931d398d357
validate-offline-certificate: Validate a offline registration certificate and print useful information
valid: yes
regcode matches: yes
uuid matches: true
subscription: Rancher PRIME unlimited
expires: 2026-01-27 11:53:51.223 +0000 UTC

# I encorage you to try generate a certificate on your own from https://github.com/SUSE/happy-customer/pull/8119 and see if it is really working
```

**Note:** Keep in mind that the public key used in SUSEConnect is the `production` public key and NOT the development key. If you test with a self generated certificate, make sure to either import the production key as development or run SCC side in production mode.

Thanks for having a look

cheers,


pkg/registration/offline_certificate.go
```diff
@@ -0,0 +1,108 @@
...
+	Subscription Subscription   `json:"subscription"`
+	RegcodeHash  string         `json:"regcode_sha256"`
+	Information  map[string]any `json:"information"`
+}
+
+// Reads an offline registration certificate from a read object
+//
+// An error indicates the certificate was probably malformed
+func OfflineCertificateFrom(reader io.Reader) (*OfflineCertificate, error) {
+	certificate := OfflineCertificate{}
```
mssola, line 41: Declare this is a pointer `certificate := &OfflineCertificate{}` and then you won't have to de-reference it in `json.Unmarshal` or when returning it.

pkg/registration/offline_certificate.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	verifyErr := rsa.VerifyPSS(key, crypto.SHA256, digest[:], signature, options)
+
+	return verifyErr == nil, nil
+}
+
+// Extracts the payload from an offline registration certificate
+func (cert *OfflineCertificate) ExtractPayload() (*OfflinePayload, error) {
+	payload := OfflinePayload{}
```
mssola, line 90: Same as with the comment on the certificate.

pkg/registration/utilities.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("base64: %s", err)
+	}
+
+	// We need to cut at the end since it is possible to include
+	// gibberish after len(input)
+	return decoded[:size], nil
+}
+
+func sCCPublicKey() (*rsa.PublicKey, error) {
```
mssola, line 30: The naming feels really off 😄  What about `sccPublicKey`?

pkg/registration/subscription.go
```diff
@@ -0,0 +1,14 @@
+package registration
+
+import "time"
+
+// A subscription used in the offline registration workflow
```
mssola, line 5: Is this really specific to offline registration? If so I'd reflect it also on the naming to avoid confusion, just like we do with `OfflineCertificate` or `OfflinePayload`. Yes, it's a bit of a pity to keep adding `Offline*` to things, but I prefer that to avoid silly mistakes in the future.

mssola (changes requested):
Great work!

pkg/registration/utilities.go
felixsch: you are right. I don't know what happened. I probably had it public and then made it private :D

pkg/registration/subscription.go
felixsch: You are right. I moved it now to `SubscriptionInfo` to match the endpoint. If we in the future implement `subscription_info` endpoint in the library it will match the structure and can be reused. Fine for you?

pkg/registration/offline_certificate.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	verifyErr := rsa.VerifyPSS(key, crypto.SHA256, digest[:], signature, options)
+
+	return verifyErr == nil, nil
+}
+
+// Extracts the payload from an offline registration certificate
+func (cert *OfflineCertificate) ExtractPayload() (*OfflinePayload, error) {
+	if cert.OfflinePayload != nil {
```
felixsch, line 94: I added buffering here to not extract the payload multiple times. Case of premature optimization? I'm unsure. What do you think @mssola

mssola: Sounds good to me!

mssola (approved):
LGTM! 🎉

PR #287, opened 2025-03-13
pkg: Remove *_test packages

mssola:
This is against Go's conventions and it makes importing these packages more difficult.

## How to test

Run `make test` and check that everything still passes.


felixsch (approved):
lgtm :+1:

PR #292, opened 2025-05-07
Rip cfg from suseconnect

mssola:


cmd/suseconnect/suseconnect.go
```diff
@@ -356,8 +348,8 @@ func exitOnError(err error, opts *connect.Options) {
 		os.Exit(64)
 	}
 	if je, ok := err.(connect.JSONError); ok {
-		if err := maybeBrokenSMTError(opts); err != nil {
-			fmt.Println(err)
+		if connect.IsOutdatedRegProxy(opts) {
```
felixsch, line 351: This triggers when enabling an extension. I'm not sure why but to reproduce:

```
> ./out/suseconnect -p sle-module-hpc/15.6/x86_64
Registering system to SUSE Customer Center

Activating sle-module-hpc 15.6 x86_64 ...
Your Registration Proxy server doesn't support this function. Please update and try again.
```

vs. on regular connect:

```
Registering system to SUSE Customer Center

Updating system details on https://scc.suse.com ...

Activating sle-module-hpc 15.6 x86_64 ...
Error: Registration server returned 'The product you are attempting to activate (HPC Module 15 SP6 x86_64) requires one of these products to be activated first: Web and Scripting Module 15 SP6 x86_64' (422)

```

I haven't checked the code, do you think this is because we are midway through the migration or is this something to look out for?

felixsch (approved):
lgtm :+1:

Questions and nitpick but nothing serious. 

- Build the branch and ran the following tests:
  * Normal registration works as expected
  * Listing extensions works
  * Checked recommended modules are activated
  * Activating a module works
  * Deactivating a module works
  * Deregistering works
  * `zypper-package-search` works as expected
  * `suse-uptime-tracker` works as expected
  * `zypper-migration` works as expected
  * `yast-registration` with `libsuseconnect` and bindings works:
    * Re-registering works
    * Selecting modules (toplevel and in the middle of the tree) works
    * **BROKEN:** Registering does not work to zypper locking the database. I'm not sure if this is really a connect problem, but it works with released connect.

As for the `libsuseconnect` testing it is easy as:

```
> zypper in -y yast2-registration
> cp out/libsuseconnect /usr/lib64/libsuseconnect.so
> ldconfig
# voila go ahead and do whatever needs to be done

```

**As always, if you think I missed something I should test, please let me know!**

internal/connect/migration.go
```diff
@@ -11,7 +11,7 @@ import (
 type MigrationPath []Product
 
 // Rollback restores system state to before failed migration
-func Rollback() error {
+func Rollback(insecure bool) error {
```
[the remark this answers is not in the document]
mssola: Note that this and the next one in the end delve into the `zypper` territory, which does not pull the `Options` struct (I wanted to keep the `Options` as an `internal/connect` thing). If you want, we can keep this in `Options` and then, whenever we are at the layer of calling the zypper method, pass only `Options.Insecure`, and then the `zypper` package would go on with a signature of `..., insecure bool)`

cmd/suseconnect/suseconnect.go
mssola: I will take another look.

internal/connect/migration.go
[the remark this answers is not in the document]
felixsch: make sense I missed the that this is in migration.go. All good then, keeping the Options in `internal/connect` makes sense.

cmd/suseconnect/suseconnect.go
mssola: Ok, so this is because product activation have not been updated yet. I have added into my todo to check that the detection of SCC vs RMT is really working.

internal/connect/wrapper.go
```diff
@@ -0,0 +1,89 @@
...
+	Connection *connection.ApiConnection
+
+	// Whether the current system is registered or not. Set after calling `New.`
+	Registered bool
+}
+
+// Returns a new Wrapper object by taking the given Options into account. Note
+// that it will also make an attempt to read any available credentials, and set
+// Wrapper.Registered accordingly.
+func New(opts *Options) *Wrapper {
```
felixsch, line 26: Can we rename this to `NewWrapper` to make it more speaking in the rest of the code?

internal/connect/wrapper.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/SUSE/connect-ng/internal/collectors"
+	"github.com/SUSE/connect-ng/internal/credentials"
+	"github.com/SUSE/connect-ng/pkg/connection"
+	"github.com/SUSE/connect-ng/pkg/registration"
+)
+
+// Wrapper is a bridge between API connections via `pkg/connection/` and
+// `internal/credentials/`. Use this wrapper in order to perform any API
+// requests on the context of SUSEConnect.
+type Wrapper struct {
```
felixsch, line 15: What if we make Wrapper an `interface` rather then a `struct`? This would allow us to mock all `pkg/` related calls in SUSEConnect and we can just check the result there.

PR #293, opened 2025-05-12
Add extra data option to register and status

felixsch:
card: https://trello.com/c/obMObgvt/3899-rr4-instance-data-support-is-in-the-public-library

This merge requests, add support for `instanceData` and all the other optional attributes we might want to send along with a registration or status request.

**How to test changes:**

You need to test this:
 * connect code base with branch checked out
 * mitmproxy (HINT: If you use mise `mise use mitmproxy`)
 * A regcode

```
$ docker run --rm --privileged --network host -ti -v $(pwd):/connect registry.suse.com/bci/golang:1.21-openssl
> cd /connect
> make build

# In another console
$ mitmdump --mode reverse:https://scc.suse.com -p 7000 --set flow_detail=3

# Run with a regcode
> SCC_URL="http://localhost:7000" REGCODE="<regcode>" ./out/public-api-demo SLES 15.6 x86_64
# Hit enter until you reach step 5 completed.

# expect: See that the request `PUT https://scc.suse.com/connect/systems` includes `instance_data`

127.0.0.1:41462: PUT https://scc.suse.com/connect/systems
    Host: scc.suse.com
    User-Agent: public-api-demo/1.0
    Content-Length: 117
    Accept: application/json,application/vnd.scc.suse.com.v4+json
    Accept-Language: DE
    Content-Type: application/json
    System-Token: 003ff9ee-6b4d-46a7-bdd7-aa90dc7bad63
    Accept-Encoding: gzip

    {
        "hostname": "public-api-demo",
        "hwinfo": {
            "uname": "public api demo - ping"
        },
        "instance_data": "<document>{}</document>"
    }

```

**As always, do not hesitate to reach out,  if you have questions, feedback or anything else**:rocket:

Thank you! :heart:

### DIFF
--- a/pkg/connection/connection.go
+++ b/pkg/connection/connection.go
@@ -43,12 +43,19 @@ func New(opts Options, creds Credentials) *ApiConnection {
 }
 
 func (conn ApiConnection) BuildRequest(verb string, path string, body any) (*http.Request, error) {
-	bodyData, err := json.Marshal(body)
+	buffer := bytes.Buffer{}
+
+	// Make sure we do not encode HTML data which might interfere with instance_data which can be valid
+	// XML.
+	encoder := json.NewEncoder(&buffer)
+	encoder.SetEscapeHTML(false)
+
+	err := encoder.Encode(body)
 	if err != nil {
 		return nil, err
 	}
 
-	request, err := http.NewRequest(verb, conn.Options.URL, bytes.NewReader(bodyData))
+	request, err := http.NewRequest(verb, conn.Options.URL, bytes.NewReader(buffer.Bytes()))
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/registration/status.go
+++ b/pkg/registration/status.go
@@ -18,16 +18,22 @@ const (
 )
 
 type statusRequest struct {
-	Hostname string            `json:"hostname"`
-	Info     SystemInformation `json:"hwinfo"`
+	Hostname string `json:"hostname"`
+
+	requestWithInformation
 }
 
 // Returns the registration status for the system pointed by the authorized
 // connection.
-func Status(conn connection.Connection, hostname string, info SystemInformation) (StatusCode, error) {
+func Status(conn connection.Connection, hostname string, systemInformation SystemInformation, extraData ExtraData) (StatusCode, error) {
 	payload := statusRequest{
 		Hostname: hostname,
-		Info:     info,
+	}
+
+	enrichWithSystemInformation(&payload.requestWithInformation, systemInformation)
+	enrichErr := enrichWithExtraData(&payload.requestWithInformation, extraData)
+	if enrichErr != nil {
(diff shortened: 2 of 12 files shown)


pkg/connection/connection.go
```diff
@@ -43,12 +43,19 @@ func New(opts Options, creds Credentials) *ApiConnection {
 }
 
 func (conn ApiConnection) BuildRequest(verb string, path string, body any) (*http.Request, error) {
-	bodyData, err := json.Marshal(body)
+	buffer := bytes.Buffer{}
+
+	// Make sure we do not encode HTML data which might interfere with instance_data which can be valid
+	// XML.
+	encoder := json.NewEncoder(&buffer)
+	encoder.SetEscapeHTML(false)
```
felixsch, line 51: It turns out that escaping is enabled by default. This interferes with the `xml/json` document style sent by public cloud clients. So I disabled it, any thoughts how this could backfire?

pkg/registration/system_information.go
```diff
@@ -1,5 +1,61 @@
...
+//   - online_at ([]string): OnlineAt definition which records hourly usage time of the system (Cloud)
+type ExtraData = map[string]any
+
+// Use NoExtraData when no extra information is used by the registration or status method.
+// Example:
+//
+//	registration.Register(conn, "regcode", "hostname", NoSystemInformation, NoExtraData)
+var NoExtraData = map[string]any{}
+
+type requestWithInformation struct {
```
felixsch, line 32: I created a generic structure here to reuse it both on `register` and `status/keepalive`. So we can just plug this in whenever we need to transport the `system_information` and `extra_data` combo.

pkg/registration/status.go
```diff
@@ -18,16 +18,22 @@ const (
...
-	Info     SystemInformation `json:"hwinfo"`
+	Hostname string `json:"hostname"`
+
+	requestWithInformation
 }
 
 // Returns the registration status for the system pointed by the authorized
 // connection.
-func Status(conn connection.Connection, hostname string, info SystemInformation) (StatusCode, error) {
+func Status(conn connection.Connection, hostname string, systemInformation SystemInformation, extraData ExtraData) (StatusCode, error) {
```
felixsch, line 28: I moved to our own definition for `SystemInformation` and `ExtraData`. Both are type alias to the `map[string]any` construct.

Btw. my linter complains about writing `map[string]interface{}` vs `map[string]any`, so I switched to it

pkg/connection/connection.go
mssola: Not that I can think of right now...

mssola (approved):
LGTM! 🎉 

I have tested the code and works as promised. Once we have this one and the other PR merged, we should make an extra effort to clean up APIs and such, just to ensure we have a way so more parameters don't creep up function arguments. But what I'm talking here is out of the scope from this PR, so let's go ahead.

PR #298, opened 2025-05-26
Wrapped library API: To make testing in the future way more easy

felixsch:
part of: https://trello.com/c/QxBSnLYB/3864-rr4-suseconnect-move-away-from-the-internal-package-for-api-related-things

There is not really a way to test this besides implementing something which uses `WrappedAPI` and uses tests which utilize the implementation.

**As always, if you think I missed something or something is unclear, please do not hesitate to reach out to me!**:rocket:

Thank you, :heart:

Felix


internal/connect/wrapper.go
```diff
@@ -23,6 +39,7 @@ type Wrapper struct {
 // Returns a new Wrapper object by taking the given Options into account. Note
 // that it will also make an attempt to read any available credentials, and set
 // Wrapper.Registered accordingly.
+// TODO(felixsch): When we are done, this should return WrappedAPI instead of Wrapper directly
```
mssola, line 42: Why can't we just return it right away?

mssola (approved):

PR #300, opened 2025-05-27
Feature tests in golang

felixsch:
part of: https://trello.com/c/QxBSnLYB/3864-rr4-suseconnect-move-away-from-the-internal-package-for-api-related-things

This merge request replaces the old ruby feature tests with more up to date golang feature tests.
I tried to migrate all existing tests and enhance them where possible.

Missing for now:
- translation based tests

This of course is only the first step, and since this is against `next` you will notice the tests failing. This already shows a few regressions added while migrating to the library.

To see tests passing run the feature tests against the release 1.13 release:

```
$ cd <connect>
$ cp .env-example .env
# Fill the values (ask me if needed)

$ docker run --rm --privileged --env-file .env --network host -ti -v $(pwd):/connect registry.suse.com/bci/golang:1.21-openssl
> cd /connect
> zypper in -y SUSEConnect
# get the 1.13 release
> go test -v features/suseconnect/*
# expect: This passes
```

**WIP until CI is passing and we resolved existing regressions!**

Thank you very much for having a look :heart:


felixsch:
Note that I have to fix CI secrets. Will do tomorrow!

.env-example
```diff
@@ -0,0 +1,5 @@
+# Needed to run feature tests
+# NOTE: Check the suseconnect - CI Account organization for registration codes
```
mssola, line 2: I'd remove this note.

features/helpers/zypper.go
```diff
@@ -0,0 +1,84 @@
...
+		t: t,
+	}
+}
+
+func (zypp *Zypper) run(params ...string) []byte {
+	cmd := exec.Command("zypper", params...)
+	result, err := cmd.CombinedOutput()
+
+	if err != nil {
+		assert.FailNow(zypp.t, "Running zypper failed", err)
```
mssola, line 28: Could we print the output on failure? Maybe it will help debugging assert failures.

features/helpers/zypper.go
```diff
@@ -0,0 +1,84 @@
...
+	return &Zypper{
+		t: t,
+	}
+}
+
+func (zypp *Zypper) run(params ...string) []byte {
+	cmd := exec.Command("zypper", params...)
+	result, err := cmd.CombinedOutput()
+
+	if err != nil {
```
mssola, line 27: You may also want to cast this to `exec.ExitError` if it's non-nil. This way you can retrieve the exit code, which is relevant because zypper has special exit codes which might further help debugging errors (see the constants on `internal/zypper.zypper.go` named `zypperErr*` and `zypperInfo*`).

.env-example
felixsch: done

PR #306, opened 2025-06-06
Migrate product removal/upgrade/activation/etc. to the public API.

mssola:
**TODO**: I still have to rebase with the latest changes pushed into `next` (e.g. NewWrapperAPI)


pkg/search/package.go
```diff
@@ -0,0 +1,49 @@
...
+	Name     string                 `json:"name"`
+	Arch     string                 `json:"arch"`
+	Version  string                 `json:"version"`
+	Release  string                 `json:"release"`
+	Products []registration.Product `json:"products"`
+}
+
+// Performs a package search request for the given search `query`. It also
+// expects the triplet identifier for the product to be used as a `base`.
+func Package(conn connection.Connection, query, base string) ([]SearchPackageResult, error) {
```
felixsch, line 24: Can you add unit tests here? I think we should have everything tested in the library.

pkg/registration/product.go
```diff
@@ -102,3 +171,56 @@ func FetchProductInfo(conn connection.Connection, identifier, version, arch stri
 
 	return &product, nil
 }
+
+// Updates the given product (i.e. upgrades/downgrades it) depending on what the
+// SCC server determines. An SCC service is given in return.
+func UpdateProduct(conn connection.Connection, product Product) (Service, error) {
```
felixsch, line 177: Can you add tests here? Within the library I think everything should be tested to be in a good spot

pkg/registration/product.go
```diff
@@ -61,6 +104,32 @@ func (pro *Product) TraverseExtensions(fn TraverseFunc) error {
 	return nil
 }
 
+// Returns the extension for this product which matches the given triplet
+// identifier.
+func (p Product) FindExtension(triplet string) (Product, error) {
```
felixsch, line 109: Can you add tests for this method to make sure it is working correctly?

pkg/registration/activate.go
```diff
@@ -43,7 +43,7 @@ type activateResponse struct {
 // Activate a product by pairing an authorized connection (which contains the
 // system at hand), plus the "triplet" being used to identify the desired
 // product.
-func Activate(conn connection.Connection, identifier, version, arch, regcode string) (*Metadata, *Product, error) {
+func Activate(conn connection.Connection, identifier, version, arch, regcode string) (*Metadata, Product, error) {
```
felixsch, line 46: Why removing the point to product here? Is there a specific case why this would be desirable? I'm wondering why keeping `*Metadata` as pointer but not `Product`?

internal/connect/system.go
```diff
@@ -43,9 +43,5 @@ func Cleanup(baseURL, basePath string) error {
 func UpdateCertificates() error {
 	cmd := []string{"/usr/sbin/update-ca-certificates"}
 	_, err := util.Execute(cmd, []int{0})
-	if err != nil {
-		return err
-	}
-	// reload CA certs in Go
-	return ReloadCertPool()
+	return err
```
felixsch, line 46: I'm wondering: Do we not need the certificate reloading anymore?

pkg/registration/product.go
felixsch: Also why not call it `Upgrade` rather than `Update`. I think this would be more speaking of what the method does?

pkg/registration/product.go
```diff
@@ -102,3 +171,56 @@ func FetchProductInfo(conn connection.Connection, identifier, version, arch stri
...
+
+// Updates the given product (i.e. upgrades/downgrades it) depending on what the
+// SCC server determines. An SCC service is given in return.
+func UpdateProduct(conn connection.Connection, product Product) (Service, error) {
+	return touchProduct(conn, product, "PUT")
+}
+
+// Removes the given product as part of the system (i.e. deactivates it). An SCC
+// service is given in return.
+func RemoveProduct(conn connection.Connection, product Product) (Service, error) {
```
felixsch, line 183: I would call it `DeactivateProduct` rather then `RemoveProduct`

pkg/registration/product.go
```diff
@@ -2,46 +2,89 @@ package registration
...
-	// optional extension products
+	// Optional extension products
 	Extensions []Product `json:"extensions,omitempty"`
 
-	Description  string       `json:"description,omitempty"`
+	Description  string       `xml:"description" json:"description,omitempty"`
 	EULAURL      string       `json:"eula_url,omitempty"`
 	FormerName   string       `json:"former_identifier,omitempty"`
 	ProductType  string       `json:"product_type,omitempty"`
+	ProductLine  string       `xml:"productline,attr"`
```
felixsch, line 39: Do you have a clue why we need this? I have not seen any application of this information but probably I just missed it

felixsch (commented):
In general looks good, I ran initial manual tests. I added a couple of comments and I think we should make the feature tests green here to be sure nothing broke on CLI side.

pkg/registration/product.go
mssola: So, this is part of the Product and ZypperProduct merge. No idea why it's needed, but since it comes from zypper and the old ZypperProduct used it (see further that it comes by parsing XML), then I thought to leave it.

felixsch: fair enough, was just curious if you know why this is there. I don't.. Lets keep it then.

felixsch (approved):

PR #310, opened 2025-06-16
[RR6] Add custom certificate support to the public library

felixsch:
card: https://trello.com/c/tOsAXMnx/3938-rr6-connect-library-supports-custom-certificates-to-connect-to-pubcloud-rmt-servers

**Description:**

Adds support to the public library for the upcoming PAYG integration which requires support of self signed certificates.

- Fixes wrong insecure settings by default
- Adds support to the `public-api-demo` application

**How to review this merge request:**

```
$ docker run --rm --privileged --network host -ti -v $(pwd):/connect registry.suse.com/bci/golang:1.21-openssl
> make build

> NON_INTERACTIVE=true REGCODE="<regcode>" ./out/public-api-demo SLES 15.6 x86_64
# expect: works without an error

# Now lets switch to RMT:
# Note: Make sure REGCODE is NOT set or empty string!
> REGISTRATION_HOST_URL="https://rmt.scc.suse.de" ./out/public-api-demo SLES 15.6 x86_64
# expect: This fails with a certificate error: certificate signed by unknown authority

# Fetch the certificate
> curl http://rmt.scc.suse.de/certs/rmt.crt > rmt.crt
> cat rmt.crt
# expect to see a valid certificate

# Now try to register again
> REGISTRATION_HOST_URL="https://rmt.scc.suse.de" API_CERT=rmt.crt ./out/public-api-demo SLES 15.6 x86_64
# expect: This now registers against RMT
# also expect: This will fail on activating certain modules since they are not mirrored on the reference host

```

**As always, if you think I missed something or having trouble with the review instructions. Please do not hesitate to reach out to me!**:rocket:

Thank you :heart:


cmd/public-api-demo/main.go
```diff
@@ -43,6 +45,26 @@ func runDemo(identifier, version, arch, regcode string) error {
...
 	}
 
+	if certificatePath := os.Getenv("API_CERT"); certificatePath != "" {
+		crt, certReadErr := os.ReadFile(certificatePath)
+		if certReadErr != nil {
+			return certReadErr
+		}
+
+		block, _ := pem.Decode(crt)
+		if block == nil {
```
josegomezr, line 55: shouldn't `err`  be part of the check?

felixsch: this is not a `err` its the `rest` of the decoded pem. I was first testing against this error... https://pkg.go.dev/encoding/pem#Decode

digitaltom (approved):
LGTM! One pending question.

PR #312, opened 2025-06-20
fix: Add non-base64 OfflineCertFrom

mallardduck:
Per title, this helps improve DX when working with k8s. The k8s apis will have already decoded the base64 data, so the current method requires that I re-encode it to properly call `OfflineCertificateFrom`. So `OfflineCertificateFromBytes` does the same things just with raw bytes.


pkg/registration/offline_certificate.go
```diff
@@ -41,6 +41,21 @@ type OfflinePayload struct {
 // Reads an offline registration certificate from a read object
 //
 // An error indicates the certificate was probably malformed
+func OfflineCertificateFromBytes(reader io.Reader) (*OfflineCertificate, error) {
```
mssola, line 44: Instead of creating this function, I'd just modify the other with a boolean flag on whether we expect base64 encoding or not. Otherwise, we'd get two functions which are almost identical.

mssola (changes requested):

mssola:
I  have ruled over feature tests since this was a fork with missing credentials.

PR #314, opened 2025-06-25
Remove link to extensions documentation

lagartoflojo:
Currently, the `--list-extensions` command prints out a message linking to the SLE documentation, where modules and extensions are described.

For this link to really make sense, it should point to the documentation page that corresponds to the underlying SLES version. The modules/extensions of 15 SP6 are not necessarily the same as 15 SP7 (and are definitely not the same as 16). Having the link always point to the “latest” version of this document, no matter which version of SLES the server is running, means that when the user clicks on it, they may or may not get accurate information that applies to their system.

It’s also worth mentioning: until recently, the link was actually broken, as it took the user to the documentation home page rather than to the intended page. No one reported this issue, which suggests that our customers are not actually clicking on it.

Given that making this a properly versioned link is more effort, and that the link itself brings little value, we are removing it from the output in an upcoming SUSEConnect release.

### DIFF
--- a/build/packaging/suseconnect-ng.changes
+++ b/build/packaging/suseconnect-ng.changes
@@ -7,6 +7,7 @@
    only /usr/sap but no installation exists (bsc#1241002)
  - Fixed modules and extension link to point to version less documentation. (bsc#1239439)
  - Fixed SAP instance detection (bsc#1244550)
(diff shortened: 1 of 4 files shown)


mssola (approved):

mssola:
Thanks!

PR #315, opened 2025-07-07
Add missing tar dependency for zypper migration

felixsch:
Noticed that `zypper migration` requires `tar` as dependency which is missing in our spec file. I added it.

- Check CI and building rpm is fine. After that you are done!

cheers,

Felix


felixsch:
I close this here in favor of pentially reworking the spec file to have the addons and its dependencies in a seperate subpackage to not further growth the main package.

PR #320, opened 2025-07-17
Add collector to get pci device data

brett060102:
Add collection of PCI data to the heartbeat

This done by adding a collector that will invoke
lspci -vmmDk  -s .0

Requires adding pciutils to required package for suseconnect

The size of the additional payload in the heartbeat depends on how many pci devices are on the system.
Each PCI device on the system generated ~290 bytes of information.

Testing on large AWS systems and couple of other systems I have access to gave the following
approximate payload sizes.

AWS g6.48xlarge          8.2KB
AWS i7ie.metal-24xl    14.0KB
HP Z840                          1.9KB
HP DL360                        3.5KB

### DIFF
--- a/build/packaging/suseconnect-ng.spec
+++ b/build/packaging/suseconnect-ng.spec
@@ -56,6 +56,7 @@ Requires:       ca-certificates
 %endif
 
 Requires:       coreutils
+Requires:       pciutils
 Requires:       zypper
 Requires:       util-linux
 Recommends:     systemd
--- a/internal/collectors/pci.go
+++ b/internal/collectors/pci.go
@@ -1,0 +1,19 @@
+package collectors
+
+import (
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+type PCI struct{}
+
+func (pci PCI) run(arch string) (Result, error) {
+        output, err := util.Execute([]string{"lspci", "-vmmDk", "-s", ".0"}, nil)
+	// Some systems may not have lspci command, so
+	// if the execute fails, 
+	// send nil to SCC to indicate that PCI data was unavailable.
+	if err != nil {
+		return Result{"PCIData": nil}, nil
+	}
+	
+	return Result{"PCIData": string(output)}, nil
+}
--- a/internal/collectors/pci_test.go
+++ b/internal/collectors/pci_test.go
@@ -1,0 +1,34 @@
+package collectors
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+
+func TestPCIRun(t *testing.T) {
+	assert := assert.New(t)
+
+        expectedPCI := `
+Slot:   0000:05:00.0
+Class:  VGA compatible controller
+Vendor: Advanced Micro Devices, Inc. [AMD/ATI]
+Device: Rembrandt [Radeon 680M]
+SVendor:        Advanced Micro Devices, Inc. [AMD/ATI]
+SDevice:        Rembrandt [Radeon 680M]
+Rev:    c7
+ProgIf: 00
+Driver: amdgpu
+Module: amdgpu
+IOMMUGroup:     17
+`
+        mockUtilExecute(expectedPCI, nil)
+	expected := Result{"PCIData": expectedPCI}
+
+	collector := PCI{}
+	result, err := collector.run(ARCHITECTURE_X86_64)
+        
+	assert.Equal(expected, result)
+	assert.Nil(err)
+}
--- a/internal/connect/api.go
+++ b/internal/connect/api.go
@@ -271,6 +271,7 @@ var mandatoryCollectors = []collectors.Collector{
 	collectors.CloudProvider{},
 	collectors.Architecture{},
 	collectors.ContainerRuntime{},
+	collectors.PCI{},
 
 	// Optional collectors
 	collectors.Uname{},


internal/collectors/lsmod.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+type KernMod struct {
+    KernModProfileID   string  // sha256 of KernModProfile
+    KernModProfile  []string     // sorted unique list of modules from lsmod
+}
+type LSMOD struct{}
+const KernModChecksumFile = "/etc/zypp/chksum-kernmoddata.txt"
```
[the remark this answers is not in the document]
mssola: > The lsmod output for server can easily exceed 8Kb and the sorted and uniqued list of modules can exceed 2KB. While this is not excessive in the context of a single system, attaching what would be duplicated data over millions of systems was considered prohibitive. The issue is actually worse with respect to PCI data.

Just to get more context, when you are hashing this data, how much does it take?

> I am not sure I understand comment 1, and that is most likely because I am still learning go."sortedMods" as currently coded is an array of strings, I want a 256 hash the entire array of strings, for my reading this required building a single string up from each array element. Am I missing something?

I'm saying that, whatever you end up sending, you don't need to:

1. Write it into a file.
2. Hash the contents of said file

Instead, whatever you are storing into the file you can simply hash it and that's it. The main rationale for this is that I'm very wary of touching anything from the user's file system, even if it's the cache something.

> The reason for storing the previous sha256 is to avoid re-sending a profile that was already sent. While the re-sending might not be an issue for these cases, the information must still be saved my the receiver until it can be processed and we have other system information being requested where the payloads are in the hundreds of KB and we really don't want to send that twice if we don't have to.

This would apply to any collector, but we don't do it because that's not the job of SUSEConnect. The backend (scc.suse.com) is the source of truth, so it's going to be on that side to decide whether it's needed or not.

In the end, this relates to my first question on how big the hashed info is for that system you are testing with, as I feel I'm missing more context 😄

build/packaging/suseconnect-ng.spec
```diff
@@ -56,6 +56,7 @@ Requires:       ca-certificates
 %endif
 
 Requires:       coreutils
+Requires:       pciutils
```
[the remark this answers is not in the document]
brett060102: I was and that covered in the https://confluence.suse.com/display/publiccloud/Design+-+PCI+Data+Collection The short form was using ghw still required installing the hwdata package in which is pretty big and using /proc/bus/pci/devices pushed the translation of PCI ID off to the analytics and people though that duplicating what lspci does was overly complicated.

internal/collectors/lsmod.go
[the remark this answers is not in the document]
brett060102: The code only store the ID on disk. not the full data.
`
$ cat /etc/zypp/chksum-kernmoddata.txt 
e59f168f9283fa3b5d434856665eab1c65231e590ae5d783c5f4eca0f386987e
`

mssola:
Ok, then I will not review this further. Ping me again when you need another review.

internal/collectors/lsmod.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return Result{lsmodTag: nil}, err
+	}
+	scanner := bufio.NewScanner(bytes.NewReader(modInfo))
+
+	// Iterate through each line get module name
+	modList := make(map[string]bool)
+	for scanner.Scan() {
+		module := strings.Fields(scanner.Text())[0]
+		// skip "Mudule" from the hearder line
```
mssola, line 35: Here and in other places, you have typos in the word "module".

internal/collectors/lsmod.go
```diff
@@ -0,0 +1,83 @@
...
+		if module != "Module" {
```
mssola, line 36: Instead of this, call scanner.Scan before the for loop so the first line is skipped (which is where the header appears).

internal/collectors/pci.go
```diff
@@ -0,0 +1,52 @@
...
+		return Result{pciTag: nil}, err
+	}
+	stringOutput := string(output)
+	// get sha256 has a pci data
+	mkSha256 := sha256.New()
+	mkSha256.Write([]byte(output))
+	sha256 := mkSha256.Sum(nil)
+	hwIndex := hex.EncodeToString(sha256)
+
+	oldHwIndex := util.GetChecksum(PCIChecksumFile)
```
mssola, line 34: I feel like all this code is just a duplicate to that of the lsmod one. Could we have a common function that deals with this?

internal/collectors/pci.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"crypto/sha256"
+	"encoding/hex"
+
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+// var InfoOnly bool // indicated that -i/--info was used
+const PCIChecksumFile = "chksum-pcidata.txt"
+const pciTag = "pci_data"
```
mssola, line 12: idem

internal/connect/collectors.go
```diff
@@ -4,24 +4,38 @@ import (
...
-// Fetch system information based on the available collectors for the system.
-func FetchSystemInformation() (collectors.Result, error) {
+// Fetch system profile information
+func FetchSystemProfiles(infoOnly bool) (collectors.Result, error) {
+	var usedCollectors = []collectors.Collector{
+		collectors.PCI{UpdateDataIDs: !infoOnly},
+		collectors.LSMOD{UpdateDataIDs: !infoOnly},
+	}
+
 	arch, err := collectors.DetectArchitecture()
```
mssola, line 39: Note that this was already called by FetchSystemInformation. Hence, why not add a "arch" flag to this function's signature?

internal/connect/collectors.go
```diff
@@ -4,24 +4,38 @@ import (
...
-	collectors.Virtualization{},
-	collectors.CloudProvider{},
-	collectors.Architecture{},
-	collectors.ContainerRuntime{},
-
-	// Optional collectors
-	collectors.Uname{},
-	collectors.SAP{},
+// Fetch system information based on the available collectors for the system.
+func FetchSystemInformation(infoOnly bool) (collectors.Result, error) {
```
mssola, line 8: Is `infoOnly` ever used? Same for the other function.

internal/collectors/lsmod.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+type LSMOD struct {
+	UpdateDataIDs bool
+}
+
+const KernModChecksumFile = "chksum-kernmoddata.txt"
+const lsmodTag = "mod_list"
```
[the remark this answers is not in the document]
brett060102: need is a strong word, but it and pciTag get used in multiple places in the code. I had those hard coded, but Fergal and I went back and forth on the name a few times, so I added the const to make things easier. I could use a var setting at the top of the routine instead, but it really is a const.

internal/collectors/lsmod.go
brett060102: great idea, thanks

internal/connect/collectors.go
brett060102: no longer in FetchSystemInformation, it was when profile data was being passwd as hwinfo.

internal/collectors/pci.go
brett060102: I will look at that.

cmd/suseconnect/suseconnect.go
```diff
@@ -276,6 +280,9 @@ func main() {
 		if jsonFlag {
 			exitOnError(errors.New("cannot use the json option with the 'cleanup' command"), api, opts)
 		}
+		// need to clear ProfileCache on cleanup
```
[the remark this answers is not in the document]
brett060102: Think I got all of them.

internal/util/profiles.go
```diff
@@ -0,0 +1,159 @@
...
+// conform with the return of util.Execute
+// it scans though the input, removes duplicated and items and sorts
+// the result. The intended use is to process things like lsmod data or rpm -qa
+// data to an uniform view of the data.
+func SortAndUnique(inputStream []byte) ([]string, error) {
+	scanner := bufio.NewScanner(bytes.NewReader(inputStream))
+
+	// Iterate through each line get module name
+	itemList := make(map[string]bool)
+	// Skip header line
```
rtamalin, line 139: This seems implementation case specific, i.e. skipping a header row, and might be better managed via parameters, e.g. supporting both head and tail skip counts.

internal/util/profiles.go
```diff
@@ -0,0 +1,159 @@
...
+// SortAndUnique takes a byte array as input. This is to
+// conform with the return of util.Execute
+// it scans though the input, removes duplicated and items and sorts
+// the result. The intended use is to process things like lsmod data or rpm -qa
+// data to an uniform view of the data.
+func SortAndUnique(inputStream []byte) ([]string, error) {
+	scanner := bufio.NewScanner(bytes.NewReader(inputStream))
+
+	// Iterate through each line get module name
+	itemList := make(map[string]bool)
```
rtamalin, line 138: This is technically a misnomer since it is a map not a list ;-)

internal/util/profiles.go
```diff
@@ -0,0 +1,159 @@
...
+		item := strings.Fields(scanner.Text())[0]
+		itemList[item] = true // update item list
+	}
+	// Return nil and erroe if errors in scanning
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	// Run through module list removing duplicates
+	var sortedList []string
```
rtamalin, line 151: If we ever move forward to go 1.23, or wanted to pull in the experimental package released with go 1.18 https://pkg.go.dev/golang.org/x/exp/maps, this slice declaration and loop can be replaced with

sortedList := maps.Keys()

internal/util/profiles.go
```diff
@@ -0,0 +1,159 @@
...
+// data to an uniform view of the data.
+func SortAndUnique(inputStream []byte) ([]string, error) {
+	scanner := bufio.NewScanner(bytes.NewReader(inputStream))
+
+	// Iterate through each line get module name
+	itemList := make(map[string]bool)
+	// Skip header line
+	scanner.Scan()
+	for scanner.Scan() {
+		item := strings.Fields(scanner.Text())[0]
```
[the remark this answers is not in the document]
brett060102: makes sense, but should wait until we have a better idea about what other collectors will need.

internal/util/profiles.go
brett060102: good to know.

internal/util/profiles.go
brett060102: renamed to itemMap

cmd/suseconnect/suseconnect.go
```diff
@@ -345,6 +354,7 @@ func main() {
 }
 
 func exitOnError(err error, api connect.WrappedAPI, opts *connect.Options) {
+	util.Debug.Println("exitOnError err: ", err)
```
mssola: Isn't this handled on each if block so it's more specific?

internal/connect/wrapper.go
```diff
@@ -148,9 +161,21 @@ func (w Wrapper) Register(regcode, instanceDataFile string) error {
...
+	// Provide profile data for registration
+	// make sure ProfileCache empty for registration
+	util.DeleteProfileCache()
+	profileData, err := FetchSystemProfiles(arch, false)
+	extraData["data_profiles"] = profileData
+
 	// NOTE: we are not interested in the code. Hence, we don't save it
 	// anywhere.
+	util.Debug.Print("hwinfo: ", hwinfo)
+	util.Debug.Printf("extraData: %+v", extraData)
```
mssola: I'd rather not have these debug prints.

internal/util/profiles.go
```diff
@@ -0,0 +1,124 @@
...
+// ProfileData: can be any data such array of strings, a single large string
+type Profile struct {
+	ProfileID   string `json:"profileId"`   // Sha256 of ProfileData
+	ProfileData any    `json:"profileData"` // Data associated with ProfileI
+}
+
+func UpdateCache(updateCache bool, tag string, cacheFile string, dataBlock any) (Result, error) {
+
+	var hashInput = ""
+	switch v := dataBlock.(type) {
```
mssola: As detailed before, let it be a string and avoid type casting shenanigans.

mssola (changes requested):

internal/util/profiles.go
```diff
@@ -0,0 +1,124 @@
...
+//	of strings, etc) it may need processing before the
+//	sha256 can be calcuated
+//
+// ProfileData: can be any data such array of strings, a single large string
+type Profile struct {
+	ProfileID   string `json:"profileId"`   // Sha256 of ProfileData
+	ProfileData any    `json:"profileData"` // Data associated with ProfileI
+}
+
+func UpdateCache(updateCache bool, tag string, cacheFile string, dataBlock any) (Result, error) {
```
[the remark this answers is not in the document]
brett060102: Will rename function to  BuildProfile since primary purpose is to build the profile record to be sent to RMT/SCC.

internal/collectors/lsmod.go
```diff
@@ -0,0 +1,60 @@
...
+	// Return nil and erroe if errors in scanning
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	// Run through module list removing duplicates
+	var sortedList []string
+	for module := range itemMap {
+		sortedList = append(sortedList, module)
+	}
```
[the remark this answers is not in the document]
brett060102: great and thank you, much better.

internal/collectors/lsmod.go
```diff
@@ -0,0 +1,60 @@
...
+	// Sort the list,
+	sort.Strings(sortedList)
+	return sortedList, nil
+}
+
+func (lsmod LSMOD) run(arch string) (Result, error) {
+	util.Debug.Print("lsmod.UpdateDataIDs: ", lsmod.UpdateDataIDs)
+	modInfo, err := util.Execute([]string{"lsmod"}, nil)
+	// Should not happen, but if command fails
+	// send nil Result if module data was unavailable.
```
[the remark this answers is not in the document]
brett060102: OK

internal/connect/wrapper.go
```diff
@@ -122,15 +123,27 @@ func (w Wrapper) KeepAlive(uptimeTracking bool) error {
 		extraData["online_at"] = data
 	}
 
-	code, err := registration.Status(w.Connection, hostname, hwinfo, extraData)
-	if code != registration.Registered {
+	profileData, err := FetchSystemProfiles(arch, false)
+	util.Debug.Print("hwinfo: ", hwinfo)
+	util.Debug.Printf("extraData: %+v", extraData)
+	util.Debug.Printf("profileData: %+v", profileData)
```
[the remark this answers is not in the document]
brett060102: I think so. -i does not include extraData and only included hwdata until I added profileData. Also, Those are different paths through the code.

internal/connect/wrapper.go
brett060102: I added this because this data was not being provided when --debug was used, and it was the only way to validate what data gets sent on registration.

internal/connect/wrapper.go
```diff
@@ -148,9 +161,21 @@ func (w Wrapper) Register(regcode, instanceDataFile string) error {
...
+	util.DeleteProfileCache()
+	profileData, err := FetchSystemProfiles(arch, false)
+	extraData["data_profiles"] = profileData
+
 	// NOTE: we are not interested in the code. Hence, we don't save it
 	// anywhere.
+	util.Debug.Print("hwinfo: ", hwinfo)
+	util.Debug.Printf("extraData: %+v", extraData)
 	_, err = registration.Register(w.Connection, regcode, hostname, hwinfo, extraData)
+	// If registraion failed, delete profile cache
```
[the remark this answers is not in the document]
brett060102: thanks

internal/util/profiles.go
```diff
@@ -0,0 +1,124 @@
...
+	mkSha256 := sha256.New()
+	mkSha256.Write([]byte(input))
+	sha256 := mkSha256.Sum(nil)
+	return hex.EncodeToString(sha256)
+}
+
+func SetProfileFilePath(newPath string) {
+	profileFilePath = newPath
+	if profileFilePath[len(profileFilePath)-1] != '/' {
+		profileFilePath = profileFilePath + "/"
```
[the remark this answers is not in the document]
brett060102: Thanks

pkg/profiles/profiles.go
```diff
@@ -0,0 +1,135 @@
...
+var profileFilePath = "/run/suseconnect/"
+
+// Profile struct is for handling lage data blobs
+// ID: an unique index to reference Profile data
+//
+//	Profile data can be any type (string, array
+//
+// Data: Can be any data such array of strings, a single large string,a map, etc.
+type Profile struct {
+	ID   string `json:"profileId"`
```
mssola: Since profile tags follow a snake case format, could we maybe do the same here? So `profile_id` and `profile_data` for the JSON export.

pkg/profiles/profiles.go
```diff
@@ -0,0 +1,135 @@
...
+	err = os.WriteFile(chkSumFilePath, []byte(value), perms)
+	return err
+}
+
+func DeleteProfileCache() {
+	// Remove profileFilePath and all files, ignore error
+	_ = os.RemoveAll(profileFilePath)
+}
+
+func CalcSha256(input string) string {
```
mssola: Do we really need to export this function?

pkg/profiles/profiles_test.go
```diff
@@ -0,0 +1,66 @@
+package profiles
+
+import (
+	//"fmt"
```
mssola: Remove 😄

mssola (changes requested):
Great work! We are getting closer 😄

internal/util/profiles.go
mssola: Just saying all of this because I don't think we need to use `any` unless there's an extremely good argument on using it.

pkg/collectors/lsmod.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/SUSE/connect-ng/internal/util"
+	"github.com/SUSE/connect-ng/pkg/profiles"
+)
+
+type LSMOD struct {
+	UpdateDataIDs bool
+}
+
+const kernModChecksumFile = "kern-modules.txt"
```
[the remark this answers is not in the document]
brett060102: OK

pkg/collectors/lsmod.go
```diff
@@ -0,0 +1,55 @@
...
+const lsmodTag = "mod_list"
+
+// process output from lsmod to remove dups and get sorted list of kernel mods.
```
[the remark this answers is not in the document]
brett060102: thanks

pkg/profiles/profiles.go
```diff
@@ -0,0 +1,135 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+type Result = map[string]interface{}
+
+var profileFilePath = "/run/suseconnect/"
```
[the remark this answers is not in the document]
brett060102: Will document, but changing that var is done via func SetProfileFilePath(newPath string)

pkg/profiles/profiles.go
```diff
@@ -0,0 +1,135 @@
...
+// Returns:
+//        Result record: Result = map[string]interface{}
+//	  Error
+// Given a data object of any type, the function contructs am unique ID for that data and builds
+// a result record of the form: tag: "id":["id_value", "data": dataBlock].
+// If the calculated ID maches local cached ID in the cacheFile, then the dataBlock is ommitted
+// in the return.
+// If updateCache is true and the alculated ID does not match the local cached ID in cacheFile,
+// cacheFile is updated with the calculated ID value
+
```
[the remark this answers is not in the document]
brett060102: thanks

internal/collectors/pci.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func (pci PCI) run(arch string) (Result, error) {
+	util.Debug.Print("pci.UpdateDataIDs: ", pci.UpdateDataIDs)
+	output, err := util.Execute([]string{"lspci", "-s", ".0"}, nil)
+	if err != nil {
+		return Result{}, err
+	}
+	pciData, _ := getPciData(output)
+	result, _ := profiles.BuildProfile(pci.UpdateDataIDs, pciTag, pciChecksumFile, pciData)
```
paragjain0910: The error from getPciData and  BuildProfile is ignored.  I would suggest to handle it properly ( log it) or if we want to propagate it further to parent functions. The child functions are throwing the errors though.

internal/connect/collectors.go
```diff
@@ -2,30 +2,56 @@ package connect
...
 }
+
+// Fetch system profile information
+func FetchSystemProfiles(arch string, updateCache bool) (collectors.Result, error) {
+	var usedCollectors = []collectors.Collector{
+		collectors.PCI{UpdateDataIDs: updateCache},
+		collectors.LSMOD{UpdateDataIDs: updateCache},
+	}
+
+	var err error = nil
```
paragjain0910:
```suggestion
	var err error
```

internal/connect/collectors.go
```diff
@@ -2,30 +2,56 @@ package connect
...
+		collectors.CloudProvider{},
+		collectors.Architecture{},
+		collectors.ContainerRuntime{},
+
+		// Optional collectors
+		collectors.Uname{},
+		collectors.SAP{},
+	}
+
+	var err error = nil
```
paragjain0910:
```suggestion
	var err error 
```

internal/connect/collectors.go
```diff
@@ -2,30 +2,56 @@ package connect
...
+	}
+
+	if err != nil {
+		return collectors.NoResult, err
+	}
+	profile, err := collectors.CollectInformation(arch, usedCollectors)
+	if err != nil {
+		profiles.DeleteProfileCache("*-profile-id")
+		return collectors.NoResult, err
+	}
```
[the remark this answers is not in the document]
brett060102: A very good point, and since collectors.CollectInformation always returns nil error, why does it return an error at all. And I don't know. That behavior predates my changes. It could be an artifact of old code that changed or as future proofing. I think the current code should stay. Since collectors.CollectInformation is speced to return an error, if the error is not nil, then the cache should be cleared. And the quest about whether collectors.CollectInformation should return an error, should be addresses separately. I have created an issue for this issue.

internal/connect/wrapper.go
```diff
@@ -122,15 +124,31 @@ func (w Wrapper) KeepAlive(uptimeTracking bool) error {
 		extraData["online_at"] = data
 	}
 
-	code, err := registration.Status(w.Connection, hostname, hwinfo, extraData)
-	if code != registration.Registered {
+	profileData, err := FetchSystemProfiles(arch, true)
```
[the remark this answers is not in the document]
brett060102: The behavior is speced differently in the two cases. It was speced the register must always go through. So, it is required to send registration even if the fetch profile fails.

internal/connect/wrapper.go
```diff
@@ -122,15 +124,31 @@ func (w Wrapper) KeepAlive(uptimeTracking bool) error {
...
+	if code == registration.ClearCache {
+		profiles.IncFailedProfileUpdate()
+		profiles.DeleteProfileCache("*-profile-id")
+	} else if code == registration.Unregistered {
+		profiles.DeleteProfileCache("*-profile-id")
 		return fmt.Errorf("trying to send a keepalive from a system not yet registered. Register this system first")
+	} else if err != nil {
+		profiles.DeleteProfileCache("*-profile-id")
+	} else {
+		profiles.ResetFailedProfileUpdate()
```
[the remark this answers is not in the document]
brett060102: yes, but I am using a bit of a hack and returning the 205 when X-System-Profile-Action: clear-cache is seen. So, lets talk about this one.

internal/connect/wrapper.go
```diff
@@ -148,9 +166,21 @@ func (w Wrapper) Register(regcode, instanceDataFile string) error {
...
+	code, _, err := registration.Register(w.Connection, regcode, hostname, hwinfo, extraData)
+
+	if code == registration.ClearCache {
+		profiles.IncFailedProfileUpdate()
+		profiles.DeleteProfileCache("*-profile-id")
+	} else if err != nil {
+		profiles.DeleteProfileCache("*-profile-id")
+	} else {
+		profiles.ResetFailedProfileUpdate()
+	}
```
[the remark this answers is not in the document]
brett060102: Sounds good.

internal/connect/collectors.go
brett060102: addressed

pkg/connection/connection.go
```diff
@@ -66,46 +66,50 @@ func (conn ApiConnection) BuildRequest(verb string, path string, body any) (*htt
...
 	}
+	request.URL.Path = path
 
 	conn.setupGenericHeaders(request)
 
 	return request, nil
 }
 
-func (conn ApiConnection) Do(request *http.Request) ([]byte, error) {
+func (conn ApiConnection) Do(request *http.Request) (int, []byte, error) {
```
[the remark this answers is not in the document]
paragjain0910: I would suggest below steps for the above : 

1) Define a simple interface in pkg/connection.go that defines the cache clearing contract

```
type ProfileCache interface {
	Clear()
}
```

2. Implement the interface in the pkg/profiles. Profile cache implements above interface.
```
func (cache *ProfileCache) Clear() {
    cache.DeleteProfileCache("*-profile-id")
}
```
3.  inject the profile Cache Manager in `internal/connect/wrapper.go` . This provide the cache manager to the connection so it can clear cache when needed.

```
// OLD
return &Wrapper{
    Connection: connection.New(connectionOpts, &creds),


// NEW
return &Wrapper{
    Connection: &connection.ApiConnection{
        Options:      connectionOpts,
        Credentials:  &creds,
        ProfileCache: &profiles.ProfileCache{},  // ← Only this added
    },
}
```
4) Handle profile cache clearing if requested by server ( Doing similar thing for system token)
```
	if response.Header.Get("X-System-Profiles-Action") == "clear-cache" {
		if conn.ProfileCache != nil {
			conn.ProfileCache.Clear()
		}
	}
```

Cache clearing happens automatically for all API calls, no manual status code checks needed.  if you want you can restrict as well to some api's. but i would suggest to just check for clear cache header irrespective of api's.

@brett060102 @felixsch @digitaltom  Please let me know of your thoughts about it. Hope it makes sense :)

internal/connect/wrapper.go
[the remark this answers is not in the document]
paragjain0910: Please have a look at the comment 
https://github.com/SUSE/connect-ng/pull/320#discussion_r2629901832

pkg/connection/connection.go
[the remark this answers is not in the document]
brett060102: https://github.com/SUSE/connect-ng/pull/341 is first pass at that.

internal/connect/wrapper.go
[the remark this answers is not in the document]
brett060102: I think that we could end up re-writing or removing information that we sent up before. But that really depends on what the SCC and RMT code does if a profiles had been specified on the last send, but are not in the current send.

internal/connect/wrapper.go
```diff
@@ -148,9 +165,15 @@ func (w Wrapper) Register(regcode, instanceDataFile string) error {
 		extraData["instance_data"] = string(data)
 	}
 
-	// NOTE: we are not interested in the code. Hence, we don't save it
-	// anywhere.
+	// Clear profile data before registration
+	profiles.DeleteProfileCache("*")
+	profileData, err := FetchSystemProfiles(arch, true)
+	extraData["data_profiles"] = profileData
```
gbuenodevsuse: Looking at [RMT](https://github.com/SUSE/rmt/blob/20e6de4c20d2009929f68b2b4110b8c394402915/app/controllers/api/connect/v3/subscriptions/systems_controller.rb#L21-L23) proposed changes, looks like the expected key is "system_profiles", not "data_profiles". Just making sure we've got this noted.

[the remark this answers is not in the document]
brett060102: fixed

mssola (approved):

PR #321, opened 2025-07-29
Bump golang to 1.24

ngetahun:
This PR upgrades the golang version from `1.21` to `1.24`.

### DIFF
--- a/internal/connect/extensions_test.go
+++ b/internal/connect/extensions_test.go
@@ -1,0 +1,149 @@
+package connect
+
+import (
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


mssola:
This has become stale, so thinking about this again, maybe it would be better to _add_ support for 1.24 (and multiple other Go versions if needed), while still maintaining 1.21. As it's done in many other places were a matrix of versions can be tested on CI. This would involve more effort than originally intended on this pull request, so let's close this one and retry again with a matrix on a new pull request whenever we have some time.

PR #324, opened 2025-08-13
Fix build for "test-yast"

mssola:
## Description

The setup for the `test-yast` Make target was utterly broken. This PR brings it back and tests are shown to be green. Moreover I have added a Github action just to be sure that we don't break it in the future.

## How to test

This should suffice:

```
$ make test-yast
```

### DIFF
--- a/.github/workflows/yast.yml
+++ b/.github/workflows/yast.yml
@@ -1,0 +1,14 @@
+name: YaST integration tests
+on: [pull_request]
+
+jobs:
+  yast:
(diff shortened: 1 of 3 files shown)


.github/workflows/yast.yml
```diff
@@ -0,0 +1,13 @@
+on: [pull_request]
```
felixsch, line 1:
```suggestion
name: YaST integration tests
on: [pull_request]
```

PR #325, opened 2025-08-13
Fix the 'check-format' make target

mssola:
The Github Action calling 'check-format' expected this target to return a bad exit code whenever code was not up to standards, but it was not becuase 'check-format' only showed the diff but did nothing on the exit code. Now this target behaves more as it was expected.

As a side-effect, a couple of files have been changed so this new check did not fail.


felixsch (approved):
lgtm :+1: 

CI is green and we running `check-format` again!

PR #330, opened 2025-09-01
collectors/sap: Don't repeat workloads

mssola:
If a workload has already been accounted for, don't add it again. This avoids situations in which a system with a vast number of the same workload is reported in an overwhelming way.

## Test

to do


mssola:
As discussed privately, we are getting this merged and then we can add a new schema if requested.

PR #333, opened 2025-09-08
Fix 2 occurrences of a module pseudo-constant being modified

rtamalin:
In Golang you can't define constants for certain types, such as maps.

However, it can be useful to define name pseudo-constants in modules but care must be taken to ensure that they are not modified by the code that uses them.

In this case registration modules defines two such pseudo-constants, NoExtraData and NoSystemInformation, but in a couple of places the code is copying those references and then modifying them which could lead to potential side effects.

Switching the problem code to just initialising an empty instance of the associated type resolves the issue.

### DIFF
--- a/cmd/offline-register-api/main.go
+++ b/cmd/offline-register-api/main.go
@@ -29,7 +29,7 @@ func main() {
 	arch := os.Args[3]
 	regcode := os.Args[4]
 
-	information := registration.NoSystemInformation
+	information := registration.SystemInformation{}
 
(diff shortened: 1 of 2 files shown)


felixsch:
I agree with this after learning about this kind of reference type. @rtamalin can you also remove this values from the `pkg/` entirely. Probably some cmd example needs to be updated too.

Thanks for the catch.

mssola (approved):

PR #353, opened 2026-03-10
Identify systemd unit files and query their properties

rtamalin:
Add a helper routine, Systemctl(), to the util package that leverages the Execute() helper to run systemctl commands.

Add a new type SystemdUnitFiles that supports querying properties of systemd unit files.

Add ListMatchingUnitFilesOfTypeAndState() helper routine, leveraging Systemctl(), returning a slice of SystemdUnitFiles, representing the set of systemd unit files of both the specified type and state that match the provided wildcard naming patern.

Add testing helper routines and associated types to simplify mocking Execute() for the various new scenarios that need to be tested in the util package. These new routines and types can be leveraged by other packages that leverage Execute() to potentially simplify any mocking needs their tests may have.

Add unit tests for the new Execute() mocking features. Additionally leverage these new mocking features to verify correct operation of the new Systemctl(), ListMatchingUnitFilesOfTypeAndState() routines and the SystemdUnitFile.Property() method.

Implements: TEL-358

### DIFF
--- a/internal/util/systemctl.go
+++ b/internal/util/systemctl.go
@@ -1,0 +1,34 @@
+package util
+
+var (
+	//
+	// systemctl settings
+	//
+
+	// path to the systemctl command
+	SystemctlBin = "/usr/bin/systemctl"
+
+	// extra options to pass to the
+	SystemctlBaseCmd = []string{
+		SystemctlBin,
+		"--no-pager",
+		"--nolegend",
+	}
+)
+
+func Systemctl(cmd string, args ...string) ([]byte, error) {
+	// construct systemctl command line
+	cmdLine := append([]string{}, SystemctlBaseCmd...)
+	cmdLine = append(cmdLine, cmd)
+	cmdLine = append(cmdLine, args...)
+
+	exitCodes := []int{0}
+
+	// execute systemctl command
+	output, err := Execute(cmdLine, exitCodes)
+	if err != nil {
+		return nil, err
+	}
+
+	return output, nil
+}
--- a/internal/util/system.go
+++ b/internal/util/system.go
@@ -40,8 +40,11 @@ func SetSystemEcho(v bool) bool {
 	return prev
 }
 
+// ExecuteFunc type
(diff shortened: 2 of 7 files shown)


internal/util/systemctl.go
```diff
@@ -0,0 +1,34 @@
...
+	//
+
+	// path to the systemctl command
+	SystemctlBin = "/usr/bin/systemctl"
+
+	// extra options to pass to the
+	SystemctlBaseCmd = []string{
+		SystemctlBin,
+		"--no-pager",
+		"--nolegend",
```
brett060102: I don't see this option in my buntu or sle 15 systemctl I do see:
--legend=BOOL       Enable/disable the legend (column headers and hints)

rjschwei (changes requested):

internal/util/systemctl.go
rtamalin: FYI the --no-legend is valid for the version of systemctl in 12 SP5 LTSS

```
admin@smt:~> systemctl --version
systemd 228
+PAM -AUDIT +SELINUX -IMA +APPARMOR -SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT -GNUTLS +ACL +XZ -LZ4 +SECCOMP +BLKID +ELFUTILS +KMOD -IDN
admin@smt:~> sudo systemctl --no-pager --no-legend show -p ExecStart smt.service
ExecStart={ path=/usr/lib/SMT/bin/smt-maintenance ; argv[]=/usr/lib/SMT/bin/smt-maintenance ; ignore_errors=no ; start_time=[Sat 2026-03-28 10:31:59 EDT] ; stop_time=[Sat 2026-03-28 10:32:01 EDT] ; pid=2347 ; code=exited ; status=0 }
```

rtamalin: I will fix the --nolegend to --no-legend

rtamalin:
Closing this as the relevant functionality that we want to keep has been extracted and enhanced as part of #393 and the rest we have decided to drop in favour of the native D-Bus based systemd query approach that was merged in #379

PR #364, opened 2026-03-25
Adds opt-in / opt-out

gbuenodevsuse:
## Collector Opt-In/Opt-Out Configuration                                                                                                                                                        
                                         
Allows enabling/disabling of optional data collectors while ensuring mandatory collectors always run.                                               
   
### Key Changes                                                                                                                                                                                                                                                     
  - Collector registry: Centralized metadata (mandatory, default enabled, type)
  - Mandatory enforcement: cannot be disabled
  - Optional collectors: Users can disable via config

  ### Config Example
  ```yaml
  url: https://scc.suse.com
  collectors:
    pci_devices:
      state: enabled
    kernel_modules:
      state: disabled


internal/connect/config.go
```diff
@@ -44,19 +44,19 @@ const (
...
+	Namespace                  string `json:"namespace" yaml:"namespace"`
 	FsRoot                     string
 	Token                      string
 	Product                    registration.Product
 	InstanceDataFile           string
-	Email                      string `json:"email"`
-	AutoAgreeEULA              bool
-	EnableSystemUptimeTracking bool
+	Email                      string `json:"email" yaml:"email"`
+	AutoAgreeEULA              bool   `yaml:"auto_agree_with_licenses"`
```
rtamalin, line 56: If we are adding the YAML tags here, is there a reason not to match them with JSON tags?

gbuenodevsuse: We are following what we had on the previous parser and which tags are expected on the config file.

rtamalin: Understood, but do we know what criteria is used to determine which fields to have JSON tags for?

Do we only JSON tag option fields that are also config file fields? If so then the YAML tagging should probably follow the same pattern...

Or is the JSON tagging just intended to expose interesting/useful option values in debug logging that relies on JSON marshalling the options as part of the log message construction... In which case it may make sense to align the JSON and YAML tags...

Or is there some other criteria used?

internal/collectors/config.go
```diff
@@ -0,0 +1,108 @@
...
+type CollectorRegistryEntry struct {
+	Metadata         CollectorMetadata
+	Collector        Collector
+	CollectorFactory func(updateDataIDs bool) Collector
+}
+
+// collectorsRegistry is the single source of truth for all collectors
+var collectorsRegistry = map[string]CollectorRegistryEntry{
+	// Mandatory system information collectors
+	"architecture":      {Metadata: CollectorMetadata{DefaultEnabled: true, Mandatory: true, Type: SystemInfoCollector}, Collector: Architecture{}},
```
rtamalin: It may be better to use a multi-line approach rather than a compact single line approach for long term readability and maintainability...

While the compact single line representation does mean that similar fields are visually aligned in columns but also leads to wrapping in narrow display windows like the side by side review setup, while also looking alike a wall of text ;-)

internal/collectors/config.go
```diff
@@ -0,0 +1,108 @@
...
+		Metadata:         CollectorMetadata{DefaultEnabled: true, Mandatory: false, Type: ProfileCollector},
+		CollectorFactory: func(updateDataIDs bool) Collector { return PCI{UpdateDataIDs: updateDataIDs} },
+	},
+	"kernel_modules": {
+		Metadata:         CollectorMetadata{DefaultEnabled: true, Mandatory: false, Type: ProfileCollector},
+		CollectorFactory: func(updateDataIDs bool) Collector { return LSMOD{UpdateDataIDs: updateDataIDs} },
+	},
+
+	// Optional/specialized collectors
+	"sap": {Metadata: CollectorMetadata{DefaultEnabled: false, Mandatory: false, Type: SystemInfoCollector}, Collector: SAP{}},
```
rtamalin: I thought that the existing SAP workload detection collector was enabled by default, and mandatory, and we just wanted to have the ability to turn it off in special cases to avoid issues like the bugs we dealt with recently?

internal/collectors/config.go
```diff
@@ -0,0 +1,108 @@
...
+	return ok
+}
+
+// DefaultCollectorState returns the default enabled state for a collector
+func DefaultCollectorState(collectorName string) bool {
+	if entry, ok := collectorsRegistry[collectorName]; ok {
+		return entry.Metadata.DefaultEnabled
+	}
+	return false
+}
```
rtamalin, line 147: Just wondering if it may better to define a CollectorsRegistry type, with the above functions as helper receiver methods, and an Init() receiver method that initialises the compiled in defaults? Potentially with an associated interface definition...

That would avoid the reliance on a hidden global variable, `collectorsRegistry`, with possible risk of side effects, and potentially could make it easier to implement complex unit testing scenarios... Though it would mean we would need to
plumb in allowing the CollectorsRegistry object to be passed around, or assigned as an attribute to an object that is already being passed around?

internal/connect/collectors.go
```diff
@@ -5,22 +5,19 @@ import (
...
-		collectors.Architecture{},
-		collectors.ContainerRuntime{},
-
-		// Optional collectors
-		collectors.Uname{},
-		collectors.SAP{},
-		collectors.Vendor{},
+	collectorOpts := GetCollectorConfig()
+
+	var usedCollectors []collectors.Collector
```
rtamalin, line 12: Minor thought... `enabledCollectors` rather than `usedCollectors` may be more descriptive?

rtamalin: However, it was the previously used variable name so fine to leave it as is...

internal/connect/collectors.go
```diff
@@ -34,11 +31,25 @@ func FetchSystemInformation(arch string) (collectors.Result, error) {
...
+// Parameters:
+//   - arch: system architecture (empty string to auto-detect)
+//   - updateCache: whether to update profile data IDs cache
 func FetchSystemProfiles(arch string, updateCache bool) (collectors.Result, error) {
-	var usedCollectors = []collectors.Collector{
-		collectors.PCI{UpdateDataIDs: updateCache},
-		collectors.LSMOD{UpdateDataIDs: updateCache},
+	collectorOpts := GetCollectorConfig()
+
+	var usedCollectors []collectors.Collector
```
rtamalin: Similar thought to above on used vs enabled in the name of this variable...

internal/connect/config.go
```diff
@@ -44,24 +44,35 @@ const (
...
+	EnableSystemUptimeTracking bool   `yaml:"enable_system_uptime_tracking"`
 	ServerType                 ServerType
-	NoZypperRefresh            bool
+	NoZypperRefresh            bool `yaml:"no_zypper_refs"`
```
rtamalin, line 59: Seems strange to want to persist this setting but I guess the original config parser supported it...

internal/connect/config.go
```diff
@@ -108,62 +119,78 @@ func ReadFromConfiguration(path string) (*Options, error) {
...
+	for collectorName, entry := range collectorConfigs {
+		if entry.Enabled == nil {
 			continue
 		}
-		key, val := strings.TrimSpace(parts[0]), strings.TrimSpace(parts[1])
-		if strings.HasPrefix(key, "#") {
+
+		// Validate collector exists
+		if !collectors.IsValidCollector(collectorName) {
+			return fmt.Errorf("unknown collector '%s' in configuration", collectorName)
```
rtamalin: Not sure we want to error out if we detect an unrecognised collector name at this point... May be better to just report a message warning that the collector name is unrecognised and continue to the next entry...

E.g. if someone downgraded to an older suseconnect version, that didn't know about a particular collector, but their config had been updated with an entry for it, we would likely want the customer to be able to run suseconnect without getting an error...

Similarly for developer testing, you may want to have a config file enabling or disabling a new connector, and be able to run both the "official" and the development versions of the tool in the same environment.

internal/connect/config.go
```diff
@@ -216,3 +243,16 @@ func (opts *Options) IsScc() bool {
 	}
 	return false
 }
+
+// Global collector configuration
+var collectorConfig collectorsconfig.CollectorOptions = &collectorsconfig.NoCollectorOptions{}
```
rtamalin: Minor concerns again about hidden global variables... But at least this time there are clear get & set routines...

internal/connect/config_test.go
```diff
@@ -80,3 +56,216 @@ func TestSaveLoad(t *testing.T) {
 		t.Errorf("got %+v, expected %+v", c2, c1)
 	}
 }
+
+func TestMinimalAndNonExistingConfiguration(t *testing.T) {
```
rtamalin, line 46: Some of these test cases could be expressed more simply using github.com/stretchr/testify's acquire and require, with expected values for the resulting configs to compare against...

And we already use stretchr/testify throughout the rest of the code...

Possibly something to consider for the tests you are adding at least, and we can go back and clean up the older tests later...

pkg/collectors/config.go
```diff
@@ -0,0 +1,23 @@
...
+type CollectorOptions interface {
+	// IsCollectorEnabled checks if a collector is enabled
+	IsCollectorEnabled(collectorName string) bool
+}
+
+// NoCollectorOptions is a default implementation that disables all collectors
+// Useful for testing and when collector options are not available
+type NoCollectorOptions struct{}
+
+func (NoCollectorOptions) IsCollectorEnabled(collectorName string) bool {
```
rtamalin, line 20: Does this mean that if no collector config is provided in the config file, we won't collect any of the mandatory collectors? That doesn't seem right... Mandatory should always be run, unless explicitly disabled by a config override...

May want to clarify the comment here to avoid ambiguity/confusion potential...

internal/collectors/config_test.go
```diff
@@ -0,0 +1,160 @@
+package collectors
```
rtamalin, line 1: For newer tests it might be better to standardise on github.com/stretchr/testify assert and require check patterns which are neater and handle things like structure equality checks well, allowing for more simple test expression.

internal/connect/collectors.go
```diff
@@ -5,22 +5,19 @@ import (
 	"github.com/SUSE/connect-ng/pkg/profiles"
 )
 
-// Fetch system information based on the available collectors for the system.
+// FetchSystemInformation collects basic system information from enabled collectors.
```
rtamalin: Would it be possible to add unit tests for the routines that are being modified/added in a collectors_test.go in this package area?

Again would be better to use stretchr/testify based testinbg patterns.

gbuenodevsuse: Added tests cases for both FetchSystemInformation and FetchSystemProfiles.

internal/collectors/config_test.go
gbuenodevsuse: Switched to stretcher/testify assertion pattern.

internal/connect/config_test.go
gbuenodevsuse: Switched to stretcher/testify assertion pattern.

internal/connect/config.go
gbuenodevsuse: My initial approach was in line with what you're pointing out. I was not sure about this one. After Tom's comment on the parser, I thought we should be more strict on the validations, but I agree. It's a simple enough change to skip it for now.

internal/connect/collectors.go
gbuenodevsuse: Maintained the previous names, but happy to change it. Had the same thought here when writing the code.

pkg/collectors/config.go
gbuenodevsuse: Clarified the comments in both stub implementation, and globals initialization. The Collectors will always respect Mandatory > User Spec > Defaults.

internal/connect/config.go
gbuenodevsuse: I don't really know the answer to this. I asked AI to help me identify JSON marshaling in the codebase and here is the answer it provided:

> The JSON tags mark fields exposed through the C library bindings (get_config/loadConfig in libsuseconnect). Fields with both json+yaml tags are part of the C API
>   AND saved to config. Fields with only yaml tags are saved to config but not exposed to C callers. The criteria for which fields get JSON tags isn't documented in the codebase - appears to be
>   historical.

internal/collectors/config.go
gbuenodevsuse: Fixed, switched to multi-line approach.

rtamalin (changes requested):
Please run `make` or `make all` in the docs subdirectory to generate the updated man pages from the corresponding .md files, and add them to the PR...

rtamalin (approved):
LGTM

I verified correct operation when registering against SCC and a local RMT. Also verified that unrecognised fields in a collector entry don't cause failure, and similarly neither do unrecognised collectors.

rtamalin (approved):
Still looks good.

PR #365, opened 2026-03-25
Adds a new rpm_packages collector

gbuenodevsuse:
This PR introduces an optional rpm_packages collector to gather installed system packages via the rpm database.

- Logic: Executes rpm -qa, filters for the "SUSE" vendor (case-insensitive), removes duplicates, and formats data into positional arrays.
- Settings: Added to internal/collectors/config.go as disabled by default (DefaultEnabled: false).
- Tests: Added unit tests in internal/collectors/rpm_packages_tests.go.

### Testing Instructions
1. Build locally:

    ```bash
    make build

2.  **Toggle configuration** in `/etc/SUSEConnect`:

    ```yaml
    collectors:
      rpm_packages:
        state: enabled # Toggle between enabled/disabled

3. Dry run to verify JSON output:

    ```bash
    sudo ./out/suseconnect -i

---

## Payload Size Estimation (openSUSE Leap 16.0)

To estimate the payload size, run:
```bash
sudo ./out/suseconnect -i | jq '.rpm_packages | tojson | length / 1024 | "\(.) KB"'
```
### Expected Output:
"118.58 KB"

**Note**: Actual payload size depends on the specific system profile and number of SUSE vendor packages installed, this is a rough estimation based on a local dev system.


internal/collectors/config.go
```diff
@@ -111,6 +111,10 @@ var collectorsRegistry = map[string]CollectorRegistryEntry{
 		Metadata:         CollectorMetadata{DefaultEnabled: true, Mandatory: false, Type: ProfileCollector},
 		CollectorFactory: func(updateDataIDs bool) Collector { return LSMOD{UpdateDataIDs: updateDataIDs} },
 	},
+	"installed_pkgs": {
+		Metadata:         CollectorMetadata{DefaultEnabled: true, Mandatory: false, Type: ProfileCollector},
```
jeremy-moffitt: The initial plan for the packages collector is to have it disabled by default and then see if we can get the openQA systems to enable it and check the incoming data volume so we don't overwhelm the SCC database.

gbuenodevsuse: Changed to disabled by default.

mssola:
This was open in March with not too much activity in it. Is there anything left to be done? Do you need more reviews to push this forward? Thanks!

gbuenodevsuse:
> This was open in March with not too much activity in it. Is there anything left to be done? Do you need more reviews to push this forward? Thanks!

Thanks for the follow up @mssola! From my side, the implementation is complete and ready for review. Happy to address any feedback if there are still any concerns.

docs/SUSEConnect.5
mssola, line 1: I'm not against this change, but it's unrelated to the PR, right? Let's have it on a separate PR.

internal/collectors/config.go
```diff
@@ -111,6 +111,10 @@ var collectorsRegistry = map[string]CollectorRegistryEntry{
 		Metadata:         CollectorMetadata{DefaultEnabled: true, Mandatory: false, Type: ProfileCollector},
 		CollectorFactory: func(updateDataIDs bool) Collector { return LSMOD{UpdateDataIDs: updateDataIDs} },
 	},
+	"installed_pkgs": {
+		Metadata:         CollectorMetadata{DefaultEnabled: false, Mandatory: false, Type: ProfileCollector},
+		CollectorFactory: func(updateDataIDs bool) Collector { return InstalledPackages{UpdateDataIDs: updateDataIDs} },
+	},
```
mssola, line 117: Do we have any agreement on why this should be optional and its default setting? Not arguing in favor or against anything, just trying to understand the context for this change :)

internal/collectors/installedpkgs.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/SUSE/connect-ng/internal/util"
+	"github.com/SUSE/connect-ng/pkg/profiles"
+)
+
+type InstalledPackages struct {
+	UpdateDataIDs bool
+}
+
+const PkgsChecksumFile = "pkgs.txt"
+const PackagesTag = "installed_pkgs"
```
mssola: Are these two constants ever used outside of the package? If not then we can write them lower case so they are not exported.

internal/collectors/installedpkgs.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (p InstalledPackages) run(arch string) (Result, error) {
+	util.Debug.Print("InstalledPackages.UpdateDataIds", p.UpdateDataIDs)
+
+	// This could come from CollectorOptions by adding a Params field to installed_pkgs collector
```
mssola: If this feature has not been added in this PR, let's not have comments which point to non-existing features 😉 

But check more on this below.

internal/collectors/installedpkgs.go
```diff
@@ -0,0 +1,103 @@
...
+	filterVendor := "SUSE"
+
+	qf := "%{VENDOR}\\t%{NAME}\\t%{VERSION}\\t%{RELEASE}\\t%{ARCH}\n"
+	pkgs, err := util.Execute([]string{"rpm", "-qa", "--qf", qf}, nil)
```
mssola: Nitpicking :)

Since this is used in code, could we have here the long parameter name "--queryformat" instead of the short "--qf"? Just realized that `rpm --help` actually does not show the short version, and you have to go to the manual to learn of it. Since the long version exists on both the manual and the help message, let's have the long one, which is, in the end, more clear.

internal/collectors/installedpkgs.go
```diff
@@ -0,0 +1,103 @@
...
+// filterPackages returns a sorted list of packages filtered by vendor
+// If filterVendor is empty, returns all packages
+func filterPackages(pkgs []byte, filterVendor string) ([]string, error) {
+	reader := bytes.NewReader(pkgs)
+	sc := bufio.NewScanner(reader)
+
+	var vendorRegex *regexp.Regexp
+	if filterVendor != "" {
+		// Regex to filter vendors (case insensitive)
+		vendorRegex = regexp.MustCompile(`(?i)` + regexp.QuoteMeta(filterVendor))
```
mssola: Having a MustCompile here is not great. Instead, MustCompile regexps are usually reserved into package variables, so they are executed right away and we can fail early. If this is something that the caller can provide, it's either a matter of providing a regexp as a parameter, or the callee should use `regexp.Compile`. The latter is better because it gives you an error object which you can then treat it appropriately. Note that now, if the string is borked, then we will actually panic if the regexp doesn't compile, which is bad (we should just return an error). Hence, for user-provided strings, using `regexp.Compile` makes more sense.

But, in the end, and as I pointed out on that "non-existing features" comment, why do we need this parameter at all? If we always pass "SUSE" (even in tests), why don't we just get rid off the "filterVendor" parameter and just go ahead and hardcode the "SUSE" filter on this function? I agree that hard-coding strings is not optimal (we can always disguise it into some "filterVendor" constant on the top of the file), but it's more realistic to what's really going on right now. This will make things simpler and more maintainable.

In the future, if we add some filtering option, then we can discuss on whether it's better to pass a string or a regexp, and all that jazz 😉

mssola (changes requested):

internal/collectors/config.go
gbuenodevsuse: We do. This was discussed as part of the design doc for this change. Any new collector coming in should be added as disabled by default for QA, and enabled on release if wanted as enabled by default.

brett060102: It is also disabled by default because this collector data is very large and while we expect some reuse across systems, we don't expect a lot. So, we don't really know how this will effect SCC DB size. So, the plan was disable by default and the enable of QA/Eng systems to get some base line data.

mssola: Perfect, thanks!

internal/collectors/installedpkgs.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return payload
+}
+
+// filterPackages returns a sorted list of packages filtered by vendor
+func filterPackages(pkgs []byte) ([]string, error) {
+	reader := bytes.NewReader(pkgs)
+	sc := bufio.NewScanner(reader)
+
+	vendorRegex, err := regexp.Compile(`(?i)` + regexp.QuoteMeta(filterVendor))
```
mssola: Since this is the only filtering we have, then it's safe to write `regexp.MustCompile` into the const definition.

internal/collectors/rpm_packages.go
```diff
@@ -0,0 +1,103 @@
...
+		version := fields[2]
+		release := fields[3]
+		arch := fields[4]
+		setKey := fmt.Sprintf("%s\t%s\t%s\t%s", name, version, release, arch)
+
+		// Remove duplicates
+		if _, ok := pkgSet[setKey]; !ok {
+			pkgSet[setKey] = struct{}{}
+			pkgList = append(pkgList, setKey)
+		}
```
mssola, line 73: Could we get an estimate on how many duplicate packages there are on a given system?

mssola (changes requested):

internal/collectors/rpm_packages.go
brett060102: I don't think that there can be any truly duplicate packages on a system package mangers won't allow that. But we do need to sort the output to make sure the list is consistent across systems.

rtamalin:
> This was open in March with not too much activity in it. Is there anything left to be done? Do you need more reviews to push this forward? Thanks!

The other thing to be aware of is that we wanted to implement a phase release approach; this functionality is dependent on the opt-in/opt-out config support functionality which is being released as part of the suseconnect-ng v1.22 release stream, so now that v1.22.1 is in train to be released, we can land this feature for inclusion in the v1.23 release stream.

brett060102 (commented):
I agree with suggested change. Checking that the regex compiles would only be needed if it was possible to change the regex. Aside from that this looks good.

internal/collectors/installedpkgs.go
rtamalin: Agreed that we should precompile the filters once, but it needs to be in a global var, as the result of a regexp.MustCompile() is not considered to be a constant.

rtamalin: Additionally right now `filterVendor` is just a simple `SUSE` without any regexp directives, so the use of this compiled regexp below could be replaced with a `strings.Contains(strings.ToUpper(vendor), filterVendor)`...

internal/collectors/installedpkgs.go
```diff
@@ -0,0 +1,103 @@
...
+
+		fields := strings.Split(pkg, "\t")
+		if len(fields) < 5 {
+			continue
+		}
+
+		vendor := fields[0]
+
+		// Apply vendor filter
+		if !vendorRegex.MatchString(vendor) {
```
rtamalin: See above comment that this could be replaced with a `strings.Contains()` if we don't plan to use a complex regexp for filtering...

internal/collectors/rpm_packages.go
```diff
@@ -0,0 +1,103 @@
...
+
+		name := fields[1]
+		version := fields[2]
+		release := fields[3]
+		arch := fields[4]
+		setKey := fmt.Sprintf("%s\t%s\t%s\t%s", name, version, release, arch)
+
+		// Remove duplicates
+		if _, ok := pkgSet[setKey]; !ok {
```
rtamalin, line 70: So we construct a `setKey` value that encodes the 4 fields we are interested in, and that uses that as value as a map key to construct a list that filters out subsequent occurrences of the packages... But then we return a list of setKey values, and we need to split them again in `formatPackagesPayload` to construct the payload value...

We could simplify this somewhat by appending `fields[1:5]` to `pkgList` rather than the `setKey` value, and then use `slices.Sort()` to sort the `pkgList` and leverage the `slices.Compare()` inside the provided sorting function.

This would allow us to return the sorted list, ready to be used as the payload, without having to process it again to split it up, as is currently the case.

internal/collectors/rpm_packages.go
```diff
@@ -0,0 +1,103 @@
...
+func filterPackages(pkgs []byte) ([]string, error) {
+	reader := bytes.NewReader(pkgs)
+	sc := bufio.NewScanner(reader)
+
+	vendorRegex, err := regexp.Compile(`(?i)` + regexp.QuoteMeta(filterVendor))
+	if err != nil {
+		return nil, fmt.Errorf("error compiling vendorRegex: %v", err)
+	}
+
+	pkgSet := make(map[string]struct{})
```
rtamalin, line 47: Is there a reason why this is a `map[string]struct{}`? We don't appear to be using it to store anything meaningful...

It could just as easily be a `map[string]bool`, in which case a lookup failure defaults to false, meaning that if you attempt to retrieve a key that is not assigned, you get a false...

This could be used to simplify the check below from needing to be an existence check to a simple true/false check, e.g. `if !pkgSet[setKey] { pkgSet[setKey] = true; ...`

internal/collectors/installedpkgs_test.go
```diff
@@ -0,0 +1,135 @@
...
+		"Another Corp\tmysuse-app\t1.0\t1\tx86_64\n"
+}
+
+func TestRunSuccessNoUpdate(t *testing.T) {
+	assert := assert.New(t)
+	setupTestData()
+
+	mockUtilExecute(pkgsTestData, nil)
+
+	// No Parameters provided - should default to filtering SUSE vendor
```
rtamalin: I think this comment about SUSE vendor is out of date with respect to the code?

rtamalin (commented):
Spotted a couple of nits and made a couple of enhancement suggestions.

internal/collectors/rpm_packages.go
mssola: Ok, then let's keep the logic for avoiding duplicates which, to be frank, shouldn't make that much of an impact performance-wise. What do you think @gbuenodevsuse

internal/collectors/rpm_packages.go
mssola: This is actually a quirk from golang, and I'd keep this as it is right now, to be honest. `struct{}` is actually the smallest type on golang, even smaller than bool considering implementation quirks from the runtime. Hence, if you just want to store that the key exists but don't care about the actual contents (as it is in this case), then, funnily enough, `struct{}` is the way to go.

Considering that we don't care about the stored value, and that it's added memory on a potentially large list, I think this micro-optimization makes sense.

gbuenodevsuse:
@rtamalin @mssola @brett060102 worked on a minor refactor to the filtering and payload formatting logic, addressing the feedback on previous comments.

internal/collectors/rpm_packages.go
gbuenodevsuse: Kept the duplicate removal logic as a safety net. Since it won't impact performance this much, IMHO better have it than not.

internal/collectors/config.go
```diff
@@ -111,6 +111,10 @@ var collectorsRegistry = map[string]CollectorRegistryEntry{
 		Metadata:         CollectorMetadata{DefaultEnabled: true, Mandatory: false, Type: ProfileCollector},
 		CollectorFactory: func(updateDataIDs bool) Collector { return LSMOD{UpdateDataIDs: updateDataIDs} },
 	},
+	"installed_pkgs": {
+		Metadata:         CollectorMetadata{DefaultEnabled: false, Mandatory: false, Type: ProfileCollector},
```
brett060102, line 115: why "installed_pkgs"  https://github.com/SUSE/happy-customer/blob/master/glue/app/models/profile.rb#L4  is ready to accept 'rpm_packages'

gbuenodevsuse: Nice catch, thanks. I was not aware of the key expected on SCC. I'll update it.

mssola (approved):
I have tested it by first running:

```
$ docker run --rm --privileged --network host -ti -v $(pwd):/connect registry.suse.com/bci/golang:1.24-openssl
```

Then inside:

```
% git config --global --add safe.directory /connect
% cd /connect
% make vendor
% make build
```

I've then verified that it returns nothing if the collector is not enabled in the configuration:

```
% ./out/suseconnect -i

# Output:
{
  "arch": "x86_64",
  "container_runtime": "docker",
  "cpus": 16,
  "hostname": "bubblegloop",
  "mem_total": 14739,
  "sockets": 1,
  "uname": "7.0.12-1-default #1 SMP PREEMPT_DYNAMIC Wed Jun 10 05:12:39 UTC 2026 (8beab0b)"
}
```

Then:

```
% cat <<HERE > /etc/SUSEConnect
> collectors:
>   rpm_packages:
>     state: enabled
> HERE

% ./out/suseconnect -i
```

That returned a big blob of rpm packages, that could be better parsed after installing `jq` into the container. Then I ran the suggested command:

```
% ./out/suseconnect -i | jq '.rpm_packages | tojson | length / 1024 | "\(.) KB"'

# Output:
"8.6533203125 KB"
```

Considering that containerized setups are quite frequent, I think this number is not so unrealistic and so having 8KB of data is not that big of a deal. But that is to be measured whenever this whole thing is deployed and real data starts kicking in.

Looks good!

PR #366, opened 2026-03-27
Bump default Go version for package builds to 1.24

rtamalin:
Switch the default Go package installed to support package builds to g1.24-openssl. This should be available on all currently supported release streams.


brett060102 (approved):
LGTM

PR #369, opened 2026-03-31
Fix system leak in feature tests

brett060102:
The feature test can fail to deregister systems because
    when suseconnect deactivates it gets a list of the release packages installed
    on the system and attempts to deactivate those on the SCC.
    
    But SCC may have activated extensions where the system does not have
    release packages installed. If the SCC has an active extension that
    depends on an release package installed on the system, the de-reg will fail.
    
    This is the situation in the current test setup. The current reg leak issue
    can be addressed by making sure the test container is in a clean state
    before starting the tests.
    
    This done by removing the unregistered products before starting the tests.

    Limit the go test to .go files, while not needed to address the
    registration leak, it keeps local testing from failing if stray files
    are in the test directory.

### DIFF
--- a/build/ci/run-feature-tests
+++ b/build/ci/run-feature-tests
@@ -7,6 +7,8 @@ SOURCE=${SOURCE:-/usr/src/connect-ng}
 # exported variables we expect to exist to run the tests
 KEYS_TO_CHECK=(REGCODE EXPIRED_REGCODE HA_REGCODE)
 
+REMOVE_PRODUCTS="sle-module-python3 sle-module-server-applications sle-module-basesystem"
+
 group() { echo "::group::$1"; }
 groupend() { echo "::groupend::"; }
 fail() { echo "::error::$1"; exit 1; }
@@ -24,8 +26,17 @@ group "cleanup test system"
   [ -f /etc/SUSEConnect ] && rm /etc/SUSEConnect
 groupend
 
+# Remove products so container in a clean state
+group "Clean Container"
+for product in ${REMOVE_PRODUCTS}; do
+    if zypper --no-refresh se --installed-only -t product ${product} | grep -q ${product}; then
+	   zypper --no-refresh --non-interactive remove -t product ${product}
+    fi
+done
+groupend
+
 group "run feature tests"
 pushd "$SOURCE"
- go test -timeout 15m -v features/suseconnect/*
+go test -timeout 15m -v features/suseconnect/*.go
 popd
 groupend
--- a/features/suseconnect/changed_root_test.go
+++ b/features/suseconnect/changed_root_test.go
@@ -14,13 +14,19 @@ func TestChangedRoot(t *testing.T) {
 	assert := assert.New(t)
 	env := helpers.NewEnv(t)
 
+	// cleanup any cruft left from container creation 
+	cli := helpers.NewRunner(t, "suseconnect -r %s", env.REGCODE)
+	cli.Run()
+	cli = helpers.NewRunner(t, "suseconnect -d")
+	cli.Run()
+
 	root := setupCustomRoot(t)
 	namespace := "test-namespace"
 
 	// No need to install rpm packages just registration
 	t.Setenv("SKIP_SERVICE_INSTALL", "true")
 
-	cli := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
+	cli = helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
 	cli.Run()
 
 	assert.Equal(0, cli.ExitCode())
@@ -33,6 +39,9 @@ func TestChangedRoot(t *testing.T) {
 	t.Run("check credentials location", func(t *testing.T) {
 		testChangedRootCredentialsLocation(t, root)
 	})
+	// dereg to remove created system from SCC
+	cli = helpers.NewRunner(t, "suseconnect --root %s -d", root)
+	cli.Run()
 }
 
 func setupCustomRoot(t *testing.T) string {
--- a/features/suseconnect/changed_root_test.go
+++ b/features/suseconnect/changed_root_test.go
@@ -14,7 +14,10 @@ func TestChangedRoot(t *testing.T) {
 	assert := assert.New(t)
 	env := helpers.NewEnv(t)
 
-	// cleanup any cruft left from container creation 
+	// cleanup any stale state left from container creation by registering and
+	// then cleanly deregistering which removes any related zypper repo services
+	// and associated release packages
+	
 	cli := helpers.NewRunner(t, "suseconnect -r %s", env.REGCODE)
 	cli.Run()
 	cli = helpers.NewRunner(t, "suseconnect -d")


features/suseconnect/changed_root_test.go
```diff
@@ -33,6 +33,12 @@ func TestChangedRoot(t *testing.T) {
 	t.Run("check credentials location", func(t *testing.T) {
 		testChangedRootCredentialsLocation(t, root)
 	})
+	// need to de-reg sle-module-systems-management first
+	cli = helpers.NewRunner(t, "suseconnect --root %s -d -p %s", root, "sle-module-systems-management/15.7/x86_64")
```
rtamalin: This hardcodes the distribution version to 15.7.

Also why do we need to specifically dereg the sle-module-systems-management before a general dereg?

brett060102: It it is not removed first then the the "-d" will fail because Systems Management Module 15 SP6 x86_64 depends on Python 3 Module 15 SP6 x86_64

`
        
        Python 3 Module 15 SP6 x86_64 (Activated)
        Deactivate with: suseconnect -d -p sle-module-python3/15.6/x86_64
            Systems Management Module 15 SP6 x86_64 (Activated)
            Deactivate with: suseconnect -d -p sle-module-systems-management/15.6/x86_64
`

rtamalin: Not seeing it when I simulated the test scenario in a 15 sp7 container run:
```
68d8fbdaa95a:/ # env SKIP_SERVICE_INSTALL=true suseconnect --root /test-root/ --namespace test-root -r ${REG_CODE} 
Registering system to SUSE Customer Center
Rooted at: /test-root/

Activating SLES 15.7 x86_64 ...

Activating sle-module-basesystem 15.7 x86_64 ...

Activating sle-module-server-applications 15.7 x86_64 ...

Activating sle-module-python3 15.7 x86_64 ...

Activating sle-module-systems-management 15.7 x86_64 ...

Successfully registered system
68d8fbdaa95a:/ # env SKIP_SERVICE_INSTALL=true suseconnect --root /test-root/ --namespace test-root -d             
Deregistering system to SUSE Customer Center
Rooted at: /test-root/

Cleaning up ...
Successfully deregistered system
68d8fbdaa95a:/ # 
```

rtamalin: I have tried a number of variations including not setting SKIP_SERVICE_INSTALL=true (requires you to do a manual `zypper ref` once after the first attempt to register fails to accept the repo signing keys, but works fine afterward) and I've not seen any issues with the deregister step:
```
68d8fbdaa95a:/ # suseconnect --root /test-root/ --namespace test-root -r ${REG_CODE} 
Registering system to SUSE Customer Center
Rooted at: /test-root/

Activating SLES 15.7 x86_64 ...
-> Adding service to system ...

Activating sle-module-basesystem 15.7 x86_64 ...
-> Adding service to system ...
-> Installing release package ...

Activating sle-module-server-applications 15.7 x86_64 ...
-> Adding service to system ...
-> Installing release package ...

Activating sle-module-python3 15.7 x86_64 ...
-> Adding service to system ...
-> Installing release package ...

Activating sle-module-systems-management 15.7 x86_64 ...
-> Adding service to system ...
-> Installing release package ...

Successfully registered system
68d8fbdaa95a:/ # suseconnect --root /test-root/ --namespace test-root -d             
Deregistering system to SUSE Customer Center
Rooted at: /test-root/

Deactivating sle-module-systems-management 15.7 x86_64 ...
-> Removing service from system ...
-> Removing release package ...

Deactivating sle-module-python3 15.7 x86_64 ...
-> Removing service from system ...
-> Removing release package ...

Deactivating sle-module-server-applications 15.7 x86_64 ...
-> Removing service from system ...
-> Removing release package ...

Deactivating sle-module-basesystem 15.7 x86_64 ...
-> Removing service from system ...
-> Removing release package ...
-> Removing service from system ...

Cleaning up ...
Successfully deregistered system
68d8fbdaa95a:/ # 
```

rtamalin: I just want to avoid the hard coding if at all possible.

rtamalin: Talking with @brett060102 I was running my container test cases with the SLE BCI 15.7 base image, not the go1.24-openssl image.

The latter image appears to retain zypper product state from when the image was created:
```
# zypper products
Refreshing service 'container-suseconnect-zypp'.
Loading repository data...
Reading installed packages...
S  | Repository | Internal Name                  | Name                                | Version | Arch   | Is Base
---+------------+--------------------------------+-------------------------------------+---------+--------+--------
i+ | @System    | sle-module-server-applications | Server Applications Module          | 15.7-0  | x86_64 | No
i+ | @System    | sle-module-python3             | Python 3 Module                     | 15.7-0  | x86_64 | No
i+ | @System    | sle-module-basesystem          | Basesystem Module                   | 15.7-0  | x86_64 | No
i+ | @System    | SLES                           | SUSE Linux Enterprise Server 15 SP7 | 15.7-0  | x86_64 | Yes
   | SLE_BCI    | SLE_BCI                        | SUSE Linux Enterprise BCI 15 SP7    | 15.7-0  | x86_64 | No
```

rtamalin: We resolved this by cleaning up the stale zypper product state as part of the test setup.

rtamalin:
We should update this branch to be ready to merge, and get it in.

brett060102:
@rtamalin I have synced to current main.

features/suseconnect/changed_root_test.go
```diff
@@ -14,13 +14,19 @@ func TestChangedRoot(t *testing.T) {
 	assert := assert.New(t)
 	env := helpers.NewEnv(t)
 
+	// cleanup any cruft left from container creation
```
rtamalin: Possibly expand this comment to explain the cleanup logic.


```suggestion
	// cleanup any stale state left from container creation by registering and
	// then cleanly deregistering which removes any related zypper repo services
	// and associated release packages
	
```

rtamalin:
Also wondering if the "cleanup" should be something we do for all test cases being run by the feature tests, rather than just these specific feature tests?

And now that I think about it, this cleanup is probably working around the side effect of the issue that @chajain is fixing in PR#377 so do we need it?

build/ci/run-feature-tests
```diff
@@ -24,8 +24,15 @@ group "cleanup test system"
   [ -f /etc/SUSEConnect ] && rm /etc/SUSEConnect
 groupend
 
+# Remove release packages so container in a clean state
+group "Clean Container"
```
rtamalin, line 30: You could define a variable (array or string) holding the list of products to cleanup and then have a for loop iterate over it to delete the specified products.

That would be more maintainable in the long run as it would mean you just need to add to/update the list of products to change things.

We could also make the list variable initialisation dependent on the detected OS distro version which will simplify the transition from SLE 15 based container images to SLE 16 ones whenever that happens, given that the set of potential products are different.

For example something like:
```bash
cleanup_products=(
    sle-module-python3
    sle-module-server-applications
    sle-module-basesystem
)
for prod in "${cleanup_products[@]}"
do
    zypper --no-refresh --non-interactive remove -t product "${prod}"
done
```

build/ci/run-feature-tests
```diff
@@ -24,8 +24,15 @@ group "cleanup test system"
...
+zypper --no-refresh --non-interactive remove -t product sle-module-python3
```
rtamalin: You may want to consider adding `|| true` on the end of these zypper commands to avoid causing a script failure due to a product name not being recognised by zypper, given that we have a `set -e` at the start of the script...

Though, not doing so will have the "benefit" that the first unlucky person to submit a PR after the underlying BCI Golang image changes to being say SLE 16 based will be forced to fix the script as part of their possibly unrelated PR submission to get the CI working again...

See my above comment about using a list to make things simpler to maintain and possibly setting things up for distro version handling... ;-)

build/ci/run-feature-tests
```diff
@@ -24,8 +24,15 @@ group "cleanup test system"
...
+zypper --no-refresh --non-interactive remove -t product sle-module-server-applications
+zypper --no-refresh --non-interactive remove -t product sle-module-basesystem
+groupend
+
 group "run feature tests"
 pushd "$SOURCE"
- go test -timeout 15m -v features/suseconnect/*
+go test -timeout 15m -v features/suseconnect/*_test.go
```
rtamalin: While there are advantages to this change, I don't think it is necessary anymore...

One of my intermediary hacking iterations for the RKE2 & K3s detection code that I shared with you, where I was trying to use embed to pull in the content of the k8s provider emulation script caused the simple wild card match to fail, but it shouldn't be an issue anymore, so I just want to understand the justification for doing so.

Note also the recommendation would be to use `*.go` rather than `*_test.go` since go test already handles figuring out the test part.

brett060102: I think that this is still needed.
I don't know what is different from you you are doing and what I am doing,
when I do "feature-tests" from origin/main
I see a registration remaining after running the TestChangedRoot test and that tests reports:

=== RUN   TestChangedRoot/check_credentials_location
[cleanup] Try running suseconnect --de-register...
 => Failed to run suseconnect --de-register. Skipping..


I added  features/suseconnect/*_test.go to avoid:
::group::run feature tests
/usr/src/connect-ng /usr/src/connect-ng
named files must be .go files: features/suseconnect/changed_root_test.go_save
FAIL	 [setup failed]
FAIL
make: *** [Makefile:127: feature-tests] Error 1
when some like makes a backup of test file  and forgets to remove it.

brett060102: Sorry, did not see "*.go" part. I will change to that.

build/ci/run-feature-tests
brett060102: @rtamalin Added comment about "*.go" change to both commit message and the PR comments. As discussed created var with list of products to remove and use for loop to run over that list. Also, put test into the loop to only remove the product if it is installed. Which I should have done in the first place.

build/ci/run-feature-tests
```diff
@@ -24,8 +26,17 @@ group "cleanup test system"
   [ -f /etc/SUSEConnect ] && rm /etc/SUSEConnect
 groupend
 
+# Remove products so container in a clean state
+group "Clean Container"
+for product in ${REMOVE_PRODUCTS}; do
+    zypper --no-refresh se --installed-only -t product ${product} | grep -q ${product} && {
+	zypper --no-refresh --non-interactive remove -t product ${product}
+	}
```
rtamalin: Since you have effectively implemented an if statement this would read better as one...

```suggestion
    if zypper --no-refresh se --installed-only -t product ${product} | grep -q ${product}; then
	    zypper --no-refresh --non-interactive remove -t product ${product}
    fi
```

brett060102: changed, but I prefer && in cases like this.

rtamalin: Just be aware though that the use of the && construct leaves you open to the same failure scenario that would require an `|| true` on the end to avoid a script failure when `set -e` is in effect.

The `set -e` exit on failure handling kicks in if a command line completes with a non-zero exit status, and since your command line uses an `&&` that means than a non-zero status from the grep becomes the non-zero status for the command line... Thus if the grep fails to match you get the command line failing and the script exits. For this reason it is usually better to construct a negative check when possible and use the `|| { do something }` pattern so that you are insulating yourself from unintended exit on error triggers.

In this case an `if ...; then ...; fi` is more readable and avoids the unintended exit on failure issue because an if block always has zero exit status, no matter whether any conditions were met.

rtamalin (approved):
LGTM

build/ci/run-feature-tests
brett060102: I tested the && {  where the grep fails which gives ! exit and that did not trigger an abort. I  checked the behavior when zypper fails and did not see any difference between && { and if/then.

PR #371, opened 2026-04-01
Add call tracing to exported libsuseconnect routines

rtamalin:
Add a helper trace() routine to libsuseconnect that wraps calling util.Info.Printf() to print tracing messages for the majority of entry and exit paths in the exported routines.

Tried using util.Debug rather than util.Info with debug output enabled but wasn't seeing any of the trace messages show up in the YaST2 logs.

Also need to avoid tracing the exit from the set_log_callback() routine as doing so can
cause a recursive deadlock failure at the Ruby level.

Implements: #363

### DIFF
--- a/third_party/libsuseconnect/libsuseconnect.go
+++ b/third_party/libsuseconnect/libsuseconnect.go
@@ -26,6 +26,12 @@ import (
 	"github.com/SUSE/connect-ng/pkg/search"
 )
 
+// trace entry & exit to exported calls
+func trace(fmt string, args ...interface{}) {
+	// use util.Info for now as util.Debug output doesn't show up in /var/log/YaST2/y2log
+	util.Info.Printf("libsuseconnect - "+fmt, args...)
+}
+
 // log level
 const (
 	llDebug   = 1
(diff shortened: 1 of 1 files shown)


brett060102:
I don't think that we should write debug like this to stdout. And you can use util.Debug to write to the yast log file but it requires a couple of changes and yast needs to fix a bug that they have.
Based on what I found in testing and from google. There should be two way to enable yast2 debug output:
1: debug = true in /etc/YaST2/log.conf 
2: set env Y2DEBUG=1

These should result in the same behavior but they don't. YasT sends us clientParms of the form:
clientParams: {"language":null,"debug":null,"verbose":true,"verify_callback":"#<Proc:0x000055fe337929f8@/usr/share/YaST2/lib/registration/registration.rb:332 (lambda)>","url":"https://scc.suse.com","token":""}

```
setting debug = true in /etc/YaST2/log.conf seems to have no effect on what is sent to us.
Y2DEBUG=1 results in our being passed "verbose":true 

So, the following will let us use util.Debug to set messages to the yast log:
 //export write_config
 func write_config(clientParams *C.char) *C.char {
-       trace("write_config - call args - clientParams: %s", C.GoString(clientParams))
        opts := loadConfig(C.GoString(clientParams))
+       trace("write_config - call args - clientParams: %s", C.GoString(clientParams))
 
        err := opts.SaveAsConfiguration()
        if err != nil {
@@ -320,10 +320,14 @@ func loadConfig(clientParams string) *connect.Options {
        // unmarshal extra config fields only for local use
        var extConfig struct {
                Debug string `json:"debug"`
+               Verbose bool `json:"verbose"`
        }
+       trace("load_config - call args - clientParams: %s", clientParams)
        json.Unmarshal([]byte(clientParams), &extConfig)
+       trace("load_config - extConfig: %+v", extConfig)
        // enable debug output if "debug" was set in json
-       if v, _ := strconv.ParseBool(extConfig.Debug); v {
+
+       if extConfig.Verbose {
                trace("loadConfig - enable debug output")
                util.Debug.SetOutput(callbackWriter{llDebug})
        }
```
when Y2DEBUG=1

The reference I found on the log.conf might be stale. It was from HPE docs and I don't see it in our docs. https://documentation.suse.com/sles/15-SP7/html/SLES-all/cha-yast-gui.html

This is supposed to be away to enable this at boot time, but I have not tried it

PR #372, opened 2026-04-03
Switch to util.Debug

brett060102:
1: change trace to use util.Debug
2: add call to trace at start of loadConfig
3: update loadConfig to set up debug output if
   "debug" : ["1", "t", "T", "true", "True", "TRUE"] or
   "verbose" : true
The any call to trace before loadConfig has enabled debug output will be discarded, this includes the first trace in loadConfig the first time loadConfig is called.


rtamalin:
Switching to be a draft PR given that it is based upon a draft PR submission.

PR #375, opened 2026-04-15
WIP: Update suseconnect-ng to be based upon go 1.24

rtamalin:
This will allow future code submissions to take advantage of newer Go features.

Updated go.mod to pull in latest version of stretchr/testify and associated indirect dependencies.

Updated PACKAGE.md to reflect current maintenance update workflows.

Also updated the clean rule to reflect historical changes to the naming of intermediate artifacts.

Created some additional Makefile variables to store values that are used repeatedly and updated impacted targets to leverage these new variables.

Added additional Makefile targets to support common activities:
  * deps - runs go get -u and go mod tidy to update dependencies
  * vet - runs go vet on the code base
  * generate-version - triggers the generation of the version.txt
  * run-tests - runs all the current test suites
  * go-clean - runs go clean
  * real-clean - cleans up all ephemeral artifacts

Updated some Makefile target dependencies to leverage vet and made vet dependent on vendor, ensuring that vendor is triggered for most relevent scenarios.

Similarly updated some target dependencies to avoid deleting the previously generated version file when being leveraged as part of a higher level target's dependencies

Address errors in identified by go vet in the main code base:
  * cmd/zypper-migration/migration.go
  * internal/connect/config.go

Updated some Makefile targets to fix environmental side-effects such as files in vendor tree being owned by root.

Added initial support for generating combined coverage statistics reporting for the existing test suites, though additional work will be required to enable generation of counter data for the Yast tests. Coverage reporting is currently disabled by default, controlled by the new COVERAGE variable. Set COVERAGE=true via Makefile command line parameter or via environment variable to enable the feature locally.

Notes about coverage reporting:
  * Further work will be needed to integrate this into the GitHub Actions CI workflows to upload the collected coverage data as artifacts from each of the CI test jobs, with an additional CI test job that downloads the artifacts to merge them and report the combined stats.
  * While we can build the libsuseconnect.go with coverage support enabled, an initial action is needed at process exit to flush the collected counters to disk, which will require exporting a additional coverage flushing routine from the library, and also require updates to either the suseconnect Ruby bindings, or the yast-registration test suite to call the flush routine at process exit.

Updated the suseconnect-ng.spec to add support for a coverage bcond.

Updated the build-rpm script to enable the coverage bcond if needed when building the RPMs from the spec file.

Updated the run-feature-tests spec file to setup coverage collection if enabled.

Updated current test suite Makefile targets to report coverage stats if coverage reporting is enabled.

Added coverage and coverage-percent Makefile targets that report the stats for the combined set of coverage data from the most recent runs of each test suite.


Makefile
```diff
@@ -1,42 +1,59 @@
...
-GOFLAGS       = -v -mod=vendor
+COVERAGE     ?= false
+GOFLAGS       = -v -mod=vendor $(if $(filter true,$(strip $(COVERAGE))),-cover)
 BINFLAGS      = -buildmode=pie
 SOFLAGS       = -buildmode=c-shared
+COVERAGE_DIR  = $(PWD)/coverage
+COV_UNIT      = $(COVERAGE_DIR)/unit
+COV_FEATURE   = $(COVERAGE_DIR)/feature
+COV_YAST      = $(COVERAGE_DIR)/yast
+COV_MERGED    = $(COVERAGE_DIR)/merged
```
gbuenodevsuse, line 16: Is the naming scheme change related to the usage of `go tool covdata`?
I'd suggest keeping the COVERAGE related envs naming consistent throughout.

```suggestion
COVERAGE_UNIT      = $(COVERAGE_DIR)/unit
COVERAGE_FEATURE   = $(COVERAGE_DIR)/feature
COVERAGE_YAST      = $(COVERAGE_DIR)/yast
COVERAGE_MERGED    = $(COVERAGE_DIR)/merged
```

gbuenodevsuse (approved):
Minor comment on the envs naming.
LGTM.

digitaltom:
What is the reason for adding the coverage in the rpm spec? Typically that would get shown as part of the CI, with a CI failure when the target percentage is not hit.

brett060102:
@digitaltom @rtamalin goal to to collect and  collate  all testing data to get a full report. This means combining feature test data with unit and yast test coverage data, but the feature test process is to build the RPM, install it and then run the tests.

rtamalin:
> What is the reason for adding the coverage in the rpm spec? Typically that would get shown as part of the CI, with a CI failure when the target percentage is not hit.

To enable coverage collection for the feature tests we need to build the suseconnect binary that the feature tests are running with coverage collection enabled.

And since we are installing the RPM packages built by the build-rpm script to make those binaries available for the feature tests to run, keeping that workflow means we need to add support to the spec file to enable building the RPM packages with coverage enabled.

Alternate approaches are possible:
* maintain a patch that gets applied to the spec file by the build-rpm script to enable coverage collection if desired.
* restructure the feature tests to run from locally built suseconnect binaries with coverage collection enabled.

The latter approach is kinda what the `make test-yast` does, i.e. build the binaries locally and copy them into the customised Yast testing container at appropriate locations.

I'm happy to re-work this feature test coverage collect approach and split the coverage stuff out into a separate PR going forward.

**NOTE**: If you just enable coverage collection for the feature test `go test ...` command you actually only collect coverage from the feature test code itself, rather than for the suseconnect code base. In fact doing so causes problems with collecting coverage data from the suseconnect binaries themselves unless you add special GOCOVERDIR setup handling into the test setup and possibly suseconnect itself because `go test -cover` sets up it's own GOCOVERDIR settings for child processes that may be invoked that supersedes any environment settings that we may have setup, causing the coverage data from the binaries to be set to a random /tmp path. ;-)

build/ci/run-feature-tests
```diff
@@ -4,6 +4,9 @@ set -e
 # Where the connect-ng source is located within the container
 SOURCE=${SOURCE:-/usr/src/connect-ng}
 
+# test coverage profile directory
+COV_FEATURE=${SOURCE}/coverage/feature
```
digitaltom, line 8: I think calling this GOCOVERDIR would leave us with one less new name.

build/packaging/suseconnect-ng.spec
```diff
@@ -150,6 +158,11 @@ ln -sf service %{buildroot}/%{_sbindir}/rcsuse-uptime-tracker
 # we currently do not ship the source for any go module
 rm -rf %{buildroot}/usr/share/go
 
+%if %{with coverage}
+# Create the directory where covdata will be written
+mkdir -p %{buildroot}/var/lib/suseconnect/coverage
```
digitaltom, line 163: Why is the coverage directory part of the rpm?

build/ci/run-feature-tests
rtamalin: I was just doing this to keep the variable in sync with the name used in the top level Makefile. But yeah we can just call this GOCOVERDIR here, and then below we can have the export just specify the variable name without an assignment...

build/packaging/suseconnect-ng.spec
rtamalin: That was part of an effort to shut up a warning message that the a Go coverage enabled binary will generate when run complaining that the GOCOVERDIR target directory doesn't exist... Having the RPM create the "default" coverage dir would avoid that complaint. However we would still need to set that "default" value somewhere, e.g. via a /etc/profiles.d scriptlet or similar...

Makefile
rtamalin: Can be done. Was just trying to keep the variable names from getting too large which would necessitate changing the whitespace in all of the other variables in this block of assignments.

build/ci/run-feature-tests
rtamalin: Note that I was avoiding passing GOCOVERDIR as an env var when calling this script because if it was set in the host env and passed from there when running within a container, the host path would not be valid in the container, and we have to determine it relative to the location of the source tree within the container.

We could use a Dockerfile to customise the container image we use to run the tests, and have that Dockerfile setup the appropriate env to inject into a run, which would allow the script to be written to accept a GOCOVERDIR env setting. That way the docker based run or a manual run by a user from the command line could both rely on an env setting without having to implement any special logic...

digitaltom: I'd keep it as simple as possible, this is not a customer feature, and for us it should be fine?

build/packaging/suseconnect-ng.spec
digitaltom: But when we call the binary in our CI/local tests, we can make sure ourselves to set GOCOVERDIR to an existing directory, no?

jeremy-moffitt:
@rtamalin  is this still relevant or superseded by https://github.com/SUSE/connect-ng/pull/380 ?

mssola:
@rtamalin I believe this has been superseded by multiple PRs, right? Can we close it?

PR #376, opened 2026-04-16
Harden pull request workflows

voving30:
This updates the pull request workflows to tighten permissions and reduce secret exposure.

- Set `contents: read` explicitly for all active PR workflows.
- Scope registration secrets to the steps that actually need them.
- Remove the unused `REGCODE` from the RPM workflow.
- Update all `actions/checkout` references to `v7`.
- Add `.env` as an `act` secret file and document the local setup.

### DIFF
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -17,22 +17,23 @@ name: feature tests
 # See: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
 on: [pull_request]
 
+permissions:
+  contents: read
+
 env:
   SOURCE: /usr/src/connect-ng
   ARTIFACT_SOURCE: artifacts
-  REGCODE: ${{ secrets.REGCODE }}
-  HA_REGCODE: ${{ secrets.HA_REGCODE }}
-  EXPIRED_REGCODE: ${{ secrets.EXPIRED_REGCODE }}
 
 jobs:
   feature-tests:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     container:
       image: registry.suse.com/bci/golang:1.24-openssl
       options: --user root --privileged
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
 
       - name: move source to /usr/src/connect-ng
         run: |
@@ -49,6 +50,10 @@ jobs:
       #     path: ${{ env.SOURCE }}/${{ env.ARTIFACT_SOURCE }}/*.rpm
 
       - name: configure Connect to run feature tests
+        env:
+          REGCODE: ${{ secrets.REGCODE }}
+          HA_REGCODE: ${{ secrets.HA_REGCODE }}
+          EXPIRED_REGCODE: ${{ secrets.EXPIRED_REGCODE }}
         run: bash build/ci/configure
 
       - name: run feature tests
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -57,4 +57,8 @@ jobs:
         run: bash build/ci/configure
 
       - name: run feature tests
+        env:
+          REGCODE: ${{ secrets.REGCODE }}
+          HA_REGCODE: ${{ secrets.HA_REGCODE }}
+          EXPIRED_REGCODE: ${{ secrets.EXPIRED_REGCODE }}
         run: bash build/ci/run-feature-tests
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -33,7 +33,7 @@ jobs:
       options: --user root --privileged
 
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
       - name: move source to /usr/src/connect-ng
         run: |
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -20,7 +20,7 @@ jobs:
       options: --user root --privileged
 
     steps:
-      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
 
       - name: move source to /usr/src/connect-ng
         run: |
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -3,23 +3,24 @@ name: feature tests
 
 on: [pull_request]
 
+permissions:
+  contents: read
+
 env:
   SOURCE: /usr/src/connect-ng
   ARTIFACT_SOURCE: artifacts
   COVERAGE_BIN: true
-  REGCODE: ${{ secrets.REGCODE }}
-  HA_REGCODE: ${{ secrets.HA_REGCODE }}
-  EXPIRED_REGCODE: ${{ secrets.EXPIRED_REGCODE }}
 
 jobs:
   feature-tests:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     container:
       image: registry.suse.com/bci/golang:1.26-openssl
       options: --user root --privileged
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
 
       - name: move source to /usr/src/connect-ng
         run: |
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -35,6 +35,10 @@ jobs:
           make build
 
       - name: run feature tests
+        env:
+          REGCODE: ${{ secrets.REGCODE }}
+          HA_REGCODE: ${{ secrets.HA_REGCODE }}
+          EXPIRED_REGCODE: ${{ secrets.EXPIRED_REGCODE }}
         run: |
           set -e
           cd $SOURCE
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -20,7 +20,7 @@ jobs:
       options: --user root --privileged
 
     steps:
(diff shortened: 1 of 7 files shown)


voving30:
Follow-up fix: pass REGCODE, HA_REGCODE and EXPIRED_REGCODE to the "run feature tests" step as well, since build/ci/run-feature-tests reads these variables directly and not only through build/ci/configure.

.github/workflows/features.yml
```diff
@@ -17,22 +17,23 @@ name: feature tests
...
   feature-tests:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     container:
       image: registry.suse.com/bci/golang:1.24-openssl
       options: --user root --privileged
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
```
rtamalin: Would be a good idea to retain a comment indicating which version of the actions/checkout this corresponds to, e.g. 
```suggestion
      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
```

.github/workflows/lint-unit.yml
```diff
@@ -26,7 +29,7 @@ jobs:
       image: registry.suse.com/bci/golang:1.24-openssl
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
```
rtamalin: See previous comment about retaining the version association via a comment at least

.github/workflows/yast.yml
```diff
@@ -1,11 +1,14 @@
...
 
+permissions:
+  contents: read
+
 jobs:
   yast:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
```
rtamalin: See previous comment about retaining the version association via a comment at least

.github/workflows/features.yml
```diff
@@ -17,22 +17,23 @@ name: feature tests
...
 env:
   SOURCE: /usr/src/connect-ng
   ARTIFACT_SOURCE: artifacts
-  REGCODE: ${{ secrets.REGCODE }}
-  HA_REGCODE: ${{ secrets.HA_REGCODE }}
-  EXPIRED_REGCODE: ${{ secrets.EXPIRED_REGCODE }}
 
 jobs:
   feature-tests:
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
rtamalin, line 16: This is a good restriction to add as explicitly called out in the workflow definitions, though the equivalent effect is currently already in effect using GitHub settings to disable running the workflows for submissions from forks, as can be seen by the fact that workflows didn't get run for this PR.

.github/workflows/features.yml
voving30: Done, added the corresponding version comment next to the pinned checkout SHA in all three workflow files

.github/workflows/lint-unit.yml
voving30: Done, added the corresponding version comment next to the pinned checkout SHA in all three workflow files

.github/workflows/yast.yml
voving30: Done, added the corresponding version comment next to the pinned checkout SHA in all three workflow files

.github/workflows/features.yml
voving30: Good point, thanks. I added it to make the trust boundary explicit in the workflow itself as a defense-in-depth measure, rather than relying only on repository settings.

mssola:
Could you rebase this PR so we can review it again? Sadly Github worflow files have been heavily modified by previous PRs.

voving30:
Rebased onto the latest main and resolved the workflow conflicts while preserving the updated upstream workflow changes.

.github/workflows/features.yml
```diff
@@ -3,23 +3,24 @@ name: feature tests
...
     runs-on: ubuntu-latest
     container:
       image: registry.suse.com/bci/golang:1.26-openssl
       options: --user root --privileged
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
mssola: There's even v7 of the Github action. Do we really need to pin to this specific sha?

mssola: Same for the rest.

mssola (commented):
Small comment from my side. Good job!

voving30:
Updated all three checkout references to the v7.0.0 release SHA while keeping full-SHA pinning. Thanks for catching it!

mssola (approved):
LGTM 🎉

voving30:
Hi @mssola @rtamalin - just checking if anything else is needed from my side to move this forward.
The remaining Agama failure appears to be related to REGCODE not being available to fork pull requests. The branch is also out of date but cleanly mergeable, so I’ve left it untouched in case you prefer a specific rebase/merge path

jeremy-moffitt:
@voving30 please rebase again and we'll try to get this landed, thanks!

voving30:
Done, rebased onto the latest main. Thanks!

rtamalin:
Yeah the feature and agama tests can't be run for forked repos because they require secrets that are only available to PRs based on branches within the main repo for security reasons.

.github/workflows/features.yml
rtamalin: So digging in to this some more I'm not sure that the benefit of of adding this check is worth the negative impacts...

As already called out, we already have the intended benefit implicitly because lack of access to the required secrets causes the feature-test run to fail in forked repos.

Additionally this change also causes the feature-tests workflow to be skipped when testing locally with github.com/nektos/act unless a specially crafted event JSON blob is specified when running `act` command.

So overall I don't think this is worth keeping.

.github/workflows/features.yml
```diff
@@ -3,23 +3,24 @@ name: feature tests
...
 on: [pull_request]
 
+permissions:
+  contents: read
+
 env:
   SOURCE: /usr/src/connect-ng
   ARTIFACT_SOURCE: artifacts
   COVERAGE_BIN: true
-  REGCODE: ${{ secrets.REGCODE }}
```
rtamalin, line 10: Removing these top-level env var settings seems to break running the feature tests locally with github.com/nektos/act as the env vars specified in the `.env` file are no longer available later in the context of the steps within the feature-tests job, unless the `--secret-file .env` file is also specified to the `act` command.

The good news is that this can be done by added a line like so to the `.actrc`
```
--secret-file .env
```

.github/workflows/features.yml
```diff
@@ -3,23 +3,24 @@ name: feature tests
...
   feature-tests:
+    if: github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     container:
       image: registry.suse.com/bci/golang:1.26-openssl
       options: --user root --privileged
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```
rtamalin: We discussed this PR among our team and today and while there are benefits to hardening our GitHub Actions workflows by using hardcoded known-to-be-good SHA references, it also comes with the on-going maintenance burden that we have to monitor the associated actions implementations for important updates such as CVE fixes for vulnerabilities in the hardcoded versions...

Given that there has been a lot of effort in the general NPM community/ecosystem to also mitigate the potential negative impact of supply chain version tag compromise attacks, it 

We already have in place security provisions to allow us to invalidate the registration codes that are used for CI testing purposes if they ever make it into the wild. Similarly, our sensitive workflows, that leverage those registration code, can only be run for PRs based upon branches within the main repo where the relevant secrets are available.

Therefore we feel that it would be better to focus on locking down the workflows to only have read access (which you are also doing in this PR) so as to severely limit what a rogue workflow can do. This represents an overall improvement in the security of our GitHub Actions workflows, without increasing the on-going maintenance workload associated with this repo.

rtamalin (changes requested):
Per the re-opened comment, the minimal benefit of the if conditional on the feature-tests workflow job is outweighed by the negative impact of breaking local `act` testing.

Additionally the `--secret-file .env` should be added to the `.actrc` to offset the negative impact of moving the top-level env var definitions to be job step specific.

And, while locking down the actions to use hardcoded SHA references rather than tags does represent a security improvement, it comes with the downside of increasing the maintenance burden for the repo as we would now need to actively monitor those action implementations to check for CVEs and security fixes and explicitly move the SHA references as needed. And recent NPM community/ecosystem enhancements have already been adopted into the GitHub Actions ecosystem to help mitigate potential for compromised tag supply chain attacks going forward. 

As such I think a better focus would be to concentrate on locking down all of the GitHub Actions workflows to limit what a rogue action implementation could do, which is something that you have already done as part of this PR.

Next Steps:
* apply the permissions config setup to all active GitHub Actions workflows, including the ones added since this PR was originally created
* switch back to using tag references for the actions, but updated to the latest versions major version tags, matching those associated with the hardcoded references you currently have
* update the `.actrc` to add the `--secret-file .env` to address the local `act` testing breakage.
* remove the conditional `if` check on the feature-tests workflow

.github/workflows/features.yml
voving30: Fixed. Added --secret-file .env to .actrc

.github/workflows/features.yml
voving30: Updated checkout references back to the major version tags as requested

.github/workflows/features.yml
voving30: Removed the feature-test if condition as requested

voving30:
Addressed the requested changes in 3c32519. Ready for another review

rtamalin:
> Addressed the requested changes in [3c32519](https://github.com/SUSE/connect-ng/commit/3c325192ddf5feb38b7b7847f1d3c16b89bb6028). Ready for another review

Can you also update the version tags in the other workflows to match the updates you are making in the original workflows, e.g. you are still only updating the actions/checkout@v4 to ...@v7 only in the original 3 workflows that you modified.

voving30:
Updated the remaining checkout references to v7. Also cleaned up the related secret handling and act documentation

rtamalin (approved):
All tests verified as passing in a local act test environment.

Issue #374, opened 2026-04-09
InstallReleasePackage() should consider zypperFilesystemRoot


jiri-belka:
https://github.com/SUSE/connect-ng/blob/9a94d6f3e97c6ba670a3a261c9b7d489470aa5fd/internal/zypper/zypper.go#L236

I used this command, executed on SLES 16.0 - installing 15.7 into a directory for a root on NFS system:

``` shell
SUSEConnect --root /tmp/foobar -e <email> -r <code> -p SLES/15.7/x86_64 
...
-> Installing release package ...
Executing: [rpm -q SLES-release]
Output: SLES-release-16.0-160000.42.1.x86_64
```

But!

``` shell
rpm --root /tmp/foobar -qa | wc -l
0
```
vs the host itself:

``` shell
rpm -qa SLES-release
SLES-release-16.0-160000.42.1.x86_64
```

That is, clearly this is querying outside the `--root /tmp/foobar`, that is, a mistake.

jiri-belka:
Something like this? (Untested)
``` go
// return if the rpm is already installed
cmd :=[]string{"rpm"}
if zypperFilesystemRoot	!= "/"	{
        cmd = append(cmd, "--root", zypperFilesystemRoot)
}       
	
args :=[]string{"-q", identifier + "-release"}
cmd = append(cmd, args)
	
if _, err := util.Execute(cmd, nil); err == nil {
	return nil
}
```

jiri-belka:
FYI https://bugzilla.suse.com/show_bug.cgi?id=1261589

jiri-belka:
https://github.com/SUSE/connect-ng/blob/9a94d6f3e97c6ba670a3a261c9b7d489470aa5fd/internal/zypper/zypper.go#L241 should probably also check if it returns valid version.

rtamalin:
PR #377 fixed this issue

PR #377, opened 2026-04-16
Issue: InstallReleasePackage should consider zypperFilesystemRoot

chajain:
Fixes Issue #347 (InstallReleasePackage currently not  considering zypperFilesystemRoot)

### DIFF
--- a/build/ci/run-feature-tests
+++ b/build/ci/run-feature-tests
@@ -26,6 +26,6 @@ groupend
 
 group "run feature tests"
 pushd "$SOURCE"
- go test -v features/suseconnect/*
+ go test -timeout 15m -v features/suseconnect/*
 popd
 groupend
--- a/build/packaging/suseconnect-ng.changes
+++ b/build/packaging/suseconnect-ng.changes
@@ -2,6 +2,8 @@
 Mon Apr  6 15:40:58 UTC 2026 - Fergal Mc Carthy <fmccarthy@suse.com>
 
 - Update version to 1.22 (pre):
+  - InstallReleasePackage should consider zypperFilesystemRoot
+    (jsc#SCC-630).
   
 -------------------------------------------------------------------
 Wed Apr  1 16:43:26 UTC 2026 - Fergal Mc Carthy <fmccarthy@suse.com>
--- a/features/helpers/filesystem.go
+++ b/features/helpers/filesystem.go
@@ -3,18 +3,27 @@ package helpers
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/SUSE/connect-ng/internal/zypper"
 	"github.com/stretchr/testify/assert"
 )
 
 func CleanupPolutedFilesystem() {
-	fmt.Printf("[cleanup] Cleanup /etc/zypp/{credentials.d, services.d, repos.d}/*...\n")
-	removeInnerRecursive("/etc/zypp/credentials.d")
-	removeInnerRecursive("/etc/zypp/services.d")
-	removeInnerRecursive("/etc/zypp/repos.d")
+	zyppDirLst := []string{"credentials.d", "services.d", "repos.d"}
+	zyppPath := "/etc/zypp"
+	fmt.Printf("[cleanup] Cleanup %s/{%s}/*...\n", zyppPath, strings.Join(zyppDirLst, ", "))
+	zyppRoot := zypper.GetFilesystemRoot()
+	for _, zyppDir := range zyppDirLst {
+		if zypper.GetFilesystemRoot() != "/" {
+			removeInnerRecursive(filepath.Join(zyppRoot, zyppPath, zyppDir))
+		} else {
+			removeInnerRecursive(filepath.Join(zyppPath, zyppDir))
+		}
+	}
 }
 
 func removeInnerRecursive(dir string) {
@@ -46,6 +55,25 @@ func FriendlyNameToServiceName(friendlyName string) string {
 
 func FriendlyNameToCredentialsName(friendlyName string) string {
 	return FriendlyNameToServiceName(friendlyName)
+}
+
+func SetupCustomRoot(t *testing.T) string {
+	assert := assert.New(t)
+	root := t.TempDir()
+	zypper.SetFilesystemRoot(root)
+
+	err := os.MkdirAll(filepath.Join(root, "etc"), 0755)
+	assert.NoError(err)
+
+	// FIXME: Once golang 1.23 is integrated we can implement this in pure go using
+	// os.CopyFS, e.g. err := os.CopyFS("dst", os.DirFS("src"))
+	err = exec.Command("cp", "-r", "/etc/zypp", filepath.Join(root, "etc/zypp")).Run()
+	assert.NoError(err)
+
+	err = exec.Command("cp", "-r", "/etc/products.d", filepath.Join(root, "etc/products.d")).Run()
+	assert.NoError(err)
+
+	return root
 }
 
 func TryCurlrcCleanup() {
--- a/features/helpers/state.go
+++ b/features/helpers/state.go
@@ -4,11 +4,21 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+
+	"github.com/SUSE/connect-ng/internal/zypper"
 )
 
 func TrySUSEConnectCleanup() {
 	fmt.Printf("[cleanup] Try running suseconnect --cleanup...\n")
-	cmd := exec.Command("suseconnect", "--cleanup")
+	args := []string{}
+
+	// Add --root flag if custom root is set
+	if zypper.GetFilesystemRoot() != "/" {
+		args = append(args, "--root", zypper.GetFilesystemRoot())
+	}
+
+	args = append(args, "--cleanup")
+	cmd := exec.Command("suseconnect", args...)
 	err := cmd.Run()
 
 	if err != nil {
@@ -18,7 +28,15 @@ func TrySUSEConnectCleanup() {
 
 func TrySUSEConnectDeregister() {
 	fmt.Printf("[cleanup] Try running suseconnect --de-register...\n")
-	cmd := exec.Command("suseconnect", "--de-register")
+	args := []string{}
+
+	// Add --root flag if custom root is set
+	if zypper.GetFilesystemRoot() != "/" {
+		args = append(args, "--root", zypper.GetFilesystemRoot())
+	}
+
+	args = append(args, "--de-register")
+	cmd := exec.Command("suseconnect", args...)
 	err := cmd.Run()
 
 	if err != nil {
--- a/features/suseconnect/changed_root_test.go
+++ b/features/suseconnect/changed_root_test.go
@@ -2,7 +2,6 @@ package features
 
 import (
 	"os"
-	"os/exec"
 	"path/filepath"
 	"testing"
 
@@ -14,7 +13,8 @@ func TestChangedRoot(t *testing.T) {
 	assert := assert.New(t)
 	env := helpers.NewEnv(t)
 
-	root := setupCustomRoot(t)
+	root := helpers.SetupCustomRoot(t)
+	t.Cleanup(helpers.CleanupAll)
 	namespace := "test-namespace"
 
 	// No need to install rpm packages just registration
@@ -33,23 +33,6 @@ func TestChangedRoot(t *testing.T) {
 	t.Run("check credentials location", func(t *testing.T) {
 		testChangedRootCredentialsLocation(t, root)
 	})
-}
-
-func setupCustomRoot(t *testing.T) string {
-	assert := assert.New(t)
-	root := t.TempDir()
-
-	err := os.MkdirAll(filepath.Join(root, "etc"), 0755)
-	assert.NoError(err)
-
-	// FIXME: Once golang 1.23 is integrated this becomes much more nice to implement in pure go
-	err = exec.Command("cp", "-r", "/etc/zypp", filepath.Join(root, "etc/zypp")).Run()
-	assert.NoError(err)
-
-	err = exec.Command("cp", "-r", "/etc/products.d", filepath.Join(root, "etc/products.d")).Run()
-	assert.NoError(err)
-
-	return root
 }
 
 func testChangedRootConfigLocation(t *testing.T, root string, namespace string) {
--- a/features/suseconnect/install_release_pkg_test.go
+++ b/features/suseconnect/install_release_pkg_test.go
@@ -1,0 +1,159 @@
+package features
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/SUSE/connect-ng/features/helpers"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInstallPackage(t *testing.T) {
+	t.Run("install PackageHub with auto-import keys - default root", testInstallPackageHubDefaultRoot)
+	t.Run("install PackageHub with auto-import keys - custom root", testInstallPackageHubCustomRoot)
+	t.Run("skip installation if package already installed - custom root", testSkipInstallationIfInstalledCustomRoot)
+	t.Run("install release package with default root", testInstallReleasePackageDefaultRoot)
+	t.Run("install release package with custom root", testInstallReleasePackageCustomRoot)
+}
+
+func testInstallPackageHubDefaultRoot(t *testing.T) {
+	t.Cleanup(helpers.CleanupAll)
+
+	assert := assert.New(t)
+	env := helpers.NewEnv(t)
+
+	// Register the system
+	register := helpers.NewRunner(t, "suseconnect -r %s", env.REGCODE)
+	register.Run()
+	assert.Equal(0, register.ExitCode())
+
+	zypp := helpers.NewZypper(t)
+	_, version, arch := zypp.BaseProduct()
+
+	extension := "PackageHub"
+	activate := helpers.NewRunner(t, "suseconnect -p %s/%s/%s --gpg-auto-import-keys", extension, version, arch)
+	activate.Run()
+
+	if activate.ExitCode() == 0 {
+		assert.Contains(activate.Stdout(), "Activating")
+	} else {
+		// If it fails, it should not be due to signature verification
+		assert.NotContains(activate.Stderr(), "signature verification failed")
+	}
+}
+
+func testInstallPackageHubCustomRoot(t *testing.T) {
+	assert := assert.New(t)
+	env := helpers.NewEnv(t)
+
+	root := helpers.SetupCustomRoot(t)
+	t.Cleanup(helpers.CleanupAll)
+	namespace := "test-packagehub"
+
+	// Register with custom root and auto-import keys
+	register := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+	register.Run()
+
+	if register.ExitCode() == 0 {
+		assert.Contains(register.Stdout(), "Rooted at: "+root)
+		assert.Contains(register.Stdout(), "Activating")
+	} else {
+		// If it fails, it should not be due to signature verification
+		assert.NotContains(register.Stderr(), "signature verification failed")
+	}
+
+	// Verify the root flag was properly used
+	credPath := filepath.Join(root, "etc/zypp/credentials.d/SCCcredentials")
+	assert.FileExists(credPath, "Credentials should be in custom root")
+}
+
+func testSkipInstallationIfInstalledCustomRoot(t *testing.T) {
+	assert := assert.New(t)
+	env := helpers.NewEnv(t)
+
+	root := helpers.SetupCustomRoot(t)
+	t.Cleanup(helpers.CleanupAll)
+	namespace := "test-skip-install"
+
+	// First registration
+	register1 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+	register1.Run()
+	assert.Contains(register1.Stdout(), "Installing release package")
+	assert.Equal(0, register1.ExitCode())
+
+	// deregister before 2nd registration
+	deregister1 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -d", root, namespace)
+	deregister1.Run()
+
+	register2 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
+	register2.Run()
+	assert.Contains(register2.Stdout(), "Installing release package")
+	assert.Equal(0, register2.ExitCode(), "Re-registration with custom root should succeed")
+
+	// Config should exist
+	configPath := filepath.Join(root, "/etc/SUSEConnect")
+	assert.FileExists(configPath)
+}
+
+func testInstallReleasePackageDefaultRoot(t *testing.T) {
+	t.Cleanup(helpers.CleanupAll)
+
+	assert := assert.New(t)
+	env := helpers.NewEnv(t)
+
+	zypp := helpers.NewZypper(t)
+	namespace := "test-install-release-package"
+
+	extension := "sle-module-basesystem"
+	cli := helpers.NewRunner(t, "suseconnect --namespace %s -r %s", namespace, env.REGCODE)
+	cli.Run()
+
+	assert.Equal(0, cli.ExitCode())
+	assert.Contains(cli.Stdout(), "Successfully registered system")
+
+	// Verify the release package was installed by checking for the product
+	products := zypp.FetchProducts()
+	foundProduct := false
+	for _, product := range products {
+		if product.Identifier == extension {
+			foundProduct = true
+			break
+		}
+	}
+	assert.True(foundProduct, "Release package should have installed the product")
+}
+
+func testInstallReleasePackageCustomRoot(t *testing.T) {
+	assert := assert.New(t)
+	env := helpers.NewEnv(t)
+
+	// Setup custom root
+	root := helpers.SetupCustomRoot(t)
+	t.Cleanup(helpers.CleanupAll)
+	zypp := helpers.NewZypper(t)
+	namespace := "test-install-release-package"
+
+	extension := "sle-module-basesystem"
+
+	// Register with custom root
+	register := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+	register.Run()
+
+	assert.Equal(0, register.ExitCode(), "Registration with custom root should succeed")
+	assert.Contains(register.Stdout(), "Rooted at: "+root)
+	assert.Contains(register.Stdout(), "Successfully registered system")
+
+	// Verify the release package was installed by checking for the product
+	products := zypp.FetchProducts()
+	foundProduct := false
+	for _, product := range products {
+		if product.Identifier == extension {
+			foundProduct = true
+			break
+		}
+	}
+	assert.True(foundProduct, "Release package should have installed the product with custom root")
+
+	configPath := filepath.Join(root, "/etc/SUSEConnect")
+	assert.FileExists(configPath, "Config should exist in custom root")
+}
--- a/internal/zypper/zypper.go
+++ b/internal/zypper/zypper.go
@@ -237,9 +237,13 @@ func InstallReleasePackage(identifier string, autoImportRepoKeys bool) error {
 	if identifier == "" {
 		return nil
 	}
-	// return if the rpm is already installed
-	args := []string{"rpm", "-q", identifier + "-release"}
-	if _, err := util.Execute(args, nil); err == nil {
+	cmd := []string{"rpm"}
+
+	if zypperFilesystemRoot != "/" {
+		cmd = append(cmd, "--root", zypperFilesystemRoot)
+	}
+	cmd = append(cmd, "-q", identifier+"-release")
+	if _, err := util.Execute(cmd, nil); err == nil {
 		return nil
 	}
 
@@ -251,7 +255,7 @@ func InstallReleasePackage(identifier string, autoImportRepoKeys bool) error {
 		validExitCodes = append(validExitCodes, zypperInfoReposSkipped)
 	}
 
-	args = []string{"--no-refresh", "--non-interactive", "install", "--no-recommends",
+	args := []string{"--no-refresh", "--non-interactive", "install", "--no-recommends",
 		"--auto-agree-with-product-licenses", "-t", "product", identifier}
 
 	if autoImportRepoKeys {
--- a/features/helpers/cleanupall.go
+++ b/features/helpers/cleanupall.go
@@ -1,0 +1,13 @@
+package helpers
+
+import (
+	"github.com/SUSE/connect-ng/internal/zypper"
+)
+
+func CleanupAll() {
+	TrySUSEConnectDeregister()
+	TrySUSEConnectCleanup()
+	CleanupPolutedFilesystem()
+
+	zypper.SetFilesystemRoot("/")
+}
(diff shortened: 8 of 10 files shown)


features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+func testInstallReleasePackageCustomRoot(t *testing.T) {
+        assert := assert.New(t)
+        env := helpers.NewEnv(t)
+
+        // Setup custom root
+        root := setupCustomRootPath(t)
+        namespace := "test-install-release-package"
+
+        // No need to install rpm packages, just test registration
```
rtamalin: I'm not sure that the naming of this test case aligns with the actions of this test for the very reason that you are using the env var that disables the setup of the zypper repo service and thus release package installs...
As such this test case if just a duplicate of the testing done for the custom root and doesn't check for the associated release package being installed...

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func testSkipInstallationIfInstalledCustomRoot(t *testing.T) {
+        assert := assert.New(t)
+        env := helpers.NewEnv(t)
+
+        root := setupCustomRootPath(t)
+        namespace := "test-skip-install"
+
+        t.Setenv("SKIP_SERVICE_INSTALL", "true")
```
rtamalin: Similar to my earlier comment this env var setting disables the installation of the zypper repo service associated with this product which implicitly disables the install of the release package...

As such this isn't really testing for the conditions of whether a release package was installed or not, but rather just that re-registering in a custom root setup, over an existing registration, works correctly, and as such this test case probably should be moved to the custom root tests, rather than being part of this set of tests...

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,195 @@
...
+        // Config should exist
+        configPath := filepath.Join(root, "etc/SUSEConnect")
+        assert.FileExists(configPath)
+}
+
+func testInstallPackageHubDefaultRoot(t *testing.T) {
+        assert := assert.New(t)
+        env := helpers.NewEnv(t)
+
+	t.Setenv("SKIP_SERVICE_INSTALL", "true")
```
rtamalin: Indentation is out of alignment for this line.

Additionally not sure why we are using this env var setting here, for a default root setup, which would disable the setup of the zypper repo services and associated release packages?

internal/zypper/zypper.go
```diff
@@ -237,9 +237,14 @@ func InstallReleasePackage(identifier string, autoImportRepoKeys bool) error {
...
 	}
-	// return if the rpm is already installed
-	args := []string{"rpm", "-q", identifier + "-release"}
-	if _, err := util.Execute(args, nil); err == nil {
+	cmd := []string{"rpm"}
+
+	if zypperFilesystemRoot != "/" {
+		cmd = append(cmd, "--root", zypperFilesystemRoot)
+	}
+	args := []string{"-q", identifier + "-release"}
```
rtamalin: For compactness you could just specify the values inside the []string{} on this line as the extra arguments in the append call on the next line...

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func testInstallPackageHubCustomRoot(t *testing.T) {
+        assert := assert.New(t)
+        env := helpers.NewEnv(t)
+
+        root := setupCustomRootPath(t)
+        namespace := "test-packagehub"
+
+        t.Setenv("SKIP_SERVICE_INSTALL", "true")
```
rtamalin: As per other comments, not sure that setting this env var aligns with the apparent goal of testing for release package installs, given that it effectively disables the setup of zypper repo services, and this the install of the release packages as part of activating an extension.

chajain: Removed SKIP_SERVICE_INSTALL

features/suseconnect/install_release_pkg_test.go
chajain: Removed SKIP_SERVICE_INSTALL flag

features/suseconnect/install_release_pkg_test.go
chajain: Removed SKIP_SERVICE_INSTALL flag

features/suseconnect/install_release_pkg_test.go
chajain: Removed SKIP_SERVICE_INSTALL

brett060102:
Looks good to me.

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+	// First registration
+	register1 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+	register1.Run()
+	assert.Contains(register1.Stdout(), "Installing release package")
+	assert.Equal(0, register1.ExitCode())
+
+	// Second registration with same root - should not fail
+	register2 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
+	register2.Run()
+	assert.Contains(register2.Stdout(), "Activating")
```
rtamalin: The `Activating` text will always be there for an original registration or a re-registration.

To test for the scenario identified by this test's name, you should be checking that the output doesn't contain `Installing release package` with an assert.NotContains() or similar check.

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+	root := setupCustomRootPath(t)
+	namespace := "test-install-release-package"
+
+	// Register with custom root
+	register := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+	register.Run()
+
+	assert.Equal(0, register.ExitCode(), "Registration with custom root should succeed")
+	assert.Contains(register.Stdout(), "Rooted at: "+root)
+
```
rtamalin, line 64: You seem to be missing any check that the release package was installed either using the zypper construct approach checking of the base system module product as used in the default root version of this test, or by checking for the relevant text in the output.

The verification steps in both test scenarios should be equivalent, with the difference being the *root* path that anchors the file system paths...

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+	env := helpers.NewEnv(t)
+
+	root := setupCustomRootPath(t)
+	namespace := "test-packagehub"
+
+	// Register with custom root and auto-import keys
+	register := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+	register.Run()
+
+	assert.Equal(0, register.ExitCode(), "Registration with custom root and auto-import should succeed")
```
rtamalin: For the default root equivalent of this test scenario you have special case checking for ExitCode being either 0 or non-zero, with that latter alternatively checking for the signature verification failure message...

But in this test scenario for the custom root, you don't have the same checking... Does this mean that there is a fundamental difference in the suseconnect operation for the custom root scenario?

Or does it mean that the default root version of the test scenario has unnecessary special handling?

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	// verifies the special exit code handling zypperInfoReposSkipped for PackageHub
+	extension := "PackageHub"
+	activate := helpers.NewRunner(t, "suseconnect -p %s/%s/%s --gpg-auto-import-keys", extension, version, arch)
+	activate.Run()
+
+	if activate.ExitCode() == 0 {
+		assert.Contains(activate.Stdout(), "Activating")
+	}
+	// If it fails, it should not be due to signature verification
```
rtamalin: Is this special case still possible when the `--gpg-auto-import-keys` flag is used? I ask because the custom root version of this test scenario doesn't implement this special case checking...

Also given that the condition for this check is the exact opposite of the previous if check, it just be an else.

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+	assert.Contains(register.Stdout(), "Rooted at: "+root)
+
+	register = helpers.NewRunner(t, "suseconnect --root %s -d", root)
+	register.Run()
+
+	configPath := filepath.Join(root, "/etc/SUSEConnect")
+	assert.FileExists(configPath, "Config should exist in custom root")
+}
+
+func setupCustomRootPath(t *testing.T) string {
```
rtamalin: This appears to be a duplicate of the setupCustomRoot() routine from the changed_root_test.go file.

There is no need to duplicate it here, as both files are in the same package, so you should be able to use the setupCustomRoot() routine rather than duplicating it.

However, given that it is being used across multiple ..._test.go files that makes it a candidate for moving it to the helpers package under features/helpers, probably in the filesystem.go file, and then update this file, and the cheanged_root_test.go file to pull it in from the helpers package.

rtamalin (changes requested):
Some inconsistencies between the checking used for the same test cases under default vs custom root.

rtamalin:
Also given that this PR implements the suggested fixes for issue #374 it would be good to add a Fixes or Implements tag line for this issue to the PR's description.

features/suseconnect/install_release_pkg_test.go
rtamalin: Further offline discussion identifies that the `Installing release package` is still being reported for the re-registration case where the release package is already installed.

As such I would recommend checking for the same string, or both strings, in both cases.

build/packaging/suseconnect-ng.changes
```diff
@@ -1,3 +1,9 @@
+------------------------------------------------------------------
+Wed Apr 22 11:55:05 UTC 2026 - Charu Jain <cjain@suse.com>
+
+- InstallReleasePackage should consider zypperFilesystemRoot
```
rtamalin: This entry needs to be moved under the v1.22 (pre) block below, as this change will be going out as part of the v1.22 release stream.

rtamalin (changes requested):

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,157 @@
...
+        register2.Run()
+        assert.Contains(register2.Stdout(), "Installing release package")
+        assert.Equal(0, register2.ExitCode(), "Re-registration with custom root should succeed")
+
+        // Config should exist
+        configPath := filepath.Join(root, "/etc/SUSEConnect")
+        assert.FileExists(configPath)
+}
+
+func testInstallReleasePackageCustomRoot(t *testing.T) {
```
rtamalin, line 126: Minor comment not requiring a change but might be a nice to do... This test routine and the first test routine share a common theme... Perhaps move this one to be the second test so as to logically group the commonly themed test cases?

build/packaging/suseconnect-ng.changes
rtamalin: And I mean just the `- InstallReleasePackage ...` part not the whole block.

See previous release blocks for examples.

rtamalin:
Testing infrastructure and code changes look good - just need to fix the .changes log entry

features/suseconnect/install_release_pkg_test.go
chajain: Having this test as the second test was giving some failures and convoluting the env for the next test and I discussed that with @brett060102 and we decided to move it at the end which seemed to resolve the issue and tests started running smoothly. Instead, I will try moving the first test before the last

rtamalin: Tests shouldn't have side effects like that...

The individual test case should be cleaning up after themselves and should be able to run independently of each other.

If testA is relying on something testB is doing then we should either replicate whatever it is that testB does into the setup of testA, and both tests should cleanup/undo that.

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,163 @@
...
+import (
+        "path/filepath"
+        "testing"
+
+        "github.com/SUSE/connect-ng/features/helpers"
+        "github.com/stretchr/testify/assert"
+)
+
+func TestInstallPackageHubPackage(t *testing.T) {
+        t.Cleanup(helpers.CleanupPolutedFilesystem)
```
rtamalin: Note that these t.Cleanup() calls are registering the cleanup handler to be called after the Test... routine finished, so that we clean up any problem state that may linger from the test runs.

As such if you want to run multiple test... sub-case routines within the same Test run, you will need to make sure that the test... sub-case routines cleanup after themselves... E.g. call an appropriate suseconnect -d, possibly with the appropriate --root arg if using a custom root, at the end of the test... sub-case routine...

Or promote them to full Test... routines, with their own t.Cleanup() blocks at the start...

But that does beg the question as to whether we are cleaning up the custom root registrations as part of the test sub-case or case implementations?

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,163 @@
...
+                assert.Contains(register.Stdout(), "Activating")
+        } else {
+                // If it fails, it should not be due to signature verification
+                assert.NotContains(register.Stderr(), "signature verification failed")
+        }
+
+        // Verify the root flag was properly used
+        credPath := filepath.Join(root, "etc/zypp/credentials.d/SCCcredentials")
+        assert.FileExists(credPath, "Credentials should be in custom root")
+}
```
rtamalin, line 68: We probably should have a custom root registration cleanup at the end of the custom root sub-case tests, e.g. something like


```suggestion
        register := helpers.NewRunner(t, "suseconnect --root %s --namespace %s", root, namespace)
        register.Run()
}
```

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,163 @@
...
+        // Second registration with same root - should not fail
+        register2 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
+        register2.Run()
+        assert.Contains(register2.Stdout(), "Installing release package")
+        assert.Equal(0, register2.ExitCode(), "Re-registration with custom root should succeed")
+
+        // Config should exist
+        configPath := filepath.Join(root, "/etc/SUSEConnect")
+        assert.FileExists(configPath)
+}
```
rtamalin, line 96: See above comment about doing a custom root deregistration at the end of the test case...

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,163 @@
...
+        root := helpers.SetupCustomRoot(t)
+        namespace := "test-skip-install"
+
+        // First registration
+        register1 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+        register1.Run()
+        assert.Contains(register1.Stdout(), "Installing release package")
+        assert.Equal(0, register1.ExitCode())
+
+        // Second registration with same root - should not fail
```
rtamalin: Possibly update this comment to indicate that this test case will leak an SCC registration


```suggestion
        // Second registration with same root - should not fail
        // NOTE: registering without deregistering will orphan the previous registration in
        // SCC, leaving us unable to clean it up.
```

We might be able to avoid the orphaned registration if we backup the SCC credentials before performing the re-registration, and restore them again afterwards to be able to run a suseconnect -d for them... But that may not work reliably if the suseconnect command performs any actions with the older credentials that updates the system token before the re-registration is performed...

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,163 @@
...
+                }
+        }
+        assert.True(foundProduct, "Release package should have installed the product with custom root")
+
+        register = helpers.NewRunner(t, "suseconnect --root %s -d", root)
+        register.Run()
+
+        configPath := filepath.Join(root, "/etc/SUSEConnect")
+        assert.FileExists(configPath, "Config should exist in custom root")
+}
```
rtamalin, line 159: Se earlier comment about performing a custom root deregistration...

brett060102:
@chajain When I tested your branch on my system I hit a test timeout, the default wgen do ing "go test" is ten minutes with the new tests and additional clean my system was just not fast enough to complete in time.
Could you please make the following change?
```
diff --git a/build/ci/run-feature-tests b/build/ci/run-feature-tests
index b032928..495c026 100755
--- a/build/ci/run-feature-tests
+++ b/build/ci/run-feature-tests
@@ -26,6 +26,6 @@ groupend
 group "run feature tests"
 pushd "$SOURCE"
- go test -v features/suseconnect/*
+ go test -timeout 20m -v features/suseconnect/*
 popd
 groupend

```
sets the timeout to 20 minutes for the complete feature test.

features/suseconnect/install_release_pkg_test.go
brett060102: should be resolved.

brett060102: This should be resolved now. The issue was not really a side effect from the test. The problem was that was the install was being done to the change root, but the system root was being checked.

features/suseconnect/install_release_pkg_test.go
brett060102: should be happening now.

features/suseconnect/install_release_pkg_test.go
brett060102: Currently test don't have orphans. Saving the token, would not have helped in the cases we have seen so far. Where we lost the token because destroy the dir holding the token before we did a dereg. What might be good is to attempt a dreg before doing a reg.

features/suseconnect/install_release_pkg_test.go
brett060102: things should be good now

features/suseconnect/install_release_pkg_test.go
brett060102: This work so long as not using --root because temp files are removed before user set defers. for --root case cleanup needs to be called at end if each test.

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,164 @@
...
+		// If it fails, it should not be due to signature verification
+		assert.NotContains(activate.Stderr(), "signature verification failed")
+	}
+}
+
+func testInstallPackageHubCustomRoot(t *testing.T) {
+	assert := assert.New(t)
+	env := helpers.NewEnv(t)
+
+	root := helpers.SetupCustomRoot(t)
```
brett060102, line 49: If you put:
t.Cleanup(helpers.CleanupAll)
after helpers.SetupCustomRoot(t)
Things should work. Then you can remove the helpers.CleanupAll at then end.

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,164 @@
...
+	register2 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
+	register2.Run()
+	assert.Contains(register2.Stdout(), "Installing release package")
+	assert.Equal(0, register2.ExitCode(), "Re-registration with custom root should succeed")
+
+	// Config should exist
+	configPath := filepath.Join(root, "/etc/SUSEConnect")
+	assert.FileExists(configPath)
+
+	helpers.CleanupAll()
```
brett060102: helpers.CleanupAll() can be removed if you make the change I noted above.

build/ci/run-feature-tests
```diff
@@ -26,6 +26,6 @@ groupend
 
 group "run feature tests"
 pushd "$SOURCE"
- go test -v features/suseconnect/*
+ go test -timeout 20m -v features/suseconnect/*
```
rtamalin: Are we sure we want to stick with 20 minutes? I'd prefer not to be too aggressive with increasing this timeout, especially if we haven't seen it in the CI yet.

Per previous discussions on this topic, there are limits on how many simultaneous GitHub runners we can have active in our organisation, so I'd prefer us to err on the side of being good neighbours.


```suggestion
 go test -timeout 15m -v features/suseconnect/*
```

features/helpers/filesystem.go
```diff
@@ -3,18 +3,26 @@ package helpers
...
-	removeInnerRecursive("/etc/zypp/repos.d")
+	if zypper.GetFilesystemRoot() != "/" {
+		removeInnerRecursive(filepath.Join(zypper.GetFilesystemRoot(), "/etc/zypp/credentials.d"))
+		removeInnerRecursive(filepath.Join(zypper.GetFilesystemRoot(), "/etc/zypp/services.d"))
+		removeInnerRecursive(filepath.Join(zypper.GetFilesystemRoot(), "/etc/zypp/repos.d"))
+	} else {
+		removeInnerRecursive("/etc/zypp/credentials.d")
+		removeInnerRecursive("/etc/zypp/services.d")
+		removeInnerRecursive("/etc/zypp/repos.d")
+	}
```
rtamalin: You could just simplify this if/etse to retrieving the zypper file system root value, and then doing the recursive remove for the filepath.Join, e.g. something like:


```suggestion
	zyppRoot := zypper.GetFilesystemRoot()
	for _, zyppDir := range []string{"credentials.d", "services.d", "repos.d"}{
	    removeInnerRecursive(filepath.Join(zyppRoot, "/etc/zypp", zyppDir))
	}
```

You can take this further and define the "/etc/zypp" and subdirectory lists in variables before the fmt.Printf() and use them in the construction of the message...

features/helpers/filesystem.go
```diff
@@ -48,6 +56,24 @@ func FriendlyNameToCredentialsName(friendlyName string) string {
...
 
+func SetupCustomRoot(t *testing.T) string {
+	assert := assert.New(t)
+	root := t.TempDir()
+	zypper.SetFilesystemRoot(root)
+
+	err := os.MkdirAll(filepath.Join(root, "etc"), 0755)
+	assert.NoError(err)
+
+	// FIXME: Once golang 1.23 is integrated this becomes much more nice to implement in pure go
```
rtamalin: We should expand this comment with a note about what is referred to here as I keep having to lookup the specific feature...


```suggestion
	// FIXME: Once golang 1.23 is integrated we can implement this in pure go using
	// os.CopyFS, e.g. err := os.CopyFS("dst", os.DirFS("src"))
```

features/helpers/filesystem.go
```diff
@@ -3,18 +3,26 @@ package helpers
...
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/SUSE/connect-ng/internal/zypper"
 	"github.com/stretchr/testify/assert"
 )
 
 func CleanupPolutedFilesystem() {
```
rtamalin, line 15: Since these cleanups are zypper specific we probably should make this a receiver method for the Zypper type in `features/helpers/zypper.go`.

features/helpers/filesystem.go
```diff
@@ -3,18 +3,26 @@ package helpers
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/SUSE/connect-ng/internal/zypper"
```
rtamalin, line 11: Note that we already have a zypper.go under features/helpers/ that imports the internal/zypper package... So perhaps we should be extending it with the desired functionality, rather than repeating it here?

features/helpers/filesystem.go
rtamalin: And yeah I know I was the one who suggested putting it in this file. ;-)

features/helpers/filesystem.go
```diff
@@ -48,6 +56,24 @@ func FriendlyNameToCredentialsName(friendlyName string) string {
 	return FriendlyNameToServiceName(friendlyName)
 }
 
+func SetupCustomRoot(t *testing.T) string {
```
rtamalin, line 60: Similar to above, since this is pretty zypper specific we probably should make it a receiver method for the Zypper type in features/helpers/zypper.go

features/helpers/state.go
```diff
@@ -4,11 +4,21 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+
+	"github.com/SUSE/connect-ng/internal/zypper"
```
rtamalin, line 8: Potentially we should be trying to use the feature/helpers/zypper.go capabilities rather than the pulling in internal/zypper...

features/suseconnect/changed_root_test.go
```diff
@@ -14,7 +13,8 @@ func TestChangedRoot(t *testing.T) {
 	assert := assert.New(t)
 	env := helpers.NewEnv(t)
 
-	root := setupCustomRoot(t)
+	root := helpers.SetupCustomRoot(t)
```
rtamalin, line 16: @brett060102 I think this is what means that this PR fixes the issue that your #369 PR is up to fix... We can close that one out once this one lands...

features/suseconnect/install_release_pkg_test.go
rtamalin: Looks like this is addressed now?

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,161 @@
...
+	register1.Run()
+	assert.Contains(register1.Stdout(), "Installing release package")
+	assert.Equal(0, register1.ExitCode())
+
+	// Second registration with same root - should not fail
+	// NOTE: registering without deregistering will orphan the previous registration in
+	// SCC, leaving us unable to clean it up.
+	deregister1 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -d", root, namespace)
+	deregister1.Run()
+
```
rtamalin, line 87: @brett060102 and I were discussing this yesterday and if we backup the credentials before doing a deregister, and then restore them after we can avoid the leaked orphan registration, and still more accurately simulate the duplicate registration scenario...

Right now, because you are deregistering before doing the second registration, we may not be fully exercising this scenario...

features/suseconnect/install_release_pkg_test.go
```diff
@@ -0,0 +1,161 @@
...
+	namespace := "test-skip-install"
+
+	// First registration
+	register1 := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s --gpg-auto-import-keys", root, namespace, env.REGCODE)
+	register1.Run()
+	assert.Contains(register1.Stdout(), "Installing release package")
+	assert.Equal(0, register1.ExitCode())
+
+	// Second registration with same root - should not fail
+	// NOTE: registering without deregistering will orphan the previous registration in
```
rtamalin: This comment is not necessary if we have the deregistration before the reregistration

features/helpers/state.go
rtamalin: Thinking about this a bit more let's defer working on this sort of re-organising until after we get this PR in. Let's not feature creep this PR too much ;-)

features/helpers/filesystem.go
rtamalin: Thinking about this a bit more let's defer working on this sort of re-organising until after we get this PR in. Let's not feature creep this PR too much ;-)

features/helpers/filesystem.go
rtamalin: Thinking about this a bit more let's defer working on this sort of re-organising until after we get this PR in. Let's not feature creep this PR too much ;-)

features/helpers/filesystem.go
rtamalin: Thinking about this a bit more let's defer working on this sort of re-organising until after we get this PR in. Let's not feature creep this PR too much ;-)

rtamalin:
@chajain It looks good now... I started to suggest we could amalgamate the new zypper related functionality in the feature test into the features/helpers/zypper.go Zypper type as receiver methods, but let's defer that to a follow on PR, and get this one in.

I have made a minor code cleanup suggestion if you want to accept that and we can then get this PR in ;-)

features/suseconnect/changed_root_test.go
brett060102: Yes, once this code is registration leak free.

features/helpers/filesystem.go
```diff
@@ -48,6 +57,25 @@ func FriendlyNameToCredentialsName(friendlyName string) string {
...
+	assert := assert.New(t)
+	root := t.TempDir()
+	zypper.SetFilesystemRoot(root)
+
+	err := os.MkdirAll(filepath.Join(root, "etc"), 0755)
+	assert.NoError(err)
+
+        // FIXME: Once golang 1.23 is integrated we can implement this in pure go using
```
rtamalin:
```suggestion
    // FIXME: Once golang 1.23 is integrated we can implement this in pure go using
```

Extra indent got added when applying the previous suggestion...

rtamalin (approved):
Looking good now - I applied a minor whitespace tweak

brett060102:
@chajain there is still a registration leak. It can be fixed with the following change:

```
diff --git a/features/suseconnect/changed_root_test.go b/features/suseconnect/changed_root_test.go
index 5da159e..6e75969 100644
--- a/features/suseconnect/changed_root_test.go
+++ b/features/suseconnect/changed_root_test.go
@@ -13,6 +13,13 @@ func TestChangedRoot(t *testing.T) {
        assert := assert.New(t)
        env := helpers.NewEnv(t)

+       // Need to register and de-register before creating
+       // --root target or de-register using --root will fail
+       cli := helpers.NewRunner(t, "suseconnect -r %s", env.REGCODE)
+       cli.Run()
+       cli = helpers.NewRunner(t, "suseconnect -d")
+       cli.Run()
+
        root := helpers.SetupCustomRoot(t)
        t.Cleanup(helpers.CleanupAll)
        namespace := "test-namespace"
@@ -20,9 +27,8 @@ func TestChangedRoot(t *testing.T) {
        // No need to install rpm packages just registration
        t.Setenv("SKIP_SERVICE_INSTALL", "true")

-       cli := helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
+       cli = helpers.NewRunner(t, "suseconnect --root %s --namespace %s -r %s", root, namespace, env.REGCODE)
        cli.Run()
-
        assert.Equal(0, cli.ExitCode())
        assert.Contains(cli.Stdout(), "Rooted at: "+root)

@@ -33,6 +39,7 @@ func TestChangedRoot(t *testing.T) {
        t.Run("check credentials location", func(t *testing.T) {
                testChangedRootCredentialsLocation(t, root)
        })
+
 }
```
if a reg and de-reg has not been done before creating the alternate root. the de-reg using the alternate root will fail because the python extension is in use.



Also can you please change the timeout to 20min, my system is failing with 15, could be poor network performance in talking to the scc.  At some point we can plumb a timeout setting in the Makefile,

rtamalin:
> Also can you please change the timeout to 20min, my system is failing with 15, could be poor network performance in talking to the scc. At some point we can plumb a timeout setting in the Makefile,

@brett060102 I'd recommend digging into why your test environment is seeing the need for an increased timeout before we increase the timeout further, given that you are the only one seeing the issue.

rtamalin:
> @chajain there is still a registration leak. It can be fixed with the following change:
> 
> ```
> diff --git a/features/suseconnect/changed_root_test.go b/features/suseconnect/changed_root_test.go

@brett060102 That is the fundamental fix from your PR #369 that we identified for the leak issue related to running the test scenarios in a BCI Golang images vs BCI Base image.

Perhaps we should let @chajain get this PR in, and then finish out fixing any remaining registration leak issues in PR #369

brett060102 (approved):
Approved, we can deal with leak issue separately.

features/helpers/filesystem.go
rtamalin: Logged SCC-709 to track the cleanup work.

features/helpers/filesystem.go
rtamalin: Logged SCC-709 to track the cleanup work.

features/helpers/filesystem.go
rtamalin: Logged SCC-709 to track the cleanup work.

features/helpers/state.go
rtamalin: Logged SCC-709 to track the cleanup work.

PR #378, opened 2026-04-21
Replaces config file parser

gbuenodevsuse:
## YAML Configuration Parser

  Replaces the custom line-by-line configuration parser with native YAML unmarshaling, enabling better structure support and future extensibility of the SUSEConnect configuration file format.

### Key Changes

  - YAML parser: Uses go.yaml.in/yaml/v3 with struct tags instead of manual line parsing
  - Nested structures: Supports collectors: section and future extensions


internal/connect/config.go
```diff
@@ -108,61 +106,37 @@ func ReadFromConfiguration(path string) (*Options, error) {
...
 	if f, err := os.Open(path); err == nil {
 		defer f.Close()
-		return parseFromConfiguration(f, cfg)
+
+		content, err := io.ReadAll(f)
+		if err != nil {
+			return nil, err
+		}
+
+		cfg.Path = path
```
rtamalin: This line is redundant as you already assign path to cfg.path on line 104.


```suggestion
```

internal/connect/config.go
```diff
@@ -108,61 +106,37 @@ func ReadFromConfiguration(path string) (*Options, error) {
...
+		cfg, err = parseConfiguration(content, cfg)
+		if err != nil {
+			return nil, err
+		}
+
+		return cfg, nil
```
rtamalin: Similarly this line is redundant as you will fall through to the return statement at the end of the routine.


```suggestion
```

internal/connect/config.go
```diff
@@ -1,16 +1,14 @@
...
 	"bytes"
 	"fmt"
 	"io"
 	"os"
-	"strconv"
-	"strings"
 
 	"github.com/SUSE/connect-ng/internal/util"
 	"github.com/SUSE/connect-ng/pkg/registration"
+	"gopkg.in/yaml.v3"
```
rtamalin, line 11: I just checked and I see that the maintainer of this project has actually put up a note saying the project is no longer being maintained, and archived just over a year ago, not long after I was doing my research into available GO YAML packages for use in SUSE/telemetry & SUSE/telemetry-client.

We probably should be using https://github.com/yaml/go-yaml/ active V4 release, aka `go.yaml.in/yaml/v4` which is being actively maintained by the YAML Org, with an MIT license so there shouldn't be any license compatibility issues, and is also a pure Go implementation with no dependencies.

There are also mechanisms to allow us to mitigate the stretchr/testify dependency on gopkg.in/yaml.v3, see https://pkg.go.dev/github.com/stretchr/testify/assert/yaml and https://github.com/stretchr/testify/pull/1579 though that will still leave the indirect reference to the unmaintained package in our go.mod, and stretchr/testify's objx also uses this...
Or we could use a simple replacement override directive in the go.mod, per https://github.com/stretchr/testify/issues/1724#issuecomment-3244537229 if we would prefer not to have any use of/reference to the old unmaintained version in our repo...

internal/connect/config.go
```diff
@@ -108,61 +106,37 @@ func ReadFromConfiguration(path string) (*Options, error) {
...
-			return nil, fmt.Errorf("cannot parse line \"%s\" from %s", line, cfg.Path)
-		}
+// parseConfiguration parses the YAML configuration file and populates the Options struct
+func parseConfiguration(content []byte, cfg *Options) (*Options, error) {
+	if len(content) == 0 {
+		return cfg, nil
+	}
+
+	decoder := yaml.NewDecoder(bytes.NewReader(content))
+	decoder.KnownFields(true)
```
rtamalin: Just something to be aware of when using the decoder.KnownFields()...

If there are any fields in the input document document that are not recognized/mapped to fields in the target structure, an error will be flagged decoding it.

This could have implications for rollback/downgrade scenarios, where a config file may have been updated to include a new field/entry by a newer/development version of suseconnect to include a new field, and if you attempt to use the config file with an older version of suseconnect afterwards, you would see an error about the unrecognized field.


```suggestion
    // Fail if input contains unrecognized fields
    // NOTE: may require external config modifications to remove new fields
    // when downgrading to an older version after a failed upgrade.
	decoder.KnownFields(true)
```

internal/connect/config_test.go
```diff
@@ -80,3 +56,143 @@ func TestSaveLoad(t *testing.T) {
...
+language: de_DE.UTF-8
+namespace: test-namespace
+email: user@example.com
+auto_agree_with_licenses: true
+enable_system_uptime_tracking: true
+no_zypper_refs: true`
+
+	opts := DefaultOptions()
+	result, err := parseConfiguration([]byte(config), opts)
+	if err != nil {
```
rtamalin: Using stretchr/testify/assert, you can used assert.NotError() for this test

internal/connect/config_test.go
```diff
@@ -80,3 +56,143 @@ func TestSaveLoad(t *testing.T) {
...
+		t.Fatalf("Unexpected error: %v", err)
+	}
+
+	if result.BaseURL != "https://example.com" {
```
rtamalin: Using stretchr/testify/assert, you could defined an expected opts structure, and a single assert.Equal() check that the resultant opts equals the expected opts.

Would likely be simpler to maintain in the long run as well... Adding support for testing new opts would just be a matter of adding the new line to the input config data and the expected value, and not require implementing an extra if check.

internal/connect/config_test.go
```diff
@@ -80,3 +56,143 @@ func TestSaveLoad(t *testing.T) {
 		t.Errorf("got %+v, expected %+v", c2, c1)
 	}
 }
+
+func TestMinimalAndNonExistingConfiguration(t *testing.T) {
+	// Test empty file - should use defaults
+	emptyContent := []byte("")
+	opts := DefaultOptions()
+	result, err := parseConfiguration(emptyContent, opts)
+	if err != nil {
```
rtamalin: Consider using stretchr/testify's assert and require to simplify the expression of the check conditions, e.g. require.NotError() for the "not an error" check and assert.Equal() checks for the individual fields, or even just an assert.Equal() against another DefaultOptions() result...

internal/connect/config_test.go
```diff
@@ -80,3 +56,143 @@ func TestSaveLoad(t *testing.T) {
...
+	}
+	if result.Insecure != defaultInsecure {
+		t.Errorf("Expected default Insecure %v, got %v", defaultInsecure, result.Insecure)
+	}
+
+	// Test minimal YAML file - should use defaults
+	minimalContent := []byte("---\n")
+	opts2 := DefaultOptions()
+	result2, err := parseConfiguration(minimalContent, opts2)
+	if err != nil {
```
rtamalin: See above comment about stretchr/testify usage.

internal/connect/config_test.go
```diff
@@ -80,3 +56,143 @@ func TestSaveLoad(t *testing.T) {
...
+		t.Errorf("Expected default URL %s, got %s", defaultBaseURL, result2.BaseURL)
+	}
+	if result2.Insecure != defaultInsecure {
+		t.Errorf("Expected default Insecure %v, got %v", defaultInsecure, result2.Insecure)
+	}
+
+	// Test non-existing file - should use defaults
+	nonExistentPath := filepath.Join(t.TempDir(), "does_not_exist")
+	result3, err := ReadFromConfiguration(nonExistentPath)
+	if err != nil {
```
rtamalin: See above comment about stretchr/testify usage.

internal/connect/config_test.go
```diff
@@ -80,3 +56,143 @@ func TestSaveLoad(t *testing.T) {
...
+invalid_option: test
+language: en_US`,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			opts := DefaultOptions()
+			_, err := parseConfiguration([]byte(tt.config), opts)
+			if err == nil {
```
rtamalin: See earlier comments about stretchr/testify usage.

internal/connect/config.go
gbuenodevsuse: That's a relevant catch. Should we just remove it instead?

felixsch (approved):
lgtm 👍 

- Checked the changes and if I want something to suggest
- Build SUSEConnect and checked registration works when registering against RMT which writes the configuration file
- Checked that it does not break with older configuration files
- Checked how `main` handles broken configuration and how the new implementation does:
```
67dfc1ad368c:/connect # cat /etc/SUSEConnect 
---
BROKEN
url: http://rmt.scc.suse.de
insecure: false
auto_agree_with_licenses: false
enable_system_uptime_tracking: false

# main branch
67dfc1ad368c:/connect # ./out/suseconnect --status         
[{"identifier":"SLES","version":"15.7","arch":"x86_64","status":"Registered"},{"identifier":"sle-module-basesystem","version":"15.7","arch":"x86_64","status":"Registered"}]

# feat/new-yaml-parser branch
67dfc1ad368c:/connect # ./out/suseconnect --status
SUSEConnect error: error parsing configuration: yaml: line 3: mapping values are not allowed in this context
```
From my perspective this is fine, the line by line parsing was of course way more permissive.

**As always, if you think I missed something I should test but haven't, please let me know, then I take a look!**:rocket:

cheers,

rtamalin (approved):
LGTM

PR #379, opened 2026-04-29
Add a new RKE2 & K3s detection collector

rtamalin:
The new collector checks for the existence of RKE2 and K3s server and agent services being enabled in the local systemd environment and will generate a "kubernetes_provider" entry that will be included in the system hardware info blob.

The format of the "kubernetes_provider" entry is as follows:

    "kubernetes_provider": {
      "type": "<solution_type>",
      "role": "<node_role>",
      "version": "<version>"
    }

A "kubernetes_provider" entry will be generated only when a single provider service is detected, as a system should only ever be a member of a single kubernetes cluster, and should only ever have a single node role assigned to it within that cluster.

If multiple kubernetes providers are detected then a debug message will be printed reporting that multiple provideres were detected but no "kubernetes_provider" entry will be generated.

To assist with collecting this information a SystemdClient interface has been defined wthin the internal/util package. The SystemdClient interface currently supports the following methods, though more can be added as needed:

  * ListUnitsByPattern List defined units similar to systemctl list-units. Requires systemd version 230+.
  * ListUnitFilesByPattern List defined unit files similar to systemctl list-units-files. Requires systemd version 230+.
  * GetExecStart Retrieve the ExecStart property value for a given unit file.
  * GetSystemdVersion Retrieve the Version property value for the target systemd.
  * GetUnitFileState Query the state (enabled/disabled/masked) for a given unit file.
  * Close Close underly systemd client connection and release resources.

Older versions of systemd, such as that found in SLE 12 SP5, do not support the List...ByPatterns methods, so we have to fall back to retrieving all of the Units or UnitFiles and use filepath.Match() pattern matching to exclude based upon the specified patterns.

Two implementations of the SystemdClient interface are provided:

  * DbusSystemdClient Leverages github.com/godbus/dbus/v5 to provide a native Golang D-Bus implementation that is used to query the local systemd.
  * MockSystemdClient A mock implementation that unit tests can leverage to test the code.

The DbusSystemdClient retrieves the systemd Version as part of the client initialisation and caches it to assist in determining if a method is available from the target systemd or not.

The MockSystemdClient can be initialised with a specific systemd Version value and similarly uses it to determine if methods should failed with the SystemdMethodNotAvailable error.

Implement unit tests leveraging the MockSystemdClient to exercise as much of the new collector code as possible.

Add optional feature tests (disabled by default), that can be used to validate correct operation of the kubernetes provider detection collector. These feature tests an environment with a working systemd ecosystem, meaning either a VM or a specially crafted containerised environment.

Update features/README.md with instructions on how to set up a viable containerised environment to be able to run the new feature tests.

Implements: TEL-317, TEL-359

### DIFF
--- a/features/README.md
+++ b/features/README.md
@@ -40,3 +40,67 @@ Also, feature tests modify the existing filesystem by adding/removing certain
 configuration files. Hence, make sure to run this into a containerized scenario
 if you don't want unexpected surprises. That's why `make feature-tests` runs in
 a container, and why the above example does it too.
+
+### Special Notes for Kubernetes Provider Info Feature Tests
+
+The kubernetes provider info feature tests require a valid functional systemd
+environment to work and customise that environment to add sumulated kubernetes
+provider services.
+
+As such they will not work inside the normal containerised feature testing
+environment, and require either a VM or customised container image with
+systemd installed and started. See below for details on how to do that.
+
+#### Environment variables used to manage kubernetes provider info tests
+
+As such these feasture tests will be skipped by default unless the required
+environment variable `KUBERNETES_PROVIDER_TESTS_ENABLED` is set in the
+environment.
+
+Additionally setting the `KUBERNETES_PROVIDER_TESTS_NO_CLEANUP` environment
+variable will disable the cleanup of the system modifications which can be
+helpful for adhoc testings and debugging.
+
+#### Starting a container with systemd enabled
+
+The recommended approach for this is to use `podman` rather than `docker` as
+`podman` is more friendly to systemd.
+
+You will need a customised container image which has systemd installed which
+can be accomplished using a Dockerfile like the following:
+
+```dockerfile
+FROM registry.suse.com/bci/golang:1.24-openssl
+
+# Install systemd and make
+RUN zypper -n install \
+    systemd \
+    dmidecode \
+    make \
+    jq \
+    && \
+    zypper clean --all
+
+# Set up systemd as the default entrypoint
+# Note: Requires running with privileged mode or specific systemd volumes in podman
+CMD ["/usr/lib/systemd/systemd"]
+```
+
+The kubernetes provider info tests can now be run as follows:
+
+* build a testing image using the above Dockerfile
+* start a named background container using that image with systemd enabled and repo dir mounted as /connect
+* exec into the container to run a bash shell
+* build the suseconnect and symlink it into the accessible path
+* run the kubernetes provider tests with K8S_PROVIDER_TESTS_ENABLED=true set in the environment
+
+```bash
+$ podman build -t k8s-provider-tester -F /tmp/Dockerfile.k8s-tester
+$ podman run -d --name k8s-testing --systemd=true --env-file .env -v $(pwd):/connect k8s-provider-tester
+$ podman exec -it k8s-testing /bin/bash
+> git config --global --add safe.directory /connect
+> cd /connect
+> make build
+> ln -s /connect/out/suseconnect /usr/local/bin/suseconnect
+> env K8S_PROVIDER_TESTS_ENABLED=true go test -v features/suseconnect/kubernetes_provider_info_test.go
+```
--- a/internal/collectors/kubernetes_provider.go
+++ b/internal/collectors/kubernetes_provider.go
@@ -1,0 +1,254 @@
+package collectors
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"path/filepath"
+	"regexp"
+	"slices"
+
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+const (
+	// recognised kubernetes providers
+	RKE2_PROVIDER = "rke2"
+	K3S_PROVIDER  = "k3s"
+
+	// recognised kubernetes provider node roles
+	KUBERNETES_SERVER = "server"
+	KUBERNETES_AGENT  = "agent"
+)
+
+// Custom Errors
+var (
+	KubernetesProviderNotRecognised    = errors.New("kubernetes provider not recognised")
+	KubernetesRoleNotRecognised        = errors.New("kubernetes role not recognised")
+	KubernetesMultipleProvidersEnabled = errors.New("multiple kubernetes providers enabled")
+)
+
+var (
+	//
+	// regexp matchers
+	//
+
+	// match lines with the format "<app> version <version> (<hash>)" and extract
+	// <version> and <hash> fields
+	versionMatcher = regexp.MustCompile(`^\S+\s+version\s+(\S+)\s+\(([[:xdigit:]]+)\)$`)
+
+	//
+	// kubernetes providers and roles that this collector recognises and reports
+	//
+	recognisedKubernetesProviders = []string{
+		RKE2_PROVIDER,
+		K3S_PROVIDER,
+	}
+	recognisedKubernetesRoles = []string{
+		KUBERNETES_SERVER,
+		KUBERNETES_AGENT,
+	}
+)
+
+type K8S struct{}
+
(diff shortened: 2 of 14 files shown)


rtamalin:
CI Tests have passed, ready for review

internal/collectors/kubernetes_provider.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+
+		// fail back to ListUnits, and ensure name filtering is completed below
+		filterNames = true
+		units, err = systemdClient.ListUnits()
+		if err != nil {
+			util.Debug.Printf("systemdClient.ListUnits() failed: %s\n", err)
+			return nil, err
+		}
+	}
```
felixsch, line 131: I checked, you are right the `ListUnitsByPattern` is not available in really old SUSE distributed versions of systemd. Given that there is now the fallback: What do you think of dropping the `ListUnitsByPattern` completely and only rely on `ListUnits` since the implementation does iterate through the whole list anyway?

features/README.md
```diff
@@ -40,3 +40,67 @@ Also, feature tests modify the existing filesystem by adding/removing certain
...
+The kubernetes provider info tests can now be run as follows:
+
+* build a testing image using the above Dockerfile
+* start a named background container using that image with systemd enabled and repo dir mounted as /connect
+* exec into the container to run a bash shell
+* build the suseconnect and symlink it into the accessible path
+* run the kubernetes provider tests with K8S_PROVIDER_TESTS_ENABLED=true set in the environment
+
+```bash
+$ podman build -t k8s-provider-tester -F /tmp/Dockerfile.k8s-tester
```
felixsch, line 98: When trying this I get:
```
Error: unknown shorthand flag: 'F' in -F
See 'podman build --help'
```
Looks like a typo to me:

```suggestion
$ podman build -t k8s-provider-tester -f /tmp/Dockerfile.k8s-tester
```

felixsch (approved):
lgtm 👍 

I added a question and a nitpick. In case you change something, dismiss the review from me and I will have another look ASAP.

- Checked the changes for suggestions I have and added them
- See the package building correctly
- Run the collector both on k3s agent and k3s server and checked that the output is valid on a default k3s installation

```json
# k3s server
{
  "arch": "x86_64",
  "cpus": 4,
  "hostname": "honing-1",
  "kubernetes_provider": {
    "role": "server",
    "type": "k3s",
    "version": "v1.33.3+k3s1"
  }
}
# k3s agent
{
  "arch": "x86_64",
  "cpus": 4,
  "hostname": "honing-3",
  "kubernetes_provider": {
    "role": "agent",
    "type": "k3s",
    "version": "v1.33.3+k3s1"
  }
}
```
- Tried the documented feature test approach to see the testing in action
```bash
cb8b6e3afa88:/connect # env K8S_PROVIDER_TESTS_ENABLED=true go test -v features/suseconnect/kubernetes_provider_info_test.go
=== RUN   TestKubernetesProviderInfo
=== RUN   TestKubernetesProviderInfo/Detects_RKE2_Server
=== RUN   TestKubernetesProviderInfo/Detects_RKE2_Agent
=== RUN   TestKubernetesProviderInfo/Detects_K3s_Server
=== RUN   TestKubernetesProviderInfo/Detects_K3s_Agent
=== RUN   TestKubernetesProviderInfo/Detects_K3s_Custom_Server
=== RUN   TestKubernetesProviderInfo/Ignores_dummy_roles
=== RUN   TestKubernetesProviderInfo/Ignores_invalid_execStart_arguments
=== RUN   TestKubernetesProviderInfo/Logs_expected_debug_message_when_no_providers_detected
=== RUN   TestKubernetesProviderInfo/Logs_expected_debug_message_when_multiple_providers_detected
=== RUN   TestKubernetesProviderInfo/Successfully_Skips_Disabled_Services
--- PASS: TestKubernetesProviderInfo (5.34s)
    --- PASS: TestKubernetesProviderInfo/Detects_RKE2_Server (0.46s)
    --- PASS: TestKubernetesProviderInfo/Detects_RKE2_Agent (0.32s)
    --- PASS: TestKubernetesProviderInfo/Detects_K3s_Server (0.30s)
    --- PASS: TestKubernetesProviderInfo/Detects_K3s_Agent (0.29s)
    --- PASS: TestKubernetesProviderInfo/Detects_K3s_Custom_Server (0.27s)
    --- PASS: TestKubernetesProviderInfo/Ignores_dummy_roles (0.51s)
    --- PASS: TestKubernetesProviderInfo/Ignores_invalid_execStart_arguments (0.81s)
    --- PASS: TestKubernetesProviderInfo/Logs_expected_debug_message_when_no_providers_detected (0.03s)
    --- PASS: TestKubernetesProviderInfo/Logs_expected_debug_message_when_multiple_providers_detected (0.53s)
    --- PASS: TestKubernetesProviderInfo/Successfully_Skips_Disabled_Services (0.35s)
PASS
ok  	command-line-arguments	5.34
```
- Run the collector against the a standard system to see it is not breaking if nothing is detected.
-  Checked the changes for suggestions I have

cheers,

internal/collectors/kubernetes_provider.go
rtamalin: So I implemented it this way so that we will attempt to use the more efficient ListUnitsByPatterns() approach first, and fall back to the less efficient ListUnits().

The efficiency difference is to do with the response size; if ListUnitsByPatterns() is available we will only get back the small number of units matching the specified patterns, whereas if we fall back to ListUnits() we have to process every active unit in the target Systemd.

I can add a comment to that effect.

features/README.md
rtamalin: That is likely a typo on my part ;-) Thanks for spotting it.

internal/collectors/kubernetes_provider.go
```diff
@@ -0,0 +1,254 @@
...
+		"version": kSvcs[0].Version,
+	}
+
+	return provider, nil
+}
+
+func getKubernetesServices(systemdClient util.SystemdClient) ([]*KubernetesService, error) {
+	kSvcs := []*KubernetesService{}
+
+	// first try to use ListUnitsByPatterns, failing back to ListUnits if not available
```
rtamalin, line 114:
```suggestion
	// first try to use ListUnitsByPatterns, failing back to ListUnits if not available
	// which is less efficient as the response will include every active unit from
	// the target systemd, which will have to be filtered by name below.
```

internal/collectors/kubernetes_provider.go
rtamalin: I'm updated the earlier comment to highlight the efficiency difference.

felixsch: Do you think this optimization is worth keeping the extra implementation? There would be also `ListUnitsFiltered` which is available also on old systemd versions (checked https://updates.ext.suse.de/SUSE/Updates/SLE-SERVER/12-SP2-ESPOS/x86_64/update/src/systemd-228-150.46.1.src.rpm)
But I haven't checked how it exactly works.

Fine by me if you think this optimization is worth keeping the implementation. Then mark it as resolved :)

rtamalin: IIRC the ListUnitsFiltered() filtering is related to states not names.

And while yes we could just use the ListUnits() approach and leave it at that, once we no longer have to support SLE 12 SP5 (LTSS Core ends October 2030) we can drop the ListUnits() failback handling at that point and just focus on using the more efficient ListUnitsByPatterns().

This way the newer systems that we will be continuing to support long term are already using the more efficient ListUnitsByPatterns(), with the version check being a minor overhead.

felixsch: Sounds good! All good then for me

rtamalin:
Rebased to address conflicts

PR #381, opened 2026-05-04
Enable unit test coverage reporting

rtamalin:
Update the Makefile to add support for generating coverage data when running the unit tests. This coverage data will be saved under the coverage directory.

Add additional coverage related targets to the Makefile that report on the coverage data from the most recent unit test run.

Update the README.md file's section on testing to outline the testing targets that are available in the Makefile and also add a section on coverage collection and reporting.

Add git ignorance of the coverage directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,16 +32,23 @@ This will create a `out/suseconnect` binary on the host.
 
 ### Testing
 
+There are three test suites available to run:
+  * unit tests
+  * feature tests
+  * YaST2 registration tests
+
+#### Unit Tests
+
 You can run all unit tests by running `make test`. If you then want to run unit
 tests for a specific package, you can simply run it as you would do for any Go
 project, for example: `go test ./internal/collectors/`.
+
+#### Feature Tests
 
 For feature tests you first need to create an `.env` file in the root directory
 of the project with the following contents:
 
 ``` sh
-BETA_VALID_REGCODE="<regcode>"
-BETA_NOT_ACTIVATED_REGCODE="<regcode>"
 VALID_REGCODE="<regcode>"
 EXPIRED_REGCODE="<regcode>"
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,3 +58,40 @@ These values can be picked up from Glue's production environment. Once that is
...
+or set it in your environment.
+
+By default the `make test` will report the percentage of statements covered
+on a per package basis as they are tested (similar to the `coverage-percent`
+target described below), and will then generate the same detailed coverage
+report as the `coverage-func` target described below.
+
+**NOTE**: Support for generating coverage testing for the feature and YaST2
+registratoon tests is under development and will be available at a later date.
+
```
felixsch, line 85: Do we really need this comment? We can track it internally but for external contributors this seems a bit overkill, no?

felixsch (approved):
lgtm 👍 

- Checked the configuration changes
- Run the tests with and without coverage to see it working
- Had my fun looking at the Makefile functions haha

cheers,

README.md
rtamalin: I added it so that it explains why we are currently limiting coverage reporting to unit tests. Happy to remove if you feel it shouldn't be there.

rtamalin:
>     * Had my fun looking at the Makefile functions haha
I dislike repetition of almost the same pattern of code, and was exploring using a multi-target shared recipe, but the associated block of rules was somewhat bloated and the function approach is more compact and easier to read... ;-)

README.md
felixsch: I would remove it and create an task internally. Up to you!

rtamalin: There is already an open internal task for the coverage work.

I'll leave it for now and update/remove the note as the feature test and yast support enhancements land.

PR #351, opened 2026-02-24
MCP server POC for suseconnect

digitaltom:
https://github.com/user-attachments/assets/9c420986-e659-4371-bb1a-7c1ab6be18d7

Jira: https://jira.suse.com/browse/SCC-536

The MCP server can get used as stdio by pointing to the suseconnect-mcp executable, or get run as an http server. 
In the SLES agentic framework, the systems MCP servers will run behind a [proxy](https://confluence.suse.com/pages/viewpage.action?pageId=2060091640&spaceKey=SLAA&title=WIP%2BMCP%2BAuthorization%2BProxy%2Bfor%2BSystem%2BManagement%2BPurposes) which takes care for permissions and uses the stdio interface of the local MCPs. 

Testing the MCP server: 
To make use of the suseconnect functionality, the MCP server needs to run as root. That means to try it out, either run Claude as root (in a container) or use a wrapper like `suseconnect-mcp.sh` as the mcp endpoint: 

```
#!/bin/bash

# The executable to wrap (customize this)
EXECUTABLE="/space/workspace/connect-ng/out/suseconnect-mcp"

# Call the executable with sudo, passing all parameters through
sudo "$EXECUTABLE" "$@"
```

Add the MCP server to your config in ~/.gemini/settings.json or ~/.claude/settings.json or .mcp.json:

```
{
    "mcpServers": {
    ￼"suseconnect-mcp-stdio": {
      "command": "/space/workspace/connect-ng/out/suseconnect-mcp.sh",
      "args": []
    }
  }
}
```

Then, gemini can get use the tools for example to answer "What is the registration status of my system".

PR #382, opened 2026-05-05
Fail if any Product triplet parts are empty in FromTriplet()

rtamalin:
Add FromTriplet() tests extracted, ported over and simplified from
the proposed tests in SUSE/connect-ng#351, and updated to test the
new failure condition.

### DIFF
--- a/pkg/registration/product_test.go
+++ b/pkg/registration/product_test.go
@@ -158,3 +158,73 @@ func TestFindExtension(t *testing.T) {
 	_, err := product.FindExtension("sle-module-fakesystem/15.5/x86_64")
 	assert.Equal("extension not found", err.Error())
 }
+
+func TestFromTriplet(t *testing.T) {
+	assert := assert.New(t)
+
+	t.Run("valid triplets with component validation", func(t *testing.T) {
+		tests := []struct {
+			triplet      string
+			expectedName string
+			expectedVer  string
+			expectedArch string
+		}{
+			{
+				triplet:      "SLES/15/x86_64",
+				expectedName: "SLES",
+				expectedVer:  "15",
+				expectedArch: "x86_64",
+			},
+			{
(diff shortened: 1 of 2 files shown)


pkg/registration/product_test.go
```diff
@@ -158,3 +158,72 @@ func TestFindExtension(t *testing.T) {
...
+
+	t.Run("invalid triplets", func(t *testing.T) {
+		tests := []struct {
+			triplet string
+			name    string
+		}{
+			{triplet: "", name: "empty string"},
+			{triplet: "single", name: "single part"},
+			{triplet: "two/parts", name: "two parts"},
+			{triplet: "four/parts/too/many", name: "four parts"},
```
digitaltom, line 218: how about adding an empty part? 

```suggestion
			{triplet: "four/parts/too/many", name: "four parts"},
			{triplet: "sle-module-basesystem//aarch64", name: "empty part"},
```

rtamalin: @digitaltom Right now that won't fail, because the current implementation of FromTriplet() just does a `strings.Split(triplet, "/")` and the only validation is to count the number of response elements, and fails if not 3. As such, "//" is valid as it will return 3 parts, though they would be empty.

If we want it to fail for an empty part then we need to implement that in the code, and I'm happy to do so.

digitaltom: I think it makes sense, while we're at it.

rtamalin: I have incorporated a version of your suggestion and updated FromTriplet() to fail if any of the parts are empty.

digitaltom (approved):

rtamalin:
Updated the title for a third time and it appears to have "taken" this time.

gbuenodevsuse (approved):
LGTM

PR #383, opened 2026-05-05
Restore exit status 71 handling for unregistered keepalive

rtamalin:
If attempting a keepalive for an unregistered system we should be returning the special ErrPingFromUnregistered error not a generic error message. Doing so will trigger the exit code 71 handling in the ExitOnError() routine in cmd/suseconnect.go.

Update suseconnect-ng.changes with entry referencing bsc#1263772.

Add a feature test to verify this scenario.

### DIFF
--- a/build/packaging/suseconnect-ng.changes
+++ b/build/packaging/suseconnect-ng.changes
@@ -4,6 +4,8 @@ Mon Apr  6 15:40:58 UTC 2026 - Fergal Mc Carthy <fmccarthy@suse.com>
 - Update version to 1.22 (pre):
   - InstallReleasePackage should consider zypperFilesystemRoot
     (jsc#SCC-630).
+  - Restore exit code 71 handling when attempting keepalive and not
+    registered (bsc#1263772)
   
 -------------------------------------------------------------------
 Wed Apr  1 16:43:26 UTC 2026 - Fergal Mc Carthy <fmccarthy@suse.com>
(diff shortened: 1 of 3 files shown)


jeremy-moffitt (approved):
looks straightforward to me. I did have to look up where ErrPingFromUnregistered comes from and I see it in the errors.go file.

felixsch (approved):
lgtm 👍 

**How I reviewed the merge request:**

- Checked that the code changes fix the issue
- Run the branch and checked an unregistered system returns status code `71` as defined in the service file
```bash
d7a743782277:/connect # ./out/suseconnect --keepalive; echo $?
Error sending keepalive: System is not registered. Use the --regcode parameter to register it.
71
d7a743782277:/connect # cat build/packaging/suseconnect-keepalive.service 
[Unit]
Description=Run SUSEConnect --keepalive
Wants=suseconnect-keepalive.timer

[Service]
Type=oneshot
ExecStart=/usr/bin/SUSEConnect --keepalive
EnvironmentFile=-/etc/sysconfig/proxy

# Filter out this special exit code, since it will be raised by `SUSEConnect
# --keepalive` whenever a system is not registered, which is a pretty typical
# scenario.
SuccessExitStatus=71
```

**As always, if you think I missed something I should test, please let me know and I have a look ASAP**:rocket:

gbuenodevsuse (approved):
LGTM

PR #386, opened 2026-05-07
Fix typo in kubernetes_provider testing

rtamalin:
Spotted by Brett, the case check should be for enabled rather than ensabled.


jeremy-moffitt (approved):

brett060102 (approved):
Thank you

PR #387, opened 2026-05-13
Add HA collector

brett060102:
The HA collector determines if pacemaker is running on the system. If pacemaker is active, the record ha_active is added to the sysinfo blob with contents returned by "pacemaked --version".

If pacemaker is not running, the ha_active record is NOT included in the sysinfo blob.

### DIFF
--- a/internal/collectors/collectors.go
+++ b/internal/collectors/collectors.go
@@ -13,6 +13,7 @@ Example:
 		collectors.Memory{},
 		collectors.UUID{},
 		collectors.Vendor{},
+		collectors.HA{},
 	}
 
 result, error := collectors.CollectInformation("x86_64", MandatoryCollectors)
--- a/internal/collectors/ha.go
+++ b/internal/collectors/ha.go
@@ -1,0 +1,87 @@
+package collectors
+
+import (
+	"bufio"
+	"bytes"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/SUSE/connect-ng/internal/util"
+)
+
+type HA struct{}
+
+// Get Pacemaker version
+func getPacemakerVersion() (string, error) {
+	pacemakerdOut, err := util.Execute([]string{"pacemakerd", "--version"}, nil)
+	if err != nil {
+		return "", err
+	}
+	scanner := bufio.NewScanner(bytes.NewReader(pacemakerdOut))
+	for scanner.Scan() {
+		testLine := scanner.Text()
+		fields := strings.Fields(testLine)
+		if len(fields) >= 2 {
+			if fields[0] == "Pacemaker" {
+				return fields[1], nil
+			}
+		}
+	}
+
+	return "", scanner.Err()
+}
+
+func isPacemakerRunning(systemdClient util.SystemdClient) (bool, error) {
+	// first try to use ListUnitsByPatterns, failing back to ListUnits if not available
+	units, err := systemdClient.ListUnitsByPatterns("pacemaker.service")
+	if err != nil {
+		fmt.Printf("!!!!!!!!!!systemdClient.ListUnitsByPatterns() failed: %s %t\n", err, errors.Is(err, util.SystemdMethodNotAvailable))
+		if !errors.Is(err, util.SystemdMethodNotAvailable) {
+			util.Debug.Printf("systemdClient.ListUnitsByPatterns() failed: %s\n", err)
+			return false, err
+		}
+		units, err = systemdClient.ListUnits()
+	}
+
+	util.Debug.Printf("pacemakerActive: units: %+v err: %+v\n", units, err)
+	if err != nil {
+		util.Debug.Printf("systemdClient.ListUnits() failed: %s\n", err)
+		return false, err
+	}
+
+	for _, unit := range units {
+		util.Debug.Printf("pacemakerActive: unit: %+v \n", unit)
+		// is pacemaker running
+		if unit.Name.Match("pacemaker.service") && unit.SubState == "running" {
+			return true, nil
+		}
+	}
+	return false, nil
+
+}
+
+func isSystemHA(systemdClient util.SystemdClient) (Result, error) {
+	haActive, err := isPacemakerRunning(systemdClient)
+	util.Debug.Print("ha:haActive ", haActive)
+	if !haActive {
+		return Result{}, err
+	}
+	pacemakerVersion, err := getPacemakerVersion()
+	if pacemakerVersion != "" {
+		result := Result{"ha_active": pacemakerVersion}
+		util.Debug.Printf("ha:pacemakerVersion:Result: %+v \n", result)
+		return result, err
+	}
+	return Result{}, err
+}
+
+func (HA) run(arch string) (Result, error) {
+	systemdClient, err := util.NewDbusSystemdClient()
+	if err != nil {
+		return nil, err
+	}
+	defer systemdClient.Close()
+	result, err := isSystemHA(systemdClient)
+	return result, err
+}
--- a/internal/collectors/ha.go
+++ b/internal/collectors/ha.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"bytes"
 	"errors"
-	"fmt"
 	"strings"
 
 	"github.com/SUSE/connect-ng/internal/util"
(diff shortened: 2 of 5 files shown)


internal/collectors/collectors.go
```diff
@@ -13,6 +13,7 @@ Example:
 		collectors.Memory{},
 		collectors.UUID{},
 		collectors.Vendor{},
+		collectors.HA{},
```
rtamalin, line 16: Not sure that the change is needed here in the comment? Unless we want to keep the comment in sync with all the mandatory collectors, in which case there are a number of other ones missing. ;-)

internal/collectors/ha.go
```diff
@@ -0,0 +1,77 @@
...
+func getPacemakerVersion() (string, error) {
+	pacemakerdOut, err := util.Execute([]string{"pacemakerd", "--version"}, nil)
+	if err != nil {
+		return "", err
+	}
+	scanner := bufio.NewScanner(bytes.NewReader(pacemakerdOut))
+	for scanner.Scan() {
+		testLine := scanner.Text()
+		fields := strings.Fields(testLine)
+		if len(fields) == 2 {
```
rtamalin: So if the output of the `pacemaker --version` gets a third field, you would ignore the line?

It may be better to have this check be >= 2 so that ensure you check lines that have at least 2 fields, but tolerate extra fields beyond the first two, which will be ignored.


```suggestion
		if len(fields) >= 2 {
```

internal/collectors/ha.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	return "", scanner.Err()
+}
+
+func isPacemakerRunning(systemdClient util.SystemdClient) (bool, error) {
+	// Use systemdClient.ListUnits because systemdClient.ListUnitsByPatterns
+	// is not avaibale on all supported systems.
+	units, err := systemdClient.ListUnits()
```
rtamalin: I'd prefer that we repeat the same pattern used for the K8s collector, as the ListUnitsByPatterns() call is more efficient, and what we should prefer using for any streams where it is available; only the SLE 12 SP5 stream requires use to use ListUnits() and LTSS Core support for that stream ends in 2030, where as SLE 15 SP7 are supported beyond that.

The fallback pattern is not much extra code, but takes advantage of the more efficient call when it is available, avoiding the need to process every defined unit on the system, since it will filter for the desired name up front.

rtamalin (changes requested):
Please use the same  ListUnitsByPatterns() call falling back to ListUnits() pattern as used in the K8s collector.

internal/collectors/ha_test.go
```diff
@@ -0,0 +1,251 @@
...
+	systemdClient := testEnv.NewClient()
+	defer testEnv.Cleanup()
+
+	// run test case
+	res, err := isPacemakerRunning(systemdClient)
+
+	// check returned values
+	assert.NoError(err, "isPacemakerRunning() returned unexpected error")
+	assert.Equal(true, res, "isPacemakerRunning() returned unexpected result")
+
```
rtamalin, line 64: Unnecessary blank line?

```suggestion
```

internal/collectors/ha_test.go
```diff
@@ -0,0 +1,251 @@
...
+	systemdClient := testEnv.NewClient()
+	defer testEnv.Cleanup()
+
+	// run test case
+	res, err := isPacemakerRunning(systemdClient)
+
+	// check returned values
+	assert.NoError(err, "isPacemakerRunning() returned unexpected error")
+	assert.Equal(false, res, "isPacemakerRunning() returned unexpected result")
+
```
rtamalin: Unnnecessary blank line?


```suggestion
```

internal/collectors/ha_test.go
```diff
@@ -0,0 +1,251 @@
...
```
rtamalin: Unnnecessary blank line?


```suggestion
```

internal/collectors/ha_test.go
```diff
@@ -0,0 +1,251 @@
...
+	systemdClient := testEnv.NewClient()
+	defer testEnv.Cleanup()
+
+	// run test case
+	res, err := isPacemakerRunning(systemdClient)
+
+	// check returned values
+	assert.Error(err, "isPacemakerRunning() failed to returned expected error")
+	assert.Equal(false, res, "isPacemakerRunning() returned unexpected result")
+
```
rtamalin, line 120: Unnnecessary blank line?


```suggestion
```

internal/collectors/ha_test.go
```diff
@@ -0,0 +1,251 @@
...
+	// setup execute for pacemakerd --version
+	mockUtilExecute(strings.Join([]string{paceMakerName, pacemakerVersion}, " "), nil)
+
+	// run test case
+	res, err := getPacemakerVersion()
+
+	// check returned values
+	assert.NoError(err, "getPacemakerVersion() returned expected error")
+	assert.Equal(pacemakerVersion, res, "getPacemakerVersion() returned unexpected result")
+
```
rtamalin: Unnnecessary blank line?


```suggestion
```

internal/collectors/ha_test.go
```diff
@@ -0,0 +1,251 @@
...
+	// run test case
+	res, err := isPacemakerRunning(systemdClient)
+
+	// check returned values
+	assert.Error(err, "isPacemakerRunning() failed to returned expected error")
+	assert.Equal(false, res, "isPacemakerRunning() returned unexpected result")
+
+}
+
+func TestGetPacemakerVersionPass(t *testing.T) {
```
rtamalin: You can probably use a single test function with a test table pattern to test the various pacemaker version scenarios. See https://github.com/SUSE/connect-ng/blob/45839a4fab1d1ee98df386e99da818aaf373514a/internal/collectors/memory_test.go#L23 or https://github.com/SUSE/connect-ng/blob/45839a4fab1d1ee98df386e99da818aaf373514a/pkg/registration/product_test.go#L162 for examples of a test table pattern.

internal/collectors/ha_test.go
brett060102: thx

internal/collectors/ha_test.go
brett060102: I will take a look, but I do like how trackable a problem is with the current method.

internal/collectors/collectors.go
brett060102: Probably not. but it doesn't hurt either.

internal/collectors/ha.go
brett060102: done

internal/collectors/ha.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return "", scanner.Err()
+}
+
+func isPacemakerRunning(systemdClient util.SystemdClient) (bool, error) {
+	// first try to use ListUnitsByPatterns, failing back to ListUnits if not available
+	units, err := systemdClient.ListUnitsByPatterns("pacemaker.service")
+	if err != nil {
+		fmt.Printf("!!!!!!!!!!systemdClient.ListUnitsByPatterns() failed: %s %t\n", err, errors.Is(err, util.SystemdMethodNotAvailable))
```
rtamalin: I don't think we need all of these `!` marks here - is this a debugging printf that was mistakenly left in?


```suggestion
```

internal/collectors/ha_test.go
rtamalin: With the table based method you can still name the tests, and the name can be part of the table of inputs, so you can know which inputs triggered the issue.

Most of these tests are the exact same pattern of common setup code, with some amount of test specific setup, and then the same pattern of checking code. An approach where there is only one occurrence of the setup and checking code is easier to maintain in the long run, say if we modify the tooling used in the setup.

The other approach would be to use named subtest functions... That would possibnly allow you to at least leverage the common setup aspects of things at the start of your main test function so long as you can get away with only having to do it once, but you still need to duplicate the checking code pattern in each subtest function.

Which is why I was suggesting table based testing, where you define you test inputs in a table, and then have a loop that iterates over the table, implementing the setup, test run and post test checking as the loop body, which is east to maintain and extend with additional test case.

However, we can live with the individual tests. ;-)

rtamalin (approved):
Looks good - we can revisit test cleanup at a later time

internal/collectors/ha_test.go
brett060102: done

PR #390, opened 2026-05-27
Add agama testing GitHub Actions workflow

rtamalin:
Add a new agama tests workflow to the GitHub Actions workflows.

The workflow consists of two active phases:
  * build the SUSE/connect-ng repo binaries using the SLE BCI golang container image, uploading the tarballed `out` directory as an artifact
  * build the agama-project/agama repo suseconnect-agama activation test using the `out` directory contents from the retrieved artifact and the SLE BCI rust container image, and then run the activation test.

Also add an `agama-tests` target to the Makefile, along with some
additional helper targets, that:
  * clones the agama-project/agama repo under testdata/agama_srcs (ignored by git)
  * builds the SUSE/connect-ng repo using the SLE BCI golang container image
  * builds the agama-projects/agama repo's rust activation testing tool using the SLE BCI rust container image and runs the activation test.

Also fixes a potential bug in libsuseconnect where the agama activation test tool would fail if no config settings were loaded.

Co-Authored-By: esampson@suse.com

Jira: SCC-775

### DIFF
--- a/.github/workflows/agama.yml
+++ b/.github/workflows/agama.yml
@@ -1,0 +1,112 @@
+---
+name: agama tests
+
+# About security when running the tests and NOT exposing
+# the secrets to externals. Currently Github Actions does
+# NOT expose the secrets if the branch is coming from a forked
+# repository.
+# See: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/
+# See: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions
+#
+# An alternate would be to set, pull_request_target but this takes the CI code
+# from master removing the ability to change the code in a PR easily.
+#
+# Aditionally, since 2021 pull requests from new contributors will not
+# trigger workflows automatically but will wait for approval from somebody
+# with write access.
+# See: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
+on: [pull_request]
+
+env:
+  CONNECTNG_SOURCE: /usr/src/connect-ng
+  AGAMA_SOURCE: /usr/src/agama
+  CONNECTNG_ARTIFACT: suseconnect-build-out
+  CONNECTNG_TARBALL: /usr/src/suseconnect-build-out.tar.gz
+  REGCODE: ${{ secrets.REGCODE }}
+
+jobs:
+
+  build-suseconnect:
+    name: Build SUSE/connect-ng inputs required by Agama
+    runs-on: ubuntu-latest
+    container:
+      image: registry.suse.com/bci/golang:1.24-openssl
+      options: --user root --privileged
+
+    steps:
+      - name: Install dependencies for artifact upload/download
+        run: |
+          zypper --non-interactive in nodejs-default
+
+      - name: Checkout SUSE/connect-ng
+        uses: actions/checkout@v4
+        with:
+          path: connect-ng
+
+      - name: Move checked out connect-ng to /usr/src
+        run: |
+          mv connect-ng ${{ env.CONNECTNG_SOURCE }}
+
+      - name: Build SUSE/connect-ng
+        working-directory: ${{ env.CONNECTNG_SOURCE }}
+        run: |
+          make vendor build
+
+      - name: Create a tarball of the build output
+        working-directory: ${{ env.CONNECTNG_SOURCE }}
+        run: |
+          tar -czf ${CONNECTNG_TARBALL} out/
+
+      - name: Upload the tarball
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ env.CONNECTNG_ARTIFACT }}
+          path: ${{ env.CONNECTNG_TARBALL }}
+          retention-days: 1 # no need to retain this for long
+
+  run-agama-rust-tests:
+    name: Run agama rust tests
+    needs: build-suseconnect
+    runs-on: ubuntu-latest
+    container:
+      image: registry.suse.com/bci/rust:1.95
+      options: --user root --privileged
+
+    steps:
+      - name: Install dependencies for artifact upload/download
+        run: |
+          zypper --non-interactive in nodejs-default
+
+      - name: Checkout SUSE/connect-ng
+        uses: actions/checkout@v4
+        with:
+          path: connect-ng
+
+      - name: Move checked out connect-ng to /usr/src
+        run: |
+          mv connect-ng ${CONNECTNG_SOURCE}
+
+      - name: Download suseconnect-build-out tarball
+        uses: actions/download-artifact@v4
+        with:
+          name: ${{ env.CONNECTNG_ARTIFACT }}
+          path: /usr/src  # dirname of env.CONNECTNG_TARBALL
+
+      - name: Extract tarball contents to expected location
+        run: |
+          tar -C ${CONNECTNG_SOURCE} -xzf ${CONNECTNG_TARBALL}
+
+      - name: Checkout agama-project/agama
+        uses: actions/checkout@v4
+        with:
+          repository: agama-project/agama
+          path: agama
+
+      - name: Move checked out agama code to /usr/src
+        run: |
+          mv agama ${AGAMA_SOURCE}
+
+      - name: Run agama rust tests
+        working-directory: ${{ env.CONNECTNG_SOURCE }}
+        run: |
+          build/ci/run-agama-rust-tests
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,15 @@ GOFLAGS       = -v -mod=vendor
 BINFLAGS      = -buildmode=pie
 SOFLAGS       = -buildmode=c-shared
 
-CONTAINER     = registry.suse.com/bci/golang:1.24-openssl
+GOCONTAINER   = registry.suse.com/bci/golang:1.24-openssl
+RUSTCONTAINER = registry.suse.com/bci/rust:1.95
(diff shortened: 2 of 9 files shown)


.github/workflows/agama.yml
```diff
@@ -0,0 +1,112 @@
...
+
+      - name: Checkout agama-project/agama
+        uses: actions/checkout@v4
+        with:
+          repository: agama-project/agama
+          path: agama
+
+      - name: Move checked out agama code to /usr/src
+        run: |
+          mv agama ${AGAMA_SOURCE}
```
felixsch, line 107:
```suggestion
      - name: Checkout agama-project/agama
        uses: actions/checkout@v4
        with:
          repository: agama-project/agama
          path: ${AGAMA_SOURCE}
```

Would this also work?

build/ci/run-agama-rust-tests
```diff
@@ -0,0 +1,101 @@
...
+
+group "install required packages"
+zypper in -y ${REQ_PKGS[@]}
+groupend
+
+group "clone the agama repo if needed"
+if [[ ! -d ${AGAMA_RUST} ]]; then
+	git clone ${AGAMA_REPO} ${AGAMA_SOURCE}
+fi
+groupend
```
felixsch: If this is in place, would it make sense to have either this cloning or the workflow action. Was there a reason to have both?

third_party/libsuseconnect/libsuseconnect.go
```diff
@@ -270,6 +270,9 @@ func loadConfig(clientParams string) *connect.Options {
 	// Read the options from the default configuration path and merge the
 	// provided clientParams into as well.
 	opts, _ := connect.ReadFromConfiguration(connect.DefaultConfigPath)
+	if opts == nil {
+		opts = connect.DefaultOptions()
+	}
```
felixsch: This looks like the fix for the options bug. Is this intended to be part of this PR?

Makefile
```diff
@@ -85,6 +90,9 @@ build: clean out internal/connect/version.txt
 	$(GO) build $(GOFLAGS) $(BINFLAGS) $(OUT) github.com/SUSE/connect-ng/cmd/suseconnect-mcp
 	$(GO) build $(GOFLAGS) $(SOFLAGS) $(OUT) github.com/SUSE/connect-ng/third_party/libsuseconnect
 
+bci-build:
+	$(CRM) $(MOUNT) -w $(WORKDIR) $(GOCONTAINER) bash -c 'make vendor build'
```
felixsch: This will lead to:
```
> make agama-tests
Cloning https://github.com/agama-project/agama under /files/suse/connect-ng/testdata/agama_srcs ...
Cloning into '/files/suse/connect-ng/testdata/agama_srcs'...
remote: Enumerating objects: 140229, done.
remote: Counting objects: 100% (104/104), done.
remote: Compressing objects: 100% (53/53), done.
remote: Total 140229 (delta 74), reused 56 (delta 51), pack-reused 140125 (from 2)
Receiving objects: 100% (140229/140229), 77.02 MiB | 4.85 MiB/s, done.
Resolving deltas: 100% (104127/104127), done.
docker run --rm -it --privileged -v /files/suse/connect-ng:/usr/src/connect-ng -w /usr/src/connect-ng registry.suse.com/bci/golang:1.24-openssl bash -c 'make vendor build'
all modules verified
go clean
mkdir -p out
go build -v -mod=vendor -buildmode=pie -o out/ github.com/SUSE/connect-ng/cmd/suseconnect
# cd /usr/src/connect-ng; git status --porcelain
fatal: detected dubious ownership in repository at '/usr/src/connect-ng'
To add an exception for this directory, call:

	git config --global --add safe.directory /usr/src/connect-ng
error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
make: *** [Makefile:83: build] Error 1
make: *** [Makefile:94: bci-build] Error 2
```
I wonder why this is needed. Doesn't the build script not already build the project in case it is not existing? If its needed and I do not see the point, then the above error should be addressed. I think on a fresh checkout without any configuration (besides `.env`) running tests should work.

.github/workflows/agama.yml
```diff
@@ -0,0 +1,112 @@
+---
+name: agama tests
```
felixsch, line 2: When trying to run `act` on this workflow I encounter:

```
$ act --list
INFO[0000] Using docker host 'unix:///var/run/docker.sock', and daemon socket 'unix:///var/run/docker.sock' 
Stage  Job ID                Job name                                        Workflow name           Workflow file  Events      
0      build-suseconnect     Build SUSE/connect-ng inputs required by Agama  agama tests             agama.yml      pull_request
0      feature-tests         feature-tests                                   feature tests           features.yml   pull_request
0      unit-tests            unit-tests                                      lint + unit tests       lint-unit.yml  pull_request
0      yast                  yast                                            YaST integration tests  yast.yml       pull_request
1      run-agama-rust-tests  Run agama rust tests                            agama tests             agama.yml      pull_request
```

Do you have a way to run only the agama tests with act? `act feature-tests` works fine but for the 2 stages workflow for agama I have not found a way to execute it without running the full suite.

rtamalin: I used the `-W .github/workflows/agama.yml' to tell it to just focus on that workflow for the run... You possibly can specify the final job in a dependency chain of jobs, e.g. use the `-j run-agama-rust-tests` and `act` may be able to figure out that it needs to run the other jobs because of the `needs` dependencies, but I don't think that has been reliable for me...

rtamalin: Just tested it and yeah it appears that you can name the last job in the chain and act figures it out, and runs the earlier jobs.

I can add a README.md section related to using `act` or `gh act` to run github workflow tests locally.

.github/workflows/agama.yml
rtamalin: That is what I used to have, but the checkout action isn't happy with it for security reasons; it will only work if AGAMA_SOURCE is under the top-level workspace directory path... And /usr/src isn't... ;-)

build/ci/run-agama-rust-tests
rtamalin: This originated from simplifying the workflow for a developer running the `make agama-tests`, but since then I have added the `agama-sources` target that takes care of cloning the repo, so yeah we can drop it here...

third_party/libsuseconnect/libsuseconnect.go
rtamalin: Yeah this was inherited from Brett's workaround hacks, and it something I need to test and hopefully can remove...

Makefile
rtamalin: So, as to why I implemented this, the `bci-build` target is the Makefile driven workflow's equivalent of the `build-suseconnect` job in the GHA workflow. This target is used so that we are building suseconnect binaries that match the ecosystem that the agama test will be running in, i.e. we build them in a SLE BCI golang image based upon (currently) SLE 15 SP7, and then use them in a SLE BCO rust image based (currently) upon SLE 15 SP7.

Otherwise, you would be building the suseconnect binaries locally on your host system, whatever distro that may be, and then attempting to use them within the SLE BCI rust image that is (currently) based upon SLE 15 SP7. While this usually works, it represents an "assumption" that may cause problems in the future... Note that things do currently work fine for me using locally built binaries within the SLE BCI rust image, on both my Leap 16.0 and Ubuntu 24.04 systems, but it might be better to use a workflow that avoids potential sources of future pain... ;-)

As for the git dubious ownership fatal error, I wasn't seeing it for git worktree add created SUSE/connect-ng directories, but I do see it if I do a fresh clone of SUSE/connect-ng... Will update to ensure we avoid this...

This whole issue is something I want to dig into this week, as the ownership issues problem causes problems with the vendor tree as well, when it is created within a container based run, the files are owned by root, and require you to use sudo rm to remove them when cleaning up from the host environment.

.github/workflows/agama.yml
rtamalin: We could also just drop the use of the checkout action and update the move step's shell action to `make agama-sources && mv testdata/agama_srcs /usr/src/agama` to do the checkout if we don't want to have duplication of the cloning stuff...

build/ci/run-agama-rust-tests
rtamalin: This should now be resolved; the run-agama-rust-tests now just checks and fails if the agama sources are not available.

third_party/libsuseconnect/libsuseconnect.go
rtamalin: Tested things and removed this code as it looks like it was required to workaround the issue that was fixed in the rmt.rs code where it was erroneously writing the creds to the /etc/SUSEConnect file.

Makefile
rtamalin: Added the suggested git config workaround to the invoked docker bash command line for now, and that seems to fix things for me in a fresh clone of the repo and also works fine for a git worktree setup.

.github/workflows/agama.yml
rtamalin: Done

felixsch (approved):
lgtm 👍 

- Read the pipeline changes and left my suggestions
- Run the pipeline using `act -W .github/workflows/agama.yml pull_request`
- Checked that the pipeline is green

cheers,

mssola (approved):
LGTM!

build/ci/run-agama-rust-tests
rtamalin: resolved

PR #391, opened 2026-05-28
Avoid double // at start of url path component

rtamalin:
Trim off any trailing / in the base URL, and any leading / in the path, and join with a / when constructing fullUrl in BuildRequestRaw.

Add a unit test to catch this scenario.

Jira: SCC-775

### DIFF
--- a/pkg/connection/connection_test.go
+++ b/pkg/connection/connection_test.go
@@ -39,21 +39,73 @@ func TestConnectionBuildRequest(t *testing.T) {
 	assert.Equal(request.Header.Get("Accept-Language"), "en_US")
 }
 
-func TestConnectionBuildRequestDoubleSlashHandling(t *testing.T) {
+func TestConnectionBuildRequestRawSlashHandling(t *testing.T) {
 	assert := assert.New(t)
 
-	opts := DefaultOptions("testApp", "1.0", "en_US")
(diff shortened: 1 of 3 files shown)


pkg/connection/connection_test.go
```diff
@@ -39,6 +39,23 @@ func TestConnectionBuildRequest(t *testing.T) {
 	assert.Equal(request.Header.Get("Accept-Language"), "en_US")
 }
 
+func TestConnectionBuildRequestDoubleSlashHandling(t *testing.T) {
```
jeremy-moffitt: should we also test the version without the extra slash to make sure nothing we want in the URL is getting accidentally trimmed?

digitaltom (approved):

pkg/connection/connection_test.go
rtamalin: That is effectively what the previous test does because it uses the default https://scc.suse.com URL.

I can make it explicit ;-)

rtamalin: I added multiple test cases, and switched to testing BuildRequestRaw directly and realised that TrimPrefix & TrimSuffix weren't viable when dealing with multiple leading or trailing slashes, so switched to TrimLeft and TrimRight respectively.

jeremy-moffitt (approved):
LGTM, thanks for the clarity around the default test

PR #393, opened 2026-06-01
util.MockExecutor helps with util.Execute mocking

rtamalin:
Leveraging the stretchr/testify/mock to provide a more capabale means for mocking util.Execute() calls.

The MockExecutor.Setup() routine is used to replace util.Execute() with the MockExecutor's Execute() handler, and returns a teardown function that should be called with defer to restore the original Execute().

The MockExecutor.Execute() routine is mock replacement for Execute() that checks for a pre-registered command line mocking and if found, handles it appropriately.

The MockExecutor.OnExecute() routine allows registration of a command line to be mocked, returing a mock.Call that can be further used to manage the mocking of that call; for further details see the https://pkg.go.dev/github.com/stretchr/testify/mock#Call docs.

The MockExecutor.OnExecuteReturn() return allows for registration of a command line and it's asssociated return values in a single command to simplify setting up mocking for a single command.

Update ha.go and kubernetes_provider.go testing in internal/collectors to leverage the new MockExecutor for mocking util.Execute() calls.

Also moved the systemdTestEnv support code that is shared by both of these collector implmentations into a separate test_helpers.go file, and added support for defining regular and dummy systemd services, and updated ha.go to use these new service definition helpers.

Tweaked ha.go and ha_test.go to leverage constants for fixed string values and used full path for pacemakerd binary when calling it.

### DIFF
--- a/internal/util/system.go
+++ b/internal/util/system.go
@@ -40,8 +40,11 @@ func SetSystemEcho(v bool) bool {
 	return prev
 }
 
+// ExecuteFunc type
+type ExecuteFunc func(cmd []string, validExitCodes []int) ([]byte, error)
+
 // Assign function for running external commands to a variable so it can be mocked by tests.
-var Execute = func(cmd []string, validExitCodes []int) ([]byte, error) {
+var Execute ExecuteFunc = func(cmd []string, validExitCodes []int) ([]byte, error) {
 	Debug.Print("Executing: ", cmd)
 	var stderr, stdout bytes.Buffer
 	comm := exec.Command(cmd[0], cmd[1:]...)
--- a/internal/collectors/ha.go
+++ b/internal/collectors/ha.go
@@ -9,11 +9,16 @@ import (
 	"github.com/SUSE/connect-ng/internal/util"
 )
 
+const (
+	// Pacemaker constants
+	pacemakerCmdPath = "/usr/sbin/pacemakerd"
+)
+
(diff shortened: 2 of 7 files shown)


internal/util/system.go
```diff
@@ -40,8 +40,11 @@ func SetSystemEcho(v bool) bool {
 	return prev
 }
 
+// ExecuteFunc type
+type ExecuteFunc func(cmd []string, validExitCodes []int) ([]byte, error)
+
 // Assign function for running external commands to a variable so it can be mocked by tests.
-var Execute = func(cmd []string, validExitCodes []int) ([]byte, error) {
+var Execute ExecuteFunc = func(cmd []string, validExitCodes []int) ([]byte, error) {
```
felixsch, line 47: Out of curiosity why did this end up being added? What was the issue you encountered?

rtamalin: I added the type definition for ExecuteFunc above so that we have a signature to verify against at compile time.

By adding this here we explicitly define that we want the Execute variable to have this specific signature, rather than it being implicitly defined by the initial function that is assigned to it.

We could live without it, but this way if someone tweaks the signature of the func() being assigned here we get a clear compile time error for this assignment statement, rather than getting errors elsewhere, i.e. in mocking code that may be less clear.

brett060102 (approved):
LGTM

PR #395, opened 2026-06-08
Use product identifier for finding product packages

rtamalin:
With the previous refactoring of the code base after the v1.13.0 release, the appropriate value to identify the product is now found in the identifier field, rather than the name field.

With this change in place I now see equivalency between the zypper and migration actions that are generated by both the v1.13.0 and latest code when performing migrations.

Jira: SCC-758
Bugzilla: bsc#1265410


gbuenodevsuse (approved):
LGTM

brett060102 (approved):
lgtm

jeremy-moffitt (approved):
LGTM

PR #397, opened 2026-06-10
Allow pkg/ clients to disable duplicate token detection completely

felixsch:
task: https://jira.suse.com/browse/SCC-801

Allows to use the API without sending the `Token-Header`. This disables the token handling (aka duplicate detection)

You need:
  - Checked out branch
  - Running golang environment
  - A regcode to test against

```
$ make build
#expect: It builds

# tokens enabled
$ REGCODE="<SLES REGCODE>" ./out/public-api-demo SLES 15.4 x86_64
# Run until "System fully activated"
# You see the link to the system in SCC. Open it and go to the admin version of the page (little shield icon next to the hostname)
# expect: The token is set as expected
# Finalize the demo

# tokens disabled
$ DISABLE_TOKEN_HANDLING=1 REGCODE="<SLES REGCODE>" ./out/public-api-demo SLES 15.4 x86_64
# Run until "System fully activated"
# You see the link to the system in SCC. Open it and go to the admin version of the page
# expect: The token is not set
# Finalize the demo
```

**Hint:** If you have trouble finding the admin version of the system details page: Just add `/admin/`  to the system details URL: `https://scc.suse.com/systems/25370112 -> https://scc.suse.com/admin/systems/25370112`

**Pro tip:** If you really want to now it you can enable credential change tracing with: `TRACE_CREDENTIAL_UPDATES=1`

**As always, if you think I missed something or you have suggestions, do not hesitate to reach out to me**:rocket:

thank you :smile_cat:

### DIFF
--- a/build/packaging/suseconnect-ng.changes
+++ b/build/packaging/suseconnect-ng.changes
@@ -17,6 +17,7 @@ Mon Apr  6 15:40:58 UTC 2026 - Fergal Mc Carthy <fmccarthy@suse.com>
     migrations. (jsc#SCC=758 bsc#1265410)
   - Add opt in/out support for collectors (jsc#TEL-312)
   - Update config parser for suseconnect to be YAML based (jsc#SCC-730)
+  - library: Allow clients to disable the token handling mechanism (jsc#SCC-801)
 
 -------------------------------------------------------------------
 Wed Apr  1 16:43:26 UTC 2026 - Fergal Mc Carthy <fmccarthy@suse.com>
--- a/cmd/public-api-demo/main.go
+++ b/cmd/public-api-demo/main.go
@@ -42,6 +42,10 @@ func runDemo(identifier, version, arch, infoPath, regcode string) error {
 		isProxy = true
 	}
 
+	if disableTokens := os.Getenv("DISABLE_TOKEN_HANDLING"); disableTokens != "" {
+		opts.DisableTokenHandling = true
+	}
+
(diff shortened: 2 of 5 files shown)


mallardduck (approved):
LGTM. Should be simple enough for someone on ORBS team to implement in SCC Operator too once this is merged w/ stable git tag.

rtamalin (approved):
LGTM

rjschwei:
But this makes it really easy in SLE to disable duplicate detection a simple entry in  `/etc/profile.local` will do the trick. IMHO this should be gated on whether or not there is a product. AFAIK Rancher does not set a product and in SLE the product is generated from the content of `/etc/products/baseproduct` or based on whetever is set on the `suseconnect` command line with the `-p` option. Anyway, I think this makes it way to simple to circumvent duplicate detection.

rtamalin:
> But this makes it really easy in SLE to disable duplicate detection a simple entry in `/etc/profile.local` will do the trick. IMHO this should be gated on whether or not there is a product. AFAIK Rancher does not set a product and in SLE the product is generated from the content of `/etc/products/baseproduct` or based on whetever is set on the `suseconnect` command line with the `-p` option. Anyway, I think this makes it way to simple to circumvent duplicate detection.

@rjschwei The env var in question is not checked for as part of the main suseconnect code base, but rather as part of the public-api-demo tool which is used for demoing and testing the new pkg/ based interfaces the core suseconnect functionality, as consumed by Rancher... So setting that env var on a system would have no impact on normal suseconnect token handling.

PR #398, opened 2026-06-10
add 1.22.1 to changes log and move post 1.22.0 change in

brett060102:
Since 1.22.0 was accepted we need to move version to 1.22.1
which requires adding entry to change log.


rtamalin (approved):
LGTM

PR #399, opened 2026-06-10
Update spec file for 1.22.1

brett060102:
Need to update spec file for 1.22.1


rtamalin (approved):
Looks good - should have spotted that myself but a little distracted ;-)

PR #400, opened 2026-06-11
Include updates certs when setting up HTTP clients

rtamalin:
Restore the earlier TLS client cert handling by ensuring that we load the latest available system certs when setting up any HTTP client connections. This is required because the Golang crypto/x509 module samples the system cert pool once, and reuses that pool state even if the system certs are updated. This is a known Golan issue tracked in https://github.com/golang/go/issues/41888 that the earlier version of suseconnect-ng worked around, but the workaround was not migrated to the new HTTP client connection setup as part of the refactoring efforts.

Moves internal/connect/cert_pool.go to pkg/connection and leverages it in the ApiConnection.setupHTTPClient() receiver to ensure that we are using a pool that includes any recently updated certs.

Added some debug messages to cert_pool.go for ignored errors.

Issues: bsc#1268017, jsc#SCC-804

### DIFF
--- a/pkg/connection/connection.go
+++ b/pkg/connection/connection.go
@@ -9,6 +9,8 @@ import (
 	"io"
 	"net/http"
 	"strings"
+
+	"github.com/SUSE/connect-ng/internal/util"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


rtamalin:
To test this I was using a modified rmt.rs from the agama-project/agama rust/suseconnect-agama/examples - see my fork https://github.com/rtamalin/agama/tree/add_optional_cert_path_second_argument_to_rmt_example

rtamalin:
I was also leveraging the Agama testing framework from #390 which can be enhanced in the future to leverage this scenario if we can get equivalent rmt.rs enhancements accepted to agama-project/agama.

pkg/connection/connection.go
```diff
@@ -9,6 +9,8 @@ import (
 	"io"
 	"net/http"
 	"strings"
+
+	"github.com/SUSE/connect-ng/internal/util"
```
felixsch, line 13: I'm not a huge fan of this. But if the debugging output is important to you then I'm fine

brett060102 (approved):

pkg/connection/connection.go
rtamalin: I'm happy to come up with a better way to do this but I think it will be useful in the long run to have some sort of debug message being printed here since otherwise we have no way of know what went wrong...

But we can also drop the debug messages is you don't want to be pulling in the internal/util here ;-)

rtamalin: Created SCC-806 to track the desire to clean this sort of thing up.

PR #403, opened 2026-06-25
Add coverage reporting for feature tests

rtamalin:
The existing feature-tests GitHub Actions workflow has been split into two separate workflows:

  * `build-rpm` which builds and installs the RPM package
  * `feature-tests` which builds suseconnect locally and runs the feature tests

This change was required to avoid needing to add support for building the RPM with coverage collection support enabled.

Updated the unit and feature test GitHub Actions workflows to add a final step that outputs the specific test suite's coverage reports.

Running the `test` or `feature-tests` targets will run generate a test suite specific coverage report after a successful run. And there are now additional Makefile targets that will report the specific coverage reports for these test suites:

  * `unit-test-coverage` shows the package and function percentage
  * `feature-tests-coverage` shows the package and function percentage

Split the Makefile `go-tool-covdata` helper routine into separate `go-tool-covdata-merge` and `go-tool-covdata-report` helpers.

The `coverage`, `coverage-func` and `coverage-percent` targets now generate reports based upon the unified coverage data, rather than the unit test specific data.

Added a new `coverage-clean` Makefile target for clearing out existing coverage data so that subsequence tests start with a clean state.

Added a new `run-tests` Makfile target that clears existing coverage data and then runs all the test suites sequentially before finally generated summary package and function level coverage reports of the unified data.

Updated building of `libsuseconnect.so` to call the built library `libsuseconnect.so` rather than `libsuseconnect` to align with how the RPM spec file actually builds the shared library. Updated the existing use cases of `libsuseconnect` to reflect this change.

Added `COVERAGE_BIN` and `COVERAGE_LIB` variables that are used to control whether the tools or libraries are built with support for collecting coverage data.

Added `build/ci/lib.bash` and moved shared scripting functionality from the other `build/ci` scripts into it; the other scripts now source it, and call the functions it provides. This will simplify maintenance of these scripts, minimising duplicated code, as we add more test scenarios in the future.

Minor cleanup to Makefile `dist` and `clean` target recipes and added a `real-clean` target that cleans up everything.

Also added a `fix-ownership` target that fixes the ownership of files and directories that may be created/modified by docker based actions.

Updated README.md to reflect these changes.

Implements: SCC-736

How To Test:
  * Clone this PR branch
  * Exercise the updated `test` and `feature-tests` Makefile targets to verify the run as expected and that they produce test suite specific coverage reporting.
  * Similarly, verify that the new `unit-test-coverage` and `feature-tests-coverage` Makefile targets report matching coverage reports for the most recent runs.
  * Verify that the `coverage`, `coverage-func` and `coverage-percent` Makefile targets report unified (aggregate) testing coverage reports.
  * Verify that the GitHub Actions workflows are being run as expected for this PR.
  * If available, use `act` (or `gh act`) to verify that the new and updated workflows run successfully, and where appropriate, generate coverage reports.
  * If running on Leap 16.0 and trying to use `act` see the new README.md section added in this PR for tips on how to resolve docker-in-docker permissions issues.

### DIFF
--- a/.github/workflows/build-rpm.yml
+++ b/.github/workflows/build-rpm.yml
@@ -1,5 +1,5 @@
 ---
-name: feature tests
+name: build and install rpm
 
 # About security when running the tests and NOT exposing
 # the secrets to externals. Currently Github Actions does
@@ -20,13 +20,10 @@ on: [pull_request]
 env:
   SOURCE: /usr/src/connect-ng
   ARTIFACT_SOURCE: artifacts
-  COVERAGE_BIN: true
   REGCODE: ${{ secrets.REGCODE }}
-  HA_REGCODE: ${{ secrets.HA_REGCODE }}
-  EXPIRED_REGCODE: ${{ secrets.EXPIRED_REGCODE }}
 
 jobs:
-  feature-tests:
+  build-rpm:
     runs-on: ubuntu-latest
     container:
       image: registry.suse.com/bci/golang:1.24-openssl
@@ -40,22 +37,14 @@ jobs:
           [ -d $SOURCE ] && rm -r $SOURCE
           cp -r $GITHUB_WORKSPACE $SOURCE
 
-      - name: build suseconnect binaries
-        run: |
-          set -e
-          cd $SOURCE
-          make vendor
-          make build
+      - name: build RPM package
+        run: bash build/ci/build-rpm
 
-      - name: run feature tests
-        run: |
-          set -e
-          cd $SOURCE
-          make coverage-dirs
-          bash build/ci/run-feature-tests
+      # - name: gather RPM build artifacts
+      #   uses: actions/upload-artifact@v4
+      #   with:
+      #     name: connect-rpms
+      #     path: ${{ env.SOURCE }}/${{ env.ARTIFACT_SOURCE }}/*.rpm
 
-      - name: show feature tests coverage reports
-        run: |
-          set -e
-          cd $SOURCE
-          make feature-tests-coverage
+      - name: configure Connect to run feature tests
+        run: bash build/ci/configure
--- a/.github/workflows/build-rpm.yml
+++ b/.github/workflows/build-rpm.yml
@@ -1,20 +1,6 @@
 ---
 name: build and install rpm
 
-# About security when running the tests and NOT exposing
-# the secrets to externals. Currently Github Actions does
-# NOT expose the secrets if the branch is coming from a forked
-# repository.
-# See: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/
-# See: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions
-#
-# An alternate would be to set, pull_request_target but this takes the CI code
-# from master removing the ability to change the code in a PR easily.
-#
-# Aditionally, since 2021 pull requests from new contributors will not
-# trigger workflows automatically but will wait for approval from somebody
-# with write access.
-# See: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
 on: [pull_request]
 
 env:
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,13 @@
 NAME          = suseconnect-ng
 VERSION       = $(shell bash -c "cat build/packaging/suseconnect-ng.spec | sed -n 's/^Version:\s*\(.*\)/\1/p'")
-DIST          = $(NAME)-$(VERSION)/
+DIST          = $(NAME)-$(VERSION)
 PWD           = $(shell pwd)
 GO            = go
 OUT           = -o out/
 # coverage testing enabled by default
 COVERAGE     ?= true
-COVERAGE_DIR  = $(PWD)/coverage
-COVERAGE_UNIT = $(COVERAGE_DIR)/unit
+COVERAGE_BIN ?= false
+COVERAGE_LIB ?= false
 GOFLAGS       = -v -mod=vendor
 BINFLAGS      = -buildmode=pie
 SOFLAGS       = -buildmode=c-shared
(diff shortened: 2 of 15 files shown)


mssola:
Looking at the code I don't see anything bad. But to me, I have the feeling that the code on CI is getting more and more complicated. As in, with time we have added complexity to CI code, which is not great in my mind, but this PR just puts even more stuff on top of it. And, in the end, we are adding maintenance burden for code that I personally wouldn't be too fond to maintain 😄 

I really don't want to step the wrong way since I see that this was already planned by the team. Hence, feel free to ignore this "old man yells at cloud" comment. But, is this really something that important that deserves the added complexity into something as brittle as CI? (brittle as in: 1. the "reliability" of Github Actions, 2. the experience with "shell scripts that used to work but something has changed and now it's next to impossible to understand where/how").

jeremy-moffitt:
> But, is this really something that important that deserves the added complexity into something as brittle as CI? (brittle as in: 1. the "reliability" of Github Actions, 2. the experience with "shell scripts that used to work but something has changed and now it's next to impossible to understand where/how").

Is there something unique about the suseconnect CI that we should take into consideration here? Code coverage in CI industry standard , and a lot of projects maintain minimum coverage thresholds for code commits (I'm not suggesting we add that at this point). I'd like to understand better what makes the suseconnect CI jobs different, perhaps theres something we can address from a stability perspective.

mssola:
@jeremy-moffitt as I was saying, this was just me "yelling at clouds", as it wouldn't be the first time that I see development blocked by flacky tests or unreliable CI processes. That's why I want to reduce complexity on the CI side as much as possible. But there's nothing intrinsically different in suseconnect's CI compared to other projects. So, if we decide to go this route, and as I was saying, feel free to ignore my comment.

.github/workflows/build-rpm.yml
```diff
@@ -0,0 +1,50 @@
...
+# See: https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/
+# See: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions
+#
+# An alternate would be to set, pull_request_target but this takes the CI code
+# from master removing the ability to change the code in a PR easily.
+#
+# Aditionally, since 2021 pull requests from new contributors will not
+# trigger workflows automatically but will wait for approval from somebody
+# with write access.
+# See: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks
```
mssola: This comment is a copy from the one on the features.yml file, right? Maybe we can remove this comment from both files and write it on the README.md file? What do you think?

.github/workflows/build-rpm.yml
```diff
@@ -0,0 +1,50 @@
...
+  SOURCE: /usr/src/connect-ng
+  ARTIFACT_SOURCE: artifacts
+  REGCODE: ${{ secrets.REGCODE }}
+
+jobs:
+  build-rpm:
+    runs-on: ubuntu-latest
+    container:
+      image: registry.suse.com/bci/golang:1.24-openssl
+      options: --user root --privileged
```
mssola, line 16: Why do we need a privileged container here? Maybe it's worth it to add a small comment here because adding `--user root --privileged` in docker is pretty daunting. In fact, I feel like this was copied from features.yml, so it would be cool to clarify why it was needed there. Maybe with the move to split `build-rpm` this is no longer needed there?

build/ci/lib.bash
```diff
@@ -0,0 +1,135 @@
...
+install_locally_built_suseconnect()
+{
+    local out_dir="$(readlink -e ${1})" src_dir bin_tools zypp_cmds rc_links systemd_libdir ruby_libdir
+
+    src_dir="$(dirname "${out_dir}")"
+    bin_tools=( suseconnect suse-uptime-tracker suseconnect-mcp )
+    zypp_cmds=( zypper-migration zypper-search-packages )
+    svc_cmds=( suseconnect-keepalive suse-uptime-tracker )
+    systemd_libdir=/usr/lib/systemd/system
+    ruby_libdir=/usr/lib64/ruby/vendor_ruby/3.4.0
```
mssola: Don't we have the ruby version in some environment variable or something like that? I'd appreciate not hard-coding ruby's version here just in case.

Makefile
```diff
@@ -1,66 +1,101 @@
...
-		exit 1; \
-	fi
-	@if $(if $(filter true,$(strip $(COVERAGE))),true,false); then \
-		$(GO) tool covdata $(1) -i=$(COVERAGE_UNIT); \
-	fi
+COVERAGE_DIR     = $(PWD)/coverage
+COVERAGE_MERGED  = $(COVERAGE_DIR)/merged
+COVERAGE_UNIT    = $(COVERAGE_DIR)/unit
+COVERAGE_FEATURE = $(COVERAGE_DIR)/feature
+# these directories aren't being used yet
```
mssola: Then, could we not have them until we add coverage support for these things? 😄

.github/workflows/build-rpm.yml
rtamalin: So those settings are inherited from the original CI workflow, and weren't necessarily documented as to why they were needed. ;-)

If I had to guess I think the `--user root` may be needed in some cases because the default user in some containers is not root, and some of our scripting and tooling assumes the user is root...

As for the `--privileged`, I believe it may be needed in some cases to permit docker in docker type operations, which the yast test run depends upon, as well as access to some system resources, which may be required by the feature test runs.

I have created SCC-847 to investigate and remediate this as a future work item.

rtamalin: As for the move to split the build-rpm out, that is required because otherwise, if we want to enable coverage collection, we need to modify the RPM build to build with coverage collection enabled, like I did in my initial hacking PR#375 where I explored feature test coverage collection.

It was decided that we probably don't want to have to modify the spec file, or have to maintain a patch to modify the spec file, for this purpose.

build/ci/lib.bash
rtamalin: Currently we don't have the ruby version stored anywhere, as the only place it is used is in the spec file, via the `%{rb_ver}` macro value...

And you make a good point as the above is the SLE 16 stream ruby version whereas the SLE 15 stream one would be different... I'll see about setting the value dynamically based upon from the target system environment.

rtamalin:
> @jeremy-moffitt as I was saying, this was just me "yelling at clouds", as it wouldn't be the first time that I see development blocked by flacky tests or unreliable CI processes. That's why I want to reduce complexity on the CI side as much as possible. But there's nothing intrinsically different in suseconnect's CI compared to other projects. So, if we decide to go this route, and as I was saying, feel free to ignore my comment.

The primary goal of this PR is non-CI feature test coverage collection so that we can aggregate it to get an overall code base coverage figure.

However, the existing local developer feature testing and the CI feature testing relied on the same process, so reworking things was required to support the coverage collection anyway... And the feedback from my earlier investigations via PR#375 was to avoid modifying the RPM build process, hence the re-working of the feature test env setup to not rely on the RPM build. And the recently added Agama testing leverages a similar test env setup, so simplifying things to only implement it once, and share that implementation between multiple testing workflows, for both local developer testing and CI workflows, represents an application of DRY principles in combination with KISS.

And there is still benefit to having CI verification testing that our spec file successfully builds an installable package.

Makefile
rtamalin: Yup they can be removed, as can the corresponding entries in the go-tool-covdata-merge tool below...

Will add a comment there below to remind that enabling additional targeted coverage collection will require adding the additional directories to the -i=... flag value.

.github/workflows/build-rpm.yml
rtamalin: Moved it and other instances to a README.md

build/ci/lib.bash
rtamalin: I made the determination of the ruby_version dynamic

Makefile
rtamalin: Done

.github/workflows/build-rpm.yml
mssola: > I have created SCC-847 to investigate and remediate this as a future work item.

Thanks!

mssola:
> The primary goal of this PR is non-CI feature test coverage collection so that we can aggregate it to get an overall code base coverage figure.
>
> However, the existing local developer feature testing and the CI feature testing relied on the same process, so reworking things was required to support the coverage collection anyway... And the feedback from my earlier investigations via PR#375 was to avoid modifying the RPM build process, hence the re-working of the feature test env setup to not rely on the RPM build. And the recently added Agama testing leverages a similar test env setup, so simplifying things to only implement it once, and share that implementation between multiple testing workflows, for both local developer testing and CI workflows, represents an application of DRY principles in combination with KISS.
>
> And there is still benefit to having CI verification testing that our spec file successfully builds an installable package.

Thanks for the detailed explanation, much appreciated!

mssola (commented):
All looks good code-wise. I've just tested it and it works like a charm. One final comment from my end. Is it possible for the code coverage report as given in stdout to be sorted by percentage? If that's just too much of a hassle don't worry about it, but otherwise I believe it will make more obvious which areas are more covered than others.

Thanks for your work!

mssola:
> I've just tested it and it works like a charm

Different environment, different rules 😅  I ran it on my laptop and I ignored feature tests failures as I did not have a proper environment there. Now on my workstation I got build errors for good reasons.

Mainly, the docker image as built after the RPM build, needs to have /usr/src/connect-ng as a trusted source. I believe this used to work, but I don't know how nor why. I guess the re-ordering of make targets missed this part somehow. Could you take a deeper look at this? I've checked that with the patch below it works, but it's an utter hack so don't upstream it 😄 

```patch
From 2a7b315d58d41032ac0445c9d910e3f38b89889b Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Miquel=20Sabat=C3=A9=20Sol=C3=A0?= <msabate@suse.com>
Date: Wed, 1 Jul 2026 11:47:41 +0200
Subject: [PATCH] Loosen permissions on /usr/src/connect-ng on feature tests
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

After the changes on SUSE/connect-ng#403, the 'feature-tests' make
target cannot be run unless the '/usr/src/connect-ng' directory is not
fully trusted by git.

Signed-off-by: Miquel Sabaté Solà <msabate@suse.com>
---
 Makefile | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/Makefile b/Makefile
index ded4ffffb60c..ad78d8d8226c 100644
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ build-rpm:
 	$(CRM) $(MOUNT) -w $(WORKDIR) $(GOCONTAINER) bash -c 'build/ci/build-rpm'

 feature-tests: coverage-dirs
-	$(CRM) $(MOUNT) --env-file $(ENVFILE) -w $(WORKDIR) $(GOCONTAINER) bash -c 'make vendor build COVERAGE_BIN=true && build/ci/run-feature-tests'
+	$(CRM) $(MOUNT) --env-file $(ENVFILE) -w $(WORKDIR) $(GOCONTAINER) bash -c 'git config --global --add safe.directory /usr/src/connect-ng && make vendor build COVERAGE_BIN=true && build/ci/run-feature-tests'
 	$(MAKE) fix-ownership
 	$(call go-tool-covdata-report,percent,$(COVERAGE_FEATURE))
 	$(call go-tool-covdata-report,func,$(COVERAGE_FEATURE))
@@ -222,4 +222,4 @@ fix-ownership:
 	@for o in artifacts coverage internal/connect/version.txt out testdata/agama_srcs vendor; do \
 		[ -e $${o} ] || continue; \
 		sudo chown -R $$(id -u):$$(id -g) $${o}; \
-	done
\ No newline at end of file
+	done
--
2.54.0
```

rtamalin:
The code coverage report is what is produced by the go tool covdata {func|percent} and it doesn't appear to come with any native sorting tools, so an output filter would be required.

That can be done in a follow on PR; I'd like to get this one in ;-)

> > I've just tested it and it works like a charm
> 
> Different environment, different rules 😅 I ran it on my laptop and I ignored feature tests failures as I did not have a proper environment there. Now on my workstation I got build errors for good reasons.
> 
> Mainly, the docker image as built after the RPM build, needs to have /usr/src/connect-ng as a trusted source. I believe this used to work, but I don't know how nor why. I guess the re-ordering of make targets missed this part somehow. Could you take a deeper look at this? I've checked that with the patch below it works, but it's an utter hack so don't upstream it 😄

This one has been elusive and I thought I'd dealt with in but I guess I missed this on... It also impacts git worktree envs differently that git clone envs. There is no negative impact to including it, so I will add it to the code now.

rtamalin:
Updated to address review feedback and git trusted directory issues and Leap 16.0 test-yast binary object compatibility issues and add README.md section Leap 16.0 act docker-in-docker permissions issues.

jeremy-moffitt (approved):

PR #404, opened 2026-06-25
Collect metrics to report that requests came in from the mcp-server

chajain:
https://jira.suse.com/browse/SCC-784

Investigation Spike - Telemetry/instrumentation of MCP-connect

To collect metrics/statistics about how customers are using features in the suseconnect mcp server

- Added a function module in susconnect_mcp.go which creates a suseconnectmcp file in /var/lib/suseconnect-mcp folder and stores mcp action name and count of how many times it is called in a struct .
            type Action struct {
                        Name  string `json:"name"`
                        Count int    `json:"count"`
            }

- Added a new profile mcp_stats which converts the data blob in file /var/lib/suseconnect-mcp/suseconnectmcp to sha256 and stores it in /run/suseconnect/mcp-profile-id

**How to Test**

- Set up the susseconnect-mcp server using the instructions in merged PR https://github.com/SUSE/connect-ng/pull/351

- use the tools for example to answer "What is the registration status of my system" and other Actions. This should create file suseconnectmcp in /var/lib/suseconnect-mcp

- Clone happy-customer and pull PR https://github.com/SUSE/happy-customer/pull/9241 . Start the glue server.
- Now when you execute keepalive command it should create file /run/suseconnect/mcp-profile-id which has the checksum of the data blob stored in /var/lib/suseconnect-mcp/suseconnectmcp

### DIFF
--- a/cmd/suseconnect-mcp/suseconnect_mcp.go
+++ b/cmd/suseconnect-mcp/suseconnect_mcp.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -11,6 +12,36 @@ import (
 	"github.com/SUSE/connect-ng/pkg/registration"
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 )
+
+// McpActions handles the file system operations for MCP actions.
+type McpActions struct {
+	actions     []Action
+	osStat      func(string) (os.FileInfo, error)
+	osMkdirAll  func(string, os.FileMode) error
+	osReadFile  func(string) ([]byte, error)
+	osWriteFile func(string, []byte, os.FileMode) error
+	osOpenFile  func(string, int, os.FileMode) (*os.File, error)
+	osRename    func(string, string) error
+	osRemove    func(string) error
+}
+
+// NewMcpActions creates a new McpActions with default file system functions.
+func NewMcpActions() *McpActions {
+	return &McpActions{
+		osStat:      os.Stat,
+		osMkdirAll:  os.MkdirAll,
+		osReadFile:  os.ReadFile,
+		osWriteFile: os.WriteFile,
+		osOpenFile:  os.OpenFile,
+		osRename:    os.Rename,
+		osRemove:    os.Remove,
+	}
+}
+
+type Action struct {
+	Name  string `json:"name"`
+	Count int    `json:"count"`
+}
 
 type ToolInput struct{}
 
@@ -43,9 +74,95 @@ var (
 	_ func(context.Context, *mcp.CallToolRequest, DeactivateInput) (*mcp.CallToolResult, JSONOutput, error) = DeactivateProduct
 )
 
+var mcpDir = "/var/lib/suseconnect-mcp"
+
+// incCount increments the count of a specific action.
+func (m *McpActions) incCount(actionName string) {
+	found := false
+	for i, action := range m.actions {
+		if action.Name == actionName {
+			m.actions[i].Count++
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		m.actions = append(m.actions, Action{Name: actionName, Count: 1})
+	}
+}
+
+func updateActionCountInit(actionName string) error {
+	m := NewMcpActions()
+	return updateActionCount(m, actionName)
+}
+
+func updateActionCount(m *McpActions, actionName string) error {
+	filePath := mcpDir + "/suseconnectmcp"
+
+	if _, err := m.osStat(mcpDir); os.IsNotExist(err) {
+		err = m.osMkdirAll(mcpDir, 0755)
+		if err != nil {
+			return fmt.Errorf("failed to create directory: %w", err)
+		}
+
+	}
+
+	// Ensure that the process has write access to the directory, otherwise fail early
+	testFile := mcpDir + "/.write_test"
+	f, openErr := m.osOpenFile(testFile, os.O_WRONLY|os.O_CREATE, 0664)
+	if openErr != nil {
+		return fmt.Errorf("no write access to %s: %w", mcpDir, openErr)
+	}
+	f.Close()
+
+	err := m.osRemove(testFile)
+	if err != nil {
+		// This shouldn't fail, but if it does, it's not critical
+		slog.Warn("failed to remove temporary file", "file", testFile, "error", err)
+	}
+
+	file, err := m.osReadFile(filePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("failed to read file: %w", err)
+		}
+	} else {
+		err = json.Unmarshal(file, &m.actions)
+		if err != nil {
+			// Can be empty file
+		}
+	}
+
+	m.incCount(actionName)
+
+	data, err := json.Marshal(m.actions)
+	if err != nil {
+		return fmt.Errorf("failed to marshal json: %w", err)
+	}
+
+	tempFilePath := fmt.Sprintf("%s.%d.tmp", filePath, os.Getpid())
+	err = m.osWriteFile(tempFilePath, data, 0664)
+	if err != nil {
+		return fmt.Errorf("failed to write to temporary file: %w", err)
+	}
+
+	err = m.osRename(tempFilePath, filePath)
+	if err != nil {
+		m.osRemove(tempFilePath)
+		return fmt.Errorf("failed to rename temporary file: %w", err)
+	}
+
+	return nil
+}
+
 func RegistrationStatus(ctx context.Context, req *mcp.CallToolRequest, input ToolInput) (
 	*mcp.CallToolResult, JSONOutput, error) {
 	slog.Info("RegistrationStatus tool called")
+
+	if err := updateActionCountInit("RegistrationStatus"); err != nil {
+		return nil, JSONOutput{Error: "Failed to update Action Count"}, err
+	}
 
 	opts, err := connect.ReadFromConfiguration(connect.DefaultConfigPath)
 	if err != nil {
@@ -63,6 +180,10 @@ func RegistrationStatus(ctx context.Context, req *mcp.CallToolRequest, input Too
 func ListExtensions(ctx context.Context, req *mcp.CallToolRequest, input ToolInput) (
 	*mcp.CallToolResult, JSONOutput, error) {
 	slog.Info("ListExtensions tool called")
+
+	if err := updateActionCountInit("ListExtensions"); err != nil {
+		return nil, JSONOutput{Error: "Failed to update Action Count"}, err
+	}
 
 	opts, err := connect.ReadFromConfiguration(connect.DefaultConfigPath)
 	if err != nil {
@@ -85,6 +206,10 @@ func RegisterSystem(ctx context.Context, req *mcp.CallToolRequest, input Registe
 	*mcp.CallToolResult, JSONOutput, error) {
 	slog.Info("RegisterSystem tool called")
 
+	if err := updateActionCountInit("RegisterSystem"); err != nil {
+		return nil, JSONOutput{Error: "Failed to update Action Count"}, err
+	}
+
 	opts, err := connect.ReadFromConfiguration(connect.DefaultConfigPath)
 	if err != nil {
 		return nil, JSONOutput{Error: "Failed to read SUSEConnect configuration"}, err
@@ -104,6 +229,10 @@ func RegisterSystem(ctx context.Context, req *mcp.CallToolRequest, input Registe
 func ActivateProduct(ctx context.Context, req *mcp.CallToolRequest, input ActivateInput) (
 	*mcp.CallToolResult, JSONOutput, error) {
 	slog.Info("ActivateProduct tool called")
+
+	if err := updateActionCountInit("ActivateProduct"); err != nil {
+		return nil, JSONOutput{Error: "Failed to update Action Count"}, err
+	}
 
 	opts, err := connect.ReadFromConfiguration(connect.DefaultConfigPath)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cmd/suseconnect-mcp/suseconnect_mcp.go
```diff
@@ -43,10 +50,64 @@ var (
...
+	if !found {
+		actions = append(actions, Action{Name: actionName, Count: 1})
+	}
+
+	data, err := json.Marshal(actions)
+	if err != nil {
+		return fmt.Errorf("failed to marshal json: %w", err)
+	}
+
+	err = os.WriteFile(filePath, data, 0644)
```
rtamalin: So this will attempt to overwrite the existing file so if anything goes wrong then potentially the original action data and the updated action data is corrupted/lost...

I would recommend using one of the following patterns here:

1. Write the updated content out to a temporary file in the same directory, and if that succeeds perform a rename of the temporary file to the original file name, which is generally an "atomic" operation from a filesystem perspective.
2. Make a temporary backup copy of the original file to a new file in the same directory, and write the updated content to the original path, and if that succeeds, remove the backup, otherwise move the backup back to the original path
3. Move the original to a temporary backup path, write the updated content to the original path, and if that succeeds, remove the backup, otherwise restore the backup.

NOTES:
* (2) and (3) are almost the same just a question of whether you are copying or moving the original to make a backup
* (3) like (1) requires you to ensure that you create the new file holding the updated content with the same user/group ownership and permissions as the original, whereas (2) has you overwriting the original file so ownership and permissions are already correct
* in all cases you should ensure that the temporary/backup files are created with equivalent ownership and permissions

Then overall goal is to minimise the risk of data loss...

cmd/suseconnect-mcp/suseconnect_mcp.go
```diff
@@ -43,10 +50,64 @@ var (
...
+	dir := profiles.GetProfileFilePath()
+	filePath := dir + "/suseconnectmcp"
+
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		err = os.MkdirAll(dir, 0755)
+		if err != nil {
+			return fmt.Errorf("failed to create directory: %w", err)
+		}
+	}
+
```
rtamalin, line 124: Extremely low priority suggestion:

Since the intent is to update the target profile data file, it may be worth doing a check at this point to ensure the process has write access to the target path before trying much else, using either the general purpose os.OpenFile(path, os.WR_ONLY, ...) or Linux specific syscall.Access() check approaches...

cmd/suseconnect-mcp/suseconnect_mcp.go
```diff
@@ -43,10 +50,64 @@ var (
 	_ func(context.Context, *mcp.CallToolRequest, DeactivateInput) (*mcp.CallToolResult, JSONOutput, error) = DeactivateProduct
 )
 
+func updateActionCount(actionName string) error {
```
rtamalin: @digitaltom Can multiple suseconnect-mcp instances be running at the same time? I ask because this updateActionCount() is called to update the profile data on every MCP action, and if multiple instances can run in parallel we likely will need to add some locking mechanism to avoid colliding updates...

internal/collectors/config.go
```diff
@@ -101,12 +101,15 @@ var collectorsRegistry = map[string]CollectorRegistryEntry{
 		Metadata:  CollectorMetadata{DefaultEnabled: true, Mandatory: true, Type: SystemInfoCollector},
 		Collector: HA{},
 	},
-
 	// Optional collectors
 	"pci_data": {
 		Metadata:         CollectorMetadata{DefaultEnabled: true, Mandatory: false, Type: ProfileCollector},
 		CollectorFactory: func(updateDataIDs bool) Collector { return PCI{UpdateDataIDs: updateDataIDs} },
 	},
+	"mcp_list": {
```
rtamalin: `mcp_list` seems a bit on the vague side... Possibly `mcp_stat` or `mcp_stats` might be a bit more discriptive? Or `mcp_actions`?

Note that the name of a profile type doesn't have to be limited to 8 characters... It was just that the first 2 profiles happened to have short names...

cmd/suseconnect-mcp/suseconnect_mcp.go
```diff
@@ -43,10 +50,64 @@ var (
...
+	dir := profiles.GetProfileFilePath()
```
rtamalin: BTW I'm not sure that it is a requirement to store the actual persistent mcp stats at this location... It is intended as the location for storing the profile id cache files, which track the previously reported profile ids to know whether or not a client needs to include the profile data field in submissions...

Potentially consider using `/var/lib/suseconnect-mcp` or `/var/cache/suseconnect-mcp` if you want a path that is safe to write to that will not be lost across reboots, and not cause clutter in transactional-update snapshots.

internal/collectors/mcp.go
```diff
@@ -0,0 +1,45 @@
...
+	err := json.Unmarshal(inputStream, &actions)
+	if err != nil {
+		return nil, err
+	}
+	return actions, nil
+}
+
+func (mcp MCP) run(arch string) (Result, error) {
+	util.Debug.Print("mcp.UpdateDataIDs: ", mcp.UpdateDataIDs)
+	mcpPath := profiles.GetProfileFilePath() + "suseconnectmcp"
```
rtamalin: See above comments about why the persistent data doesn't need to be stored in the Profile ID cache path...

internal/collectors/mcp_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/SUSE/connect-ng/pkg/profiles"
+	"github.com/stretchr/testify/assert"
+)
+
+var mcpDataBlob profiles.Profile
+var mcpTestData string
+var mcpPath string
+
+func setupMcpTestData() {
+	testProfilePath, _ := os.MkdirTemp("/tmp/", "__suseconnect")
```
rtamalin: It would be better to use a tempDir randomised path, rather than a fixed path here to avoid issues of inheriting stale data from previous tests or causing problems if two test incarnations are run in parallel...

cmd/suseconnect-mcp/suseconnect_mcp.go
```diff
@@ -43,10 +50,64 @@ var (
...
+	filePath := dir + "/suseconnectmcp"
+
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		err = os.MkdirAll(dir, 0755)
+		if err != nil {
+			return fmt.Errorf("failed to create directory: %w", err)
+		}
+	}
+
+	var actions []Action
```
rtamalin: Note that you could potentially define type to represent the []Action slice of actions, and implement receiver methods on it to wrap the file system actions, potentially with non-public variables for the `os` module or the methods called under it that would allow for testing of this code by mocking the `os` actions...

cmd/suseconnect-mcp/suseconnect_mcp.go
digitaltom: It's just an executable, so in theory it can get called in parallel. But I wouldn't add complexity due to that edge case and be fine by not counting the parallel call.

internal/collectors/mcp.go
chajain: Is it ok to store the id file mcp-profile-id in /run/suseconnect sine we are using BuildProfile function module from github.com/SUSE/connect-ng/pkg/profiles to generate this file ?

cmd/suseconnect-mcp/suseconnect_mcp.go
rtamalin: I was thinking more about avoiding corruption risks if two processes attempt to write to the same file at the same time, but the above suggestions for safer file updating should minimise that risk ;-)

internal/collectors/mcp.go
rtamalin: Yeah that is the intended purpose of the profiles.GetProfileFilePath() location, namely to hold the cached profile ids, but not the actual profile data.

cmd/suseconnect-mcp/suseconnect_mcp_test.go
```diff
@@ -76,3 +84,210 @@ func TestToolExecutionWithInvalidInputs(t *testing.T) {
...
+	for _, action := range m.actions {
+		if action.Name == "action1" {
+			assert.Equal(t, 2, action.Count)
+		}
+	}
+}
+
+// updateActionCountHelper is a test helper that mimics the structure of the original
+// updateActionCount but allows for injecting a mockable McpActions object.
+func updateActionCountHelper(m *McpActions, actionName, dir, filePath string) error {
```
rtamalin: So this "helper" doesn't really help us get coverage on the updateActionCount() routine itself...

A better approach might be the following split the existing updateActionCount() routine up into an outer and inner part...

The inner part implements the logic to update the mcp actions, and takes as argument a *McpActions, while the outer part is a thin wrapper to create/init the McpActions object, that then calls the inner part with a pointer to McpAction object...

Then your testing can call that inner function with a McpActions obhject that has initialised with appropriate mock versions of the os routines.

This would allow you to get actual coverage of the inner part of the updateActionCount() routine.

You could even take this a step further and have the updateActionCount be a var pointing to an implementation of the updateActionCount() routine, that gets initialised with the normal implementation of that function. And then in your testing you can temporarily replace that implementation with a customised testing version, that sets up your desired mocking for the test scenario before calling the "inner" function, so that you, in your testing, you can either call updateActionCount() as normal, or call routines that call it, to exercise those code paths...

There are examples of this approach to a variable holding a helper routine such as [util.Execute](internal/util/system.go) which is then mocked in a number of different ways with varying levels of complexity. ;-)  You actually use this below for the mcpOsReadFile method overriding in the collector testing.

cmd/suseconnect-mcp/suseconnect_mcp_test.go
```diff
@@ -76,3 +84,210 @@ func TestToolExecutionWithInvalidInputs(t *testing.T) {
...
+	err = m.osRename(tempFilePath, filePath)
+	if err != nil {
+		m.osRemove(tempFilePath)
+		return fmt.Errorf("failed to rename temporary file: %w", err)
+	}
+
+	return nil
+}
+
+func TestUpdateActionCount(t *testing.T) {
```
rtamalin, line 276: This test function is a bit big; it might be better to restructure it so that the subtests are top level functions that are invoked via t.Run() wrapping function, which take as parameters any shared values or resources...

cmd/suseconnect-mcp/suseconnect_mcp.go
```diff
@@ -43,10 +72,91 @@ var (
...
+	}
+
+	m.incCount(actionName)
+
+	data, err := json.Marshal(m.actions)
+	if err != nil {
+		return fmt.Errorf("failed to marshal json: %w", err)
+	}
+
+	tempFilePath := filePath + ".tmp"
```
rtamalin: This is an improvement as we only replace the existing content if we succeeded in writing to the tmp file... However the tmpfile name is fixed, so if we did have 2 different processes running at the same time, attempting to update the actions file at the same time, you could still see corruption issues if both try to write to the file at the same time... It would be good to introduce a differentiating factor into the constructed tmp file name, such as the pid, to avoid that risk...

Also you have an implicit assumption that the newly created file will have the same ownership (uid/gid) as the original... This is very likely to be the case, but if ever this file needs to be setup with specific ownership that may not match the euid/egid of the mcp process, your new file would have different ownership.

cmd/suseconnect-mcp/suseconnect_mcp.go
rtamalin: This is mostly resolved - see my comments below on the new code with regard to ownership preservation...

cmd/suseconnect-mcp/suseconnect_mcp.go
```diff
@@ -43,10 +72,91 @@ var (
...
+		f, openErr := os.OpenFile(testFile, os.O_WRONLY|os.O_CREATE, 0664)
+		if openErr != nil {
+			return fmt.Errorf("no write access to %s: %w", mcpDir, openErr)
+		}
+		f.Close()
+
+		err = m.osRemove(testFile)
+		if err != nil {
+			// This shouldn't fail, but if it does, it's not critical
+			slog.Warn("failed to remove temporary file", "file", testFile, "error", err)
```
brett060102: This test should be outside of the directory create. Since it possible for the directory to exist on routine entrance, but not be readable or writeable.

cmd/suseconnect-mcp/suseconnect_mcp.go
brett060102: The change made for this comment looks good. So resolving this conversation.

cmd/suseconnect-mcp/suseconnect_mcp.go
brett060102: issues here are covered below. so resolving this conversation.

cmd/suseconnect-mcp/suseconnect_mcp.go
brett060102: code was added, but don't think it was added in the right place. comment on that below. So, I think this issue cab be resolved.

cmd/suseconnect-mcp/suseconnect_mcp_test.go
brett060102: It looks like this was done. Though a little hard to tell without access to the starting point code. But looks good to me so resolving this.

cmd/suseconnect-mcp/suseconnect_mcp.go
brett060102: changes have been made. I noted below that the test need to be done outside of directory create. bit resolving this conversation.

cmd/suseconnect-mcp/suseconnect_mcp_test.go
brett060102: the most recent changes addressed this issue. At least I think they do what Fergal wanted.

cmd/suseconnect-mcp/suseconnect_mcp.go
brett060102: most recent changes addressed this issue.

brett060102 (approved):
LGTM

PR #407, opened 2026-07-01
Build package with Go 1.26 toolchain

rtamalin:
The SUSE/connect-ng project continues to be based upon the Go 1.24 language version, so the version in the go.mod file is not being changed.

Update the build/packaging/suseconnect-ng.spec BuildRequires to use go1.26-openssl as the build toolchain to use when building.

Update all SLE BCI Golang container versions to use go1.26-openssl.

Add a build/packaging/suseconnect-ng.changes entry in the v1.23 block to track this change.

### DIFF
--- a/.github/workflows/agama.yml
+++ b/.github/workflows/agama.yml
@@ -17,7 +17,7 @@ jobs:
     name: Build SUSE/connect-ng inputs required by Agama
     runs-on: ubuntu-latest
     container:
-      image: registry.suse.com/bci/golang:1.24-openssl
+      image: registry.suse.com/bci/golang:1.26-openssl
       options: --user root --privileged
 
     steps:
--- a/.github/workflows/build-rpm.yml
+++ b/.github/workflows/build-rpm.yml
@@ -12,7 +12,7 @@ jobs:
   build-rpm:
     runs-on: ubuntu-latest
     container:
-      image: registry.suse.com/bci/golang:1.24-openssl
+      image: registry.suse.com/bci/golang:1.26-openssl
       options: --user root --privileged
 
     steps:
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -15,7 +15,7 @@ jobs:
   feature-tests:
     runs-on: ubuntu-latest
     container:
-      image: registry.suse.com/bci/golang:1.24-openssl
+      image: registry.suse.com/bci/golang:1.26-openssl
       options: --user root --privileged
 
     steps:
--- a/.github/workflows/lint-unit.yml
+++ b/.github/workflows/lint-unit.yml
@@ -6,7 +6,7 @@ jobs:
   unit-tests:
     runs-on: ubuntu-latest
     container:
-      image: registry.suse.com/bci/golang:1.24-openssl
+      image: registry.suse.com/bci/golang:1.26-openssl
 
     steps:
       - uses: actions/checkout@v4
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ GOFLAGS       = -v -mod=vendor
 BINFLAGS      = -buildmode=pie
 SOFLAGS       = -buildmode=c-shared
 
-GOCONTAINER   = registry.suse.com/bci/golang:1.24-openssl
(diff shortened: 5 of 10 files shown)


rtamalin:
See https://build.suse.de/package/show/home:fmccarthy:suseconnect/suseconnect-ng-test for a test build environment.
I have validated that a go1.26-openssl or go1.26 is available for use building the packages for all active release streams using osc buildinfo, but the build service is running slow.

brett060102 (approved):
Looks good to me:
Review process:
Inspect changes with respect to description. Change match stated work.
Checkout associated branch and do:
$ make  build
$ make build-rpm
$ make feature-tests
While this repeats what is done in CI, it does so in a different context.
Everything passed.

mssola:
Better yet, could we still have 1.24 somewhere so we make a smoke test that building keeps on working on that version? I'm thinking maybe in unit tests to have a matrix between 1.24 and whatever next version we want (1.26 now). What do you think? Otherwise looks good, thanks!

mallardduck:
Hey all,

I noticed this PR looking to bump go to `1.26` and wanted to raise a concern and propose an alternative. As @felixsch and I chatted about eventually (soon ™️) the ECM group plans to send some PRs related to ECM integration. The short version that we will take re-usable parts of `scc-operator` for Rancher and maintain them here as a sub-module.

**The Concern**: Since [1.21](https://go.dev/doc/go1.21) the `go` directive is a strict runtime constraint - specifically a strict minimum version. Bumping the core library to Go 1.26 immediately locks out any downstream projects that are still bound to 1.25 (or 1.24, but that's EOL).

**The Alternative**: Start using a multi-module repo layout.
- Bump the root `go.mod` to 1.25 so that `/pkg` and `/internal` remain flexible for downstream use,
- Create a new `go.mod` under `cmd` pinned to 1.26 with a local `replace` back to root for `connect-ng`

This would give us the best of both worlds without restricting downstream consumers into a forced toolchain upgrade. Thoughts?

rtamalin:
> The SUSE/connect-ng project continues to be based upon the Go 1.24 language version, so the version in the `go.mod` file is not being changed.

@mallardduck @mssola Just to be clear, as the first line of the description says, the intent of this change is purely a switch of the Go toolchain version being used to build the suseconnect-ng packages, specifically to move forward to using the Go 1.26-openssl toolchain when building for Leap and SLE, or vanilla Go 1.26 when building for Tumbleweed, though the latter automatically moves forward to newer Go API versions as they become available.

There is no plan to touch or change the `go 1.24` version specification in the go.mod as part of this work.

This work is being done at the request of the SUSE Security Team to move the packages forward to being built with a Go toolchain version that isn't EOL. Nothing more.

The recent switch of the `go.mod` Go version to 1.24 was purely to enable access to core library functionality that became available from 1.23 onwards, and I chose 1.24 to align with the targeted Go toolchain version at that time... Currently I'm not aware of any need to move the version in `go.mod` forward beyond 1.24 right now.

Going forward we can definitely coordinate any `go.mod` Go versioning changes, but with respect to things change, I believe we should be good?

rtamalin:
> Better yet, could we still have 1.24 somewhere so we make a smoke test that building keeps on working on that version? I'm thinking maybe in unit tests to have a matrix between 1.24 and whatever next version we want (1.26 now). What do you think? Otherwise looks good, thanks!

@mssola Just to clarify you want to have some tests to continue to use the go1.24-openssl container images rather than the go1.26-openssl container images? Because as already discussed in the other comment, and as the first line of the description says, we are not changing the `go 1.24` line in the `go.mod` file...

I'm not sure that I see a whole lot of benefit to doing so, but it can be done for the CI workflows at least if we leverage a matrix scheme to control which container image versions are used... However, that would likely add complexity and increased risk of CI job flakiness to the CI, which seems counter to your recent concerns expressed in other PRs about adding CI job complexity?

mallardduck:
> There is no plan to touch or change the `go 1.24` version specification in the go.mod as part of this work.

😮‍💨 Sorry - totally mush brain over here. I overlooked those details and did not check the changed files. Ignore my concerns there as they don't apply here.

mssola:
@rtamalin understood and fine by me too.

mssola (approved):

PR #412, opened 2026-07-10
zypper-migration support interactive package installation

brett060102:
zypper-migration while supporting the --non-interactive option, it defaults to interactive mode for number of zypper operations including dist-upgrade. However hardcoded "--non-interactive" when doing package installations.

Change code to support the interactive/non-interactive state in zypper.InstallReleasePackage.

Change is to make zypper.InstallReleasePackage handle  interactive/non-interactive in the same way as zypper. DistUpgrade. Change only effects use through zypper-migration, where the interactive state defaults to true. 
Since suseconnect does not provide a  --non-interactive option, calls from in internal/connect/client.go pass in nonInteractive==true to preserve existing behavior.

To test:
1: create Leap 15.6/16.0/16.1 VM.
2: build suseconnect rpms
3: copy suseconnect-ng and libsuseconnect rpms from ./artifacts
4: on VM:
4.1 install new suseconnect-ng and libsuseconnect (need to use zypper so zypper-migration is correctly installed
4.2: register the VM (sudo suseconnect -r REG_CODE)
4.3 sudo zypper migration
there will be a number of user prompts. notably:
```
Checking for file conflicts: [..error]
Detected 1 file conflict:

File /usr/lib/rpm/gnupg/keys/gpg-pubkey-287a0027-682477e3.asc
  from install of
     PackageHub-release-16.1-160099.11.1.x86_64 (PackageHub-16.1)
  conflicts with file from package
     openSUSE-build-key-1.0-lp161.3.1.x86_64 (@System)

File conflicts happen when two packages attempt to install files with the same name but different contents. If you continue, conflicting files will be replaced losing the previous content.
Continue? [yes/no] (no): 

```
answer yes will allow migration to proceed, no will block the package install and terminate the migration.
This prompt is the change in behavior from the existing code to this code. The existing code will output the message and terminate the migration. The new behavior is consistent with the 15.x "yast migration" behavior.

### DIFF
--- a/cmd/zypper-migration/migration.go
+++ b/cmd/zypper-migration/migration.go
@@ -77,6 +77,7 @@ func main() {
 		repo                     multiArg
 		download                 multiArg // using multiArg here to make flags simpler to visit
 		autoImportRepoKeys       bool
+		echo                     bool
 	)
 
 	flag.Usage = func() {
@@ -172,7 +173,9 @@ func main() {
 		// reset root (if set) as the update stack can be outside of
 		// the to be updated system
 		opts.FsRoot = ""
-		echo := util.SetSystemEcho(true)
+		if !nonInteractive {
+			echo = util.SetSystemEcho(true)
+		}
 		if pending, err := zypper.PatchCheck(true, quiet, verbose, nonInteractive, false); err != nil {
 			fmt.Printf("patch pre-check failed: %v\n", err)
 			os.Exit(1)
@@ -202,7 +205,9 @@ func main() {
 				os.Exit(1)
 			}
 		}
-		util.SetSystemEcho(echo)
+		if !nonInteractive {
+			util.SetSystemEcho(echo)
+		}
 		// restore root if needed
 		if fsRoot != "" {
 			opts.FsRoot = fsRoot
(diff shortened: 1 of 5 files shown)


cmd/zypper-migration/migration.go
```diff
@@ -391,16 +391,17 @@ func main() {
...
 		return systemProducts, err
 	}
-
 	releasePackageMissing := false
 	for _, p := range systemProducts {
 		// if a release package for registered product is missing -> try install it
-		err := zypper.InstallReleasePackage(p.Identifier, autoImportRepoKeys)
+		echo := util.SetSystemEcho(true)
+		err := zypper.InstallReleasePackage(p.Identifier, autoImportRepoKeys, nonInteractive)
+		util.SetSystemEcho(echo)
```
mssola: Why do we need this echo? Is it just for the interactivity part? If so, could we only do it depending on `nonInteractive`?

mssola (commented):
Just a small question from my side. Good job!

cmd/zypper-migration/migration.go
brett060102: Yes, the echo is only needed for the interactive part. I thought about making that conditional, but that is not what we do for dist-update, so I decided to follow that pattern.

mssola: I´d still have it conditional to `nonInteractive` if that's fine by you.

brett060102: Done and applied to all places using util.SetSystemEcho

mssola (approved):
LGTM 🎉

PR #417, opened 2026-07-22
suseconnect does not allow de-register when subscription is expired

chajain:
https://jira.suse.com/browse/SCC-865
suseconnect does not allow de-register when subscription is expired

**How to Test**

Register a system using a reg code.
In you setup, bring up the psql console and using sql update the expires_at to an earlier date. For example 
-  UPDATE subscriptions SET expires_at = <earlier date> WHERE id = <subscription_id>;
Try to de-register. It should not fail with the fix.

### DIFF
--- a/internal/connect/client.go
+++ b/internal/connect/client.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 
 	cred "github.com/SUSE/connect-ng/internal/credentials"
 	"github.com/SUSE/connect-ng/internal/util"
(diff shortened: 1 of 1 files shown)


internal/connect/client.go
```diff
@@ -237,8 +238,10 @@ func Deregister(api WrappedAPI, opts *Options) error {
 	}
 
 	if !opts.SkipServiceInstall {
-		if err := localRemoveOrRefreshService(baseMeta.Name, opts); err != nil {
-			return err
+		if baseMeta != nil {
```
brett060102: I think these two if should be combined 
if !opts.SkipServiceInstall  && baseMeta != nil {
}

brett060102 (approved):
LGTM

mssola (approved):
It's sad that the error doesn't include which kind of error it is, but this is how things are so we have to manually hack for these strings.

Overall, then LGTM.

That being said. It doesn't have to be on this PR, but could we add an end-to-end test for this? (feature test)

PR #418, opened 2026-07-31
Add public-api-demo support for profiles submission

rtamalin:
A new PROFILE_DIR env var can be used to specify a directory with files named for profile types containing the corresponding profile data as a JSON blob.

Add cmd/public-api-demo/examples/profile with reference examples of pci_data and mod_list profiles.

Added a new RMT_HOST env var that can be used to specify an upstream RMT server url ast the target; doing so will disable any interactions that are not supported by an RMT. This env var setting will supercede any SCC_HOST env var setting.

Modified SCC_HOST env var handling to not disable any interactions with the upstream target.

Also added a README.md to the cmd/public-api-demo providing so basic details of what the tool is and how to use it.


rtamalin:
@felixsch Can you give this a once over next week to confirm that you are happy with the changes to the SCC_HOST handling, and that my README.md explanations are correct...

felixsch:
Nice addition @rtamalin. Lgtm to me

felixsch:
Ideally the idea was to push each functionality into its own file and document it properly as example and assemble them in the main file (then they can be dynamically configured). This would fit better for documentation purposes (rather then a big file with a bunch of lines). But it never happened. In case this is something you have fun and the time: Maybe something to consider?

rtamalin:
> Ideally the idea was to push each functionality into its own file and document it properly as example and assemble them in the main file (then they can be dynamically configured). This would fit better for documentation purposes (rather then a big file with a bunch of lines). But it never happened. In case this is something you have fun and the time: Maybe something to consider?

You might need to expand the explanation of that a bit as I'm not sure I clearly grok your meaning.

However, in principle it sounds like something to work towards... In a follow up PR ;-)

felixsch (approved):

PR #425, opened 2026-08-14
suseconnect-ng status output not consistently shown in requested format (bsc# 1233147)

chajain:
Customer has some tooling in place to query the status of the subscription on their SLE instances, and they expect the output to be consistently shown in json format as indicated on the manpage. 
Currently In case when an error message is output, it is just plain string and not in json format


brett060102 (commented):
build/packaging/suseconnect-ng.changes needs to be updated or this change.

brett060102 (approved):
LGTM

PR #428, opened 2026-08-20
Import GPG keys when installing products

brett060102:
Installing products will fail if one of the required repos requires importing a GPG key

to test:
Requires SLE 15 SP  7 installed wothout workstation extension

1: build suseconnect and copy to VM /tmp/
2: if not registered register system
$ /tmp/suseconnect -r $SLES_SUBSCRIPTION

3: get desktop extension
$/tmp/suseconnect -p sle-module-desktop-applications/15.7/x86_64

$ activate sle-we
suseconnect -p sle-we/15.7/x86_64 -r $SLE_DESKTOP_REGCODE

### DIFF
--- a/internal/zypper/zypper.go
+++ b/internal/zypper/zypper.go
@@ -259,7 +259,7 @@ func InstallReleasePackage(identifier string, autoImportRepoKeys bool, nonIntera
 	if nonInteractive {
 		args = append(args, "--non-interactive")
 	}
-	args = append(args, []string{"install", "--no-recommends", "--auto-agree-with-product-licenses", "-t", "product", identifier}...)
+	args = append(args, []string{"--gpg-auto-import-keys", "install", "--no-recommends", "--auto-agree-with-product-licenses", "-t", "product", identifier}...)
 
 	if autoImportRepoKeys {
 		args = append([]string{"--gpg-auto-import-keys"}, args...)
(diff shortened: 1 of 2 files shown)


mssola (changes requested):
No. Just look at the `if` block just right after what you have changed. You need to tell SUSEConnect (via the `--gpg-auto-import-keys` option) that this option is to be passed to zypper. Otherwise it could potentially be a security issue.

internal/zypper/zypper.go
```diff
@@ -259,7 +259,7 @@ func InstallReleasePackage(identifier string, autoImportRepoKeys bool, nonIntera
 	if nonInteractive {
 		args = append(args, "--non-interactive")
 	}
-	args = append(args, []string{"install", "--no-recommends", "--auto-agree-with-product-licenses", "-t", "product", identifier}...)
+	args = append(args, []string{"--gpg-auto-import-keys", "install", "--no-recommends", "--auto-agree-with-product-licenses", "-t", "product", identifier}...)
 
 	if autoImportRepoKeys {
 		args = append([]string{"--gpg-auto-import-keys"}, args...)
```
rtamalin, line 265: Per @mssola this line here already takes care of ensuring the flag is added before the action in the args list, if the corresponding suseconnect flag is specified...

This is something we had to fix for the DMS tool's use of zypper migration and suseconnect in the past, for the v1.10 release.

rtamalin (changes requested):
Agree with @mssola this is not the right fix

internal/zypper/zypper.go
brett060102: Thank you. I was so focused on the interactive/non-interactive part of this I totally missed that.
I will close this PR.

brett060102:
Change not needed. the suseconnect option "--gpg-auto-import-keys" combined withe change 
https://github.com/SUSE/connect-ng/pull/412 - zypper-migration support interactive package installation
Should be sufficient to address the bugzillas.

PR #429, opened 2026-08-23
Add libsuseconnect error JSON tests

### LINKED ISSUE #362: libsuseconnect errorToJSON() needs unit tests to ensure error handling consistency is maintained
Add unit tests to libsuseconnect.go that ensure that for each supported error type the appropriate JSON response is being generated.

This will help to reduce the risk of regressions going forward.

vsolano9:
## Description

Add table-driven regression coverage for every current `errorToJSON` branch in `libsuseconnect`, including API, timeout, certificate, network, JSON, credentials, and fallback errors. Wire the package into the existing `make test` target so the new coverage runs in CI.

* Related Issue / Ticket / Trello card: Closes #362

## How to test the change

- `go test -count=1 -race ./third_party/libsuseconnect`
- `go vet ./third_party/libsuseconnect`
- `make check-format`
- `make vendor build`
- `go test -coverprofile=libsuseconnect.cover ./third_party/libsuseconnect` reports 100% statement coverage for `errorToJSON`

I also mutation-checked the table locally: changing the API error type mapping made the named API-error case fail, and restoring it returned the suite to green.

The full `make test` target ran every selected package and the new `libsuseconnect` package passed. Its existing `internal/zypper/TestBackupAndRestore` case remains Linux-specific and fails on macOS because the expected parent-directory copy is not created; the upstream Linux job is the authoritative full-suite gate.

## Review

Please check out our [review guidelines](https://github.com/SUSE/scc-docs/blob/master/team/workflow/code_review.md) and get in touch with the author to get a shared understanding of the change.

❤️ Thank you! ❤️

### DIFF
--- a/third_party/libsuseconnect/libsuseconnect_test.go
+++ b/third_party/libsuseconnect/libsuseconnect_test.go
@@ -13,6 +13,8 @@ import (
 	"github.com/SUSE/connect-ng/internal/connect"
 	cred "github.com/SUSE/connect-ng/internal/credentials"
 	"github.com/SUSE/connect-ng/pkg/connection"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 type timeoutError struct{}
@@ -143,14 +145,13 @@ func TestErrorToJSON(t *testing.T) {
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			var got errorResponse
-			if err := json.Unmarshal([]byte(errorToJSON(test.err)), &got); err != nil {
-				t.Fatalf("errorToJSON returned invalid JSON: %v", err)
-			}
+			assert := assert.New(t)
 
-			if got != test.want {
-				t.Fatalf("errorToJSON() = %#v, want %#v", got, test.want)
-			}
+			var got errorResponse
+			err := json.Unmarshal([]byte(errorToJSON(test.err)), &got)
+			require.NoError(t, err, "errorToJSON must return valid JSON")
+
+			assert.Equal(test.want, got)
 		})
 	}
 }
--- a/third_party/libsuseconnect/libsuseconnect_test.go
+++ b/third_party/libsuseconnect/libsuseconnect_test.go
@@ -1,0 +1,156 @@
+package main
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"errors"
+	"net"
+	"net/url"
+	"testing"
+
+	"github.com/SUSE/connect-ng/internal/connect"
+	cred "github.com/SUSE/connect-ng/internal/credentials"
+	"github.com/SUSE/connect-ng/pkg/connection"
+)
+
+type timeoutError struct{}
+
+func (timeoutError) Error() string { return "request timed out" }
+func (timeoutError) Timeout() bool { return true }
+
+type errorResponse struct {
+	ErrType string `json:"err_type"`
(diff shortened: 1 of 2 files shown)


mssola (approved):
LGTM. @rtamalin you were the original reporter. What do you think?

third_party/libsuseconnect/libsuseconnect_test.go
```diff
@@ -0,0 +1,156 @@
...
+			name: "generic error",
+			err:  genericError,
+			want: errorResponse{Message: genericError.Error()},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var got errorResponse
+			if err := json.Unmarshal([]byte(errorToJSON(test.err)), &got); err != nil {
```
rtamalin: Where possible with newer unit tests we are trying to use [stretchr/testify](https://github.com/stretchr/testify) testing patterns, e.g. using assert and/require to check for conditions.

rtamalin (changes requested):
Generally looks good, minor change suggested for the actual testing pattern.

vsolano9:
Converted to the testify pattern at `eea54b0`, thanks — this is the right call for the failure output alone.

The subtest now takes `assert := assert.New(t)`, matching the dominant style across `features/` and `internal/`, and the JSON decode uses `require.NoError`. That one has to halt the subtest rather than continue: a failed unmarshal leaves `got` zero-valued, so an `assert` there would report a struct diff that describes the decode failure as a `errorToJSON` mismatch. The comparison itself is `assert.Equal(test.want, got)`, which prints a field-level diff instead of two `%#v` dumps — that matters most for the SSLError cases where `Data` carries multi-line PEM.

I re-ran mutations after converting, because a rewritten assertion that no longer discriminates is a worse outcome than the pattern nit. Four against `errorToJSON`, each caught by the subtest that owns the behaviour:

| mutation | result |
| --- | --- |
| `UnknownAuthorityError` code 19 → 18 | FAIL `unknown_certificate_authority` |
| `HostnameError` emits no cert PEM | FAIL `certificate_hostname` |
| `certsToPEM` chain order not reversed | FAIL `TLS_certificate_verification` |
| `"JSONError"` → `"JsonError"` | FAIL `JSON_decoding` |

Also rebased onto main (it was one commit behind). `gofmt` and `go vet` clean; `./third_party/libsuseconnect`, `./cmd/suseconnect`, `./cmd/suseconnect-mcp` and every `./pkg/...` package pass.

One disclosure: `internal/zypper` `TestBackupAndRestore` fails on my machine, and it fails identically on unmodified `main` — a path assumption that does not hold under macOS temp directories. Not touched by this change, and I mention it only so a failing local run is not mistaken for something this PR introduced.

@mssola @rtamalin ready for another look.

third_party/libsuseconnect/libsuseconnect_test.go
vsolano9: Done in eea54b05 — the libsuseconnect error-JSON tests now use `assert`/`require` from stretchr/testify (`require.NoError` for the unmarshal, `assert.Equal` for the payload comparison). Re-review requested.

vsolano9:
@rtamalin the testify conversion from your review is in eea54b0 (`assert.New` + `require.NoError` for the JSON decode, `assert.Equal` for the payload). I cannot re-request the review from this account (`RequestReviewsByLogin` is denied). Please take another look when you have a chance.

rtamalin (approved):
Looks good and I have verified that the tests pass locally using act.

PR #431, opened 2026-09-01
Enhance public-api-demo to support PubCloud RMT dev-env testing

rtamalin:
Update the instance_data XML document to include an instance_data tag that specifies a product attribute with a value of "SUSE". This will ensure that a PubCloud enabled RMT dev-env, relying on the default Example Provider in the PubCloud instance_verification engine, is assigned a meaningful instance identifier, e.g. 1234, allowing basic BYOS registration and keepalive update testing to work correctly.

Also update the registration.Register() call includes the recently constructed extraData containing the instance_data value.

Relates: jsc#SCC-822

Signed-off-by: Fergal Mc Carthy

### DIFF
--- a/cmd/public-api-demo/main.go
+++ b/cmd/public-api-demo/main.go
@@ -115,12 +115,13 @@ func runDemo(identifier, version, arch, infoPath, regcode string, systemProfiles
 	}
 
 	extraData := registration.ExtraData{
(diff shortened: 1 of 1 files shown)


felixsch (approved):
lgtm! :+1: 

This will always send the `instance_data` but glue ignores it. Maybe a switch would be nice but this is nitpicking

PR #432, opened 2026-09-14
Add feature/intgration tests for profiles

brett060102:
1: Update run-feature-tests to install pciutils and kmod
   These are needed for pci data and kernel mode data profiles

2: add helpers DeleteFile and CreateFileWithContent
   Testing the packages requires enabling rpm_packages in /etc/SUSEConnect
   which requires creating /etc/SUSEConnect with the correct contents

3: Create profiles_test.go which
   Creates /etc/SUSEConnect with correct contents.
   Registers system
   Validates that needed files were created in /run/suseconnect/
   Deregisters system
   Validated the the profile ID files were removed from /run/suseconnect/
   Deletes /etc/SUSEConnect

4: Cleanup /etc/SUSEConnect as part of test cleanup

5: Change feature test timeout to 20min, 15 not always enough.

Relates: jsc#tel-300

To test:
# checkout the branch (assumes upstream remote is origin) 
$ git checkout origin/tst-profiles

# run feature tests
$ make feature-tests
and validate:
```
=== RUN   TestProfiles
=== RUN   TestProfiles/Test_Profile_Creation
=== RUN   TestProfiles/Test_Profile_Cleanup
[cleanup] Try running suseconnect --de-register...
 => Failed to run suseconnect --de-register. Skipping..
[cleanup] Try running suseconnect --cleanup...
[cleanup] Cleanup /etc/zypp/{credentials.d, services.d, repos.d}/*...
--- PASS: TestProfiles (44.13s)

```
[cleanup] Try running suseconnect --de-register fails because part of the test is to deregister.


build/ci/run-feature-tests
```diff
@@ -20,8 +20,11 @@ install_locally_built_suseconnect "${SOURCE}/out"
 
 cleanup_testing_environment
 
+# install packages needed for profile testing
+zypper in -y pciutils kmod
```
mssola, line 24: This should go into the OS provider. As in, if it's in a container, have it in the Dockerfile. If it's on a Github action, have it on the action definition. Having it here might not be a good idea.

features/helpers/filesystem.go
```diff
@@ -84,3 +84,13 @@ func TryCurlrcCleanup() {
...
+
+func DeleteFile(file string) {
+	_ = os.Remove(file)
+}
+
+
+func CreateFileWithContent(file string, data string) {
+	DeleteFile(file)
+	_ = os.WriteFile(file, []byte(data), 0644)
+}
```
mssola, line 96: Instead of all of this, I would create a new helper called `TryConfigCleanup`, which should be called via `t.Cleanup` in the tests like the other cleanups. Then, it would check if it's there, and if so it would remove it via the already existing `RemoveFile` function.

As for `CreateFileWithContent`, I'd have a more specific function like `CreateConfig` which accepts the data and the path is assumed.

All of this because it's not like SUSEConnect creates random files we don't care about, but we know where they are, how they are spun, and when we should clean them up. Opening up the door to something so open like `CreateFileWithContent`, might clash with how feature tests have been developed so far.

brett060102: Happy to make this change. I was going more generic because I thought it might be useful at some point in the future.

build/ci/run-feature-tests
mssola: As discussed privately, you can safely ignore this comment.

mssola (approved):
Thanks! 🎉

